PR #5, opened 2020-08-24
added support for source definition lookup

### LINKED ISSUE #1: Sources definition lookup only goes to first line, not to definition

ShaoYuan-Weng:
Closes #1


mdesmet:
It should go towards the actual objectdefinition within the source file.

PR #32, opened 2020-12-16
added dbtInstaller

### LINKED ISSUE #8: Show run logs of a model

### LINKED ISSUE #19: Refresh the manifest on load
I think there might be a good case for triggering a refresh of the manifest as you load the project (or actually generate it if it doesn't exist yet)

the `dbt ls` command generates the artifacts which should be the fastest way to get it.

One caveat I see is that if people are using dbt from a venv then we might have to find what that venv is. One solution perhaps is offer a config option where people point to the venv so that it can be activated. A bit like what the Python extension is doing for its formatters and linters.

That'd be super neat!

### LINKED ISSUE #24: DBT run command is sometimes ran before venv is activated
Hi,

In the latest version, a `cd` command is sent to the terminal but it fails when the path contains some spaces.
https://github.com/innoverio/vscode-dbt-power-user/blob/f7749f2b5c462eb105d10295b4f126491ac6b70a/src/commands/runModel.ts#L45

It should be a quick fix by surrounding the path with single or double-quotes.

### LINKED ISSUE #33: Issue with path of some files when running on Windows
I am currently using the extension on a laptop running Windows.

I ran into an issue where some of the files would not show the parent/children when other ones did. After a bit of debugging, it seems to be related to the fact that `document.uri.path` returns a string with the drive letter sometimes capital and sometimes not.

Some files were showing: `path: "/c:/Users/xxx/model1.sql"` when others are showing `path: "/C:/Users/xxx/model2.sql"` (in one case the C: drive is uppercase and in the other one it is lower case).

A quick fix for this use case might be to change the file `manifestContainer.ts` and use `fsPath` instead of `path`, while catering for both Unix and Windows file systems. Replacing
```
if (currentFilePath.path.startsWith(projectRootUri.path + "/"))
```
by 
```
if (currentFilePath.fsPath.startsWith(projectRootUri.fsPath + "/") || currentFilePath.fsPath.startsWith(projectRootUri.fsPath + "\\"))
```

Happy to submit a PR if it's OK

mdesmet:
Fixes #19 
Fixes #24 
Fixes #8 
Fixes #33


src/manifest/manifest.ts
```diff
@@ -82,6 +84,11 @@ export class Manifest {
     this.createProjectConfigWatcher();
     const projectConfig = this.readAndParseProjectConfig();
 
+    const { pythonPath, onDidChangeExecutionDetails } = await getPythonPathFromExtention();
+    onDidChangeExecutionDetails(() => this.tryRefresh());
```
mdesmet, line 88: There is a manifest file for each DBT project (monorepo setup), however there is only one venv per vscode window. Therefore this is not the right place. If the user has a lot of projects he will be prompted for each DBT project for DBT installation

src/manifest/dbtInstaller.ts
```diff
@@ -0,0 +1,101 @@
...
+	constructor(pythonPath: string) {
+		this.pythonPath = pythonPath;
+	}
+
+	public async checkDBTInstalled(): Promise<void> {
+		const checkDBTInstalledProcess = this.checkDBTInstalledCommand();
+		try {
+			await checkDBTInstalledProcess.complete();
+		} catch (err) {
+			if (err.match(DBTClient.IS_NOT_INSTALLED)) {
```
mdesmet, line 30: This should be a fallback, we should not check for "no module named ...", if we can't find installed version then that means DBT is not installed.

src/manifest/dbtInstaller.ts
```diff
@@ -0,0 +1,101 @@
...
+		return new CommandProcessExecution(this.pythonPath, ["-c", "import dbt.main; dbt.main.main(['--version'])"]);
+	}
+
+	private checkDBTUpToDate(message: string): boolean | undefined {
+		const installedVersionMatch = message.match(DBTClient.IS_INSTALLED_VERSION);
+		if (installedVersionMatch === null) {
+			return;
+		}
+		const installedVersion = installedVersionMatch[0];
+		const latestVersionMatch = message.match(DBTClient.IS_LATESTEST_VERSION);
```
mdesmet, line 53: typo LATESTEST

src/manifest/manifestContainer.ts
```diff
@@ -31,7 +31,7 @@ class ManifestContainer {
   }
 
   addEventHandler(provider: OnManifestCacheChanged): void {
-    if(this.manifestMetaMap === undefined) {
+    if (this.manifestMetaMap === undefined) {
```
mdesmet, line 34: This style change should be aligned, not sure since when we do if ( instead if(

src/manifest/dbtProject.ts
```diff
@@ -99,7 +96,7 @@ export class Manifest {
         new Map(),
         this.projectRoot
       );
-      this.onManifestCacheChangedHandlers.forEach((handler) => handler(event));
+      dbtProjectContainer.passEventToProviders(event);
```
mdesmet, line 99: dbtProjectContainer should have static methods like raiseSourceFileChangedEvent and raiseManifestChangedEvent.

Within the dbtProject we don't care to whom the event is raised, so we don't need to know that only the providers are interested in this event.

src/dbt_client/dbtClientFactory.ts
```diff
@@ -0,0 +1,23 @@
...
+        onDidChangeExecutionDetails(async () => {
+            const { pythonPath } = await getPythonPathFromExtention();
+            await DBTClientFactory.setDBTClient(pythonPath);
+        });
+        await DBTClientFactory.setDBTClient(pythonPath);
+    }
+
+    static setDBTClient(pythonPath: string): Promise<void> {
+        dbtClient.pythonPath = pythonPath;
+        return dbtClient.checkDBTInstalled();
```
mdesmet, line 17: Also when the pythonpath is null you wil check if dbt is installed...

src/dbt_client/dbtClient.ts
```diff
@@ -0,0 +1,124 @@
+import { OutputChannel, window } from 'vscode';
+import { OnSourceFileChanged, SourceFileChangedEvent } from '../manifest/sourceFileChangedEvent';
+import { CommandProcessExecution } from '../utils';
+import { dbtClientCommandQueue as dbtClientCommandQueue } from './dbtClientCommandQueue';
+
+enum DBT {
+	Install,
```
mdesmet, line 7: Enum values are some kind of constants, according convention should be upperc

src/dbt_client/dbtClient.ts
```diff
@@ -0,0 +1,124 @@
...
+	Update
+}
+
+enum PromptAnswer {
+	Yes = "Yes",
+	No = "No"
+}
+
+class DBTClient implements OnSourceFileChanged {
+	public pythonPath!: string;
```
mdesmet, line 17: I would say the pythonPath is a required constructor parameter for the DBTClient, adding the exclamation mark WILL generate runtime exceptions. as pythonPath can be null if you don't select a python interpreter.

src/dbt_client/dbtClient.ts
```diff
@@ -0,0 +1,124 @@
...
+	static readonly IS_INSTALLED_VERSION = /(?<=installed\sversion:\s)(\d+.\d+.\d+)(?=\D+)/g;
+	static readonly IS_LATEST_VERSION = /(?<=latest\sversion:\s)(\d+.\d+.\d+)(?=\D+)/g;
+	static readonly IS_INSTALLED = /installed\sversion/g;
+	public outputChannel: OutputChannel;
+
+	constructor() {
+		this.outputChannel = window.createOutputChannel('DBT Client');
```
mdesmet, line 24: DBT Client is a weird name from a user perspective, just call it DBT

src/dbt_client/dbtClientCommandQueue.ts
```diff
@@ -0,0 +1,36 @@
+import { StatusBarAlignment, StatusBarItem, window } from "vscode";
+
+type Command = () => Promise<any>;
+
+class DBTClientCommandQueue {
+    private queue: Command[] = [];
+    private currentPromise: any;
+    private statusBar: StatusBarItem;
+
+    constructor() {
```
mdesmet, line 10: I'm not sure the statusbar should be integrated in this class.

I think you can always show the status bar and show the version of the installed DBT, and when it's working you update what you are running exactly, for example '<spinning wheel> listing dbt models...' or '<spinning wheel> running model(s)...'

src/dbt_client/dbtClient.ts
```diff
@@ -0,0 +1,124 @@
...
+		if (installedVersionMatch === null) {
+			return;
+		}
+		const installedVersion = installedVersionMatch[0];
+		const latestVersionMatch = message.match(DBTClient.IS_LATEST_VERSION);
+		if (latestVersionMatch === null) {
+			return;
+		}
+		const latestVersion = latestVersionMatch[0];
+		if (installedVersion === latestVersion) {
```
mdesmet, line 79: convoluted logic..

return installedVersion === latestVersion does the same as your 3 lines

src/dbt_client/dbtClient.ts
```diff
@@ -0,0 +1,124 @@
...
+
+	private checkDBTInstalledCommand(): CommandProcessExecution {
+		return new CommandProcessExecution(this.pythonPath, ["-c", this.dbtCommand("'--version'")]);
+	}
+
+	private dbtCommand(cmd: string | string[]): string {
+		return `import dbt.main; dbt.main.main([${cmd}])`;
+	}
+
+	private checkDBTUpToDate(message: string): boolean | undefined {
```
mdesmet, line 68: why this can return undefined? is that not rather an exception?

src/dbt_client/dbtClientFactory.ts
```diff
@@ -0,0 +1,23 @@
+import { SourceFileChangedEvent } from "../manifest/sourceFileChangedEvent";
+import { getPythonPathFromExtention } from "../utils";
+import { dbtClient } from "./dbtClient";
+
+export class DBTClientFactory {
+    static async createDBTClient() {
```
mdesmet, line 6: I would move the instantiation logic of this to the DBTProjectContainer. That way other code only needs to integrate with that. If there is a DBTClient on the DBTProjectContainer, that means that you have a pythonPath. That way you can also get rid of the singleton integration of DBTClient, because in that way you have to support a DBTClient without a pythonPath.

src/manifest/watchers/projectConfigWatcherFactory.ts
```diff
@@ -0,0 +1,16 @@
+import { FileSystemWatcher, RelativePattern, Uri, workspace } from "vscode";
+import { setupWatcherhHandler } from "../../utils";
+import { DBTProject } from "../dbtProject";
+
+export class ProjectConfigWatcherFactory {
```
mdesmet, line 5: This logic can be moved back towards DBTProject

src/manifest/watchers/projectConfigWatcherFactory.ts
```diff
@@ -0,0 +1,16 @@
+import { FileSystemWatcher, RelativePattern, Uri, workspace } from "vscode";
+import { setupWatcherhHandler } from "../../utils";
```
mdesmet, line 2: typo

src/utils.ts
```diff
@@ -60,3 +61,87 @@ export const isEnclosedWithinCodeBlock: (
...
+    });
+  }
+}
+
+interface PythonExecutionDetails {
+  pythonPath: string;
+  onDidChangeExecutionDetails: Event<Uri | undefined>;
+}
+
+export const getPythonPathFromExtention = async (): Promise<PythonExecutionDetails> => {
```
mdesmet, line 117: typo

src/utils.ts
```diff
@@ -60,3 +61,87 @@ export const isEnclosedWithinCodeBlock: (
...
+      this.commandProcess.once('error', error => {
+        reject(`Error occurred during process execution: ${error}`);
+      });
+    });
+  }
+}
+
+interface PythonExecutionDetails {
+  pythonPath: string;
+  onDidChangeExecutionDetails: Event<Uri | undefined>;
```
mdesmet, line 114: undefined?

src/utils.ts
```diff
@@ -60,3 +61,87 @@ export const isEnclosedWithinCodeBlock: (
 export const notEmpty = <T>(value: T | null | undefined): value is T => {
   return value !== null && value !== undefined;
 };
+
+export class CommandProcessExecution {
```
mdesmet, line 65: should be moved to where it is used

src/utils.ts
```diff
@@ -60,3 +61,87 @@ export const isEnclosedWithinCodeBlock: (
...
+export const debounce = (fn: Function, wait: number) => {
+  let timeout: NodeJS.Timeout;
+
+  return function () {
+    clearTimeout(timeout);
+    timeout = setTimeout(fn(), wait);
+  };
+};
+
+export const setupWatcherhHandler = (watcher: FileSystemWatcher, handler: Function): void => {
```
mdesmet, line 143: typo

PR #42, opened 2021-01-13
Implement DI container

mdesmet:


src/inversify.config.ts
```diff
@@ -0,0 +1,114 @@
...
+import { ProjectConfigChangedEvent } from "./manifest/event/projectConfigChangedEvent";
+import { DBTProjectLog } from "./manifest/handlers/dbtProjectLog";
+import { SourceFileWatchers } from "./manifest/handlers/sourceFileWatchers";
+import { TargetWatchers } from "./manifest/handlers/targetWatchers";
+import { ModelTreeviewProvider } from "./treeview_provider/ModelParentTreeviewProvider";
+
+export const container = new Container();
+container.load(buildProviderModule());
+
+container
```
mdesmet, line 25: if only the eventemitter is required, these might better be moved into a singleton and just provided by the default mechanism

src/inversify.config.ts
```diff
@@ -0,0 +1,114 @@
...
+      const { container } = context;
+      return new DBTWorkspaceFolder(
+        container.get("DBTProjectFactory"),
+        workspaceFolder,
+        _onManifestChanged
+      );
+    };
+  });
+
+container
```
mdesmet, line 57: Here is maybe better to create this is as an absract class and inherit with super(Treetype) and inject an instance for each type.

src/definition_provider/modelDefinitionProvider.ts
```diff
@@ -12,16 +12,18 @@ import {
   Disposable,
 } from "vscode";
 import { NodeMetaMap } from "../domain";
-import { dbtProjectContainer } from "../manifest/dbtProjectContainer";
+import { DbtProjectContainer } from "../manifest/dbtProjectContainer";
 import { ManifestCacheChangedEvent } from "../manifest/event/manifestCacheChangedEvent";
+import { provideSingleton } from "../utils";
 
+@provideSingleton(ModelDefinitionProvider)
```
mdesmet, line 19: The question I ask myself is do we need to set it as provideSingleton if we only use it in one place?

PR #60, opened 2021-04-09
Support dots in model names

monti-python:
Make the extension work with models that contain dots, as suggested in this [dbt post](https://discourse.getdbt.com/t/extracting-schema-and-model-names-from-the-filename/575)


mdesmet:
Great work! Thanks, I'll merge it and will make a new release shortly.

PR #65, opened 2021-04-15
run sql through dbt task

### LINKED ISSUE #41: See resultset of model

mdesmet:
Fixes #41

### DIFF
--- a/src/commands/executeSQL.ts
+++ b/src/commands/executeSQL.ts
@@ -1,0 +1,60 @@
+import path = require("path");
+import { Disposable, window } from "vscode";
+import { NodeMetaMap } from "../domain";
+import { DBTProjectContainer } from "../manifest/dbtProjectContainer";
+import { ManifestCacheChangedEvent } from "../manifest/event/manifestCacheChangedEvent";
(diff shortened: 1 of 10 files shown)


src/commands/executeSQL.ts
```diff
@@ -0,0 +1,60 @@
...
+      if (node === undefined) {
+        return;
+      }
+      let fqn = "";
+      if(node.database) {
+        fqn += `${node.database}.`;
+      }
+      fqn += `${node.schema}.${node.alias}`;
+
+      this.dbtProjectContainer.executeSQL(dbtProject.projectRoot, `SELECT * FROM ${fqn} LIMIT 1000`);
```
Elektry-On, line 44: The LIMIT clause only works for some databases. Would be cool to also support the TOP syntax (maybe over a setting) for community supported adapters.

mdesmet: I doubt that dbt has a native way to limit the resultset they just provide us with a bool fetch. I might leave out that limit clause in the initial version and only provide a query that can be modified at will.

PR #112, opened 2022-05-13
✨ Added a DBT file icon

Luttik:
This plugin adds a DBT file icon to all files with the `jinja-sql` language.
In doing so the language is also added to the contribution list. This also allows more configuration if desired [(documentation)](https://code.visualstudio.com/api/references/contribution-points#contributes.languages)

Result:
image


mdesmet:
Thanks for your contribution!

Luttik:
Hé @mdesmet Could you trigger another release to the vscode marketplace?

PR #136, opened 2022-09-29
Restrict execute dbt SQl to only execute when the editor has focus

### LINKED ISSUE #135: windows: ctrl+Enter commit overridden by this extension
## issue
Currently when having this installed, if I put in a commit message and press control+enter, `dbtPowerUser.executeSQL` runs.  

## Possible fix
In [package.json](https://github.com/innoverio/vscode-dbt-power-user/blob/master/package.json#L204)

change

```json
        "when": "resourceLangId == jinja-sql"
```

to 

```json
        "when": "editorTextFocus && resourceLangId == 'jinja-sql'"
```

VDFaller:
Made the ctrl+enter shortcut work only while in the editor

### DIFF
--- a/package.json
+++ b/package.json
@@ -201,7 +201,7 @@
         "key": "Ctrl+Enter",
         "mac": "Cmd+Enter",
         "command": "dbtPowerUser.executeSQL",
-        "when": "editorTextFocus && resourceLangId == 'jinja-sql'"
+        "when": "editorFocus && resourceLangId == 'jinja-sql'"
       },
       {
         "key": "Ctrl+'",
--- a/package.json
+++ b/package.json
@@ -201,7 +201,7 @@
         "key": "Ctrl+Enter",
         "mac": "Cmd+Enter",
         "command": "dbtPowerUser.executeSQL",
-        "when": "resourceLangId == jinja-sql"
+        "when": "editorTextFocus && resourceLangId == 'jinja-sql'"
       },
       {
         "key": "Ctrl+'",


package.json
```diff
@@ -201,7 +201,7 @@
         "key": "Ctrl+Enter",
         "mac": "Cmd+Enter",
         "command": "dbtPowerUser.executeSQL",
-        "when": "resourceLangId == jinja-sql"
+        "when": "editorTextFocus && resourceLangId == 'jinja-sql'"
```
mdesmet, line 204: Why not `editorFocus`?

 https://code.visualstudio.com/api/references/when-clause-contexts

Do you think this should be applied to other commands as well?

VDFaller: My thought is that I should actively be in a jinja sql textbox for ctrl+enter to use this command.  (since a lot of things use it)

for the sql preview one.  I don't have anything installed that uses ctrl+', so it's not like it will overwrite another command.  

The other ones I didn't change because if I'm selecting something from the menu, then I'm explicitly asking for it.  So if I'm doing that I don't want to have to be focused in a certain spot.

VDFaller: Thinking about this a tiny bit more.  Wouldn't it make sense for the SQL preview to be the same as markdown preview
`ctrl+shift+v` (on windows)?

mdesmet: What's the behavior if you change to `editorFocus`, if it also fixes the issue, it would capture the user intent in more cases.

VDFaller: Fair enough.  It still fixes the issue.

PR #143, opened 2022-09-30
[feat] Webview and Server enhancements

alex-harness:
TODO: update PR description

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
         "properties": {
           "dbt.dbtPythonPathOverride": {
             "type": "string",
-            "description": "Path to a python executable or entrypoint. If undefined, we will use the Python extension configured environment. Most users will not need to change this setting without specific reasoning such as custom wrappers."
+            "description": "Path to a python executable or entrypoint. If undefined, we will use the Python extension's configured environment. Most users will not need to change this setting without specific reasoning such as custom wrappers and should instead modify the Python extensions selected interpreter."
           },
           "dbt.listModelsDisabled": {
             "type": "boolean",


z3z1ma:
@mdesmet 

So my questions with this PR. 

1: The only way I can test the extension in production is via the python entrypoint setting. Its probably easy to split that into another PR but also I wonder if its fine to keep it here. There are very few users it would affect. Perhaps just making its purpose more clear is sufficient to leave it in so I can continue to reliably test the extension at scale. Any ideas?

2: The server portion can be yanked into another PR but I pose a similar question for discussion. The way it is set up right now, the user may know nothing of the server and nothing is negatively affected. Is it valid to leave it as is with a follow-on PR for the changes we discussed around auto-instantiation, project registration, and so on.

The faster we clear this PR (if we like it) the faster we can merge in the code quality work in #144 and I can create a new PR dedicated to the rest of the server stuff. I suppose my argument is it does no harm if our intention is not to publicize it to users yet until the next version release. 
An added benefit is it also lets a good number of users on my fork of the extension as well as datacoves fork migrate early from the separate exts to this primary one post-merge.

mdesmet:
I would still prefer to split up the PR and do a release with the webview view only.

> 1: The only way I can test the extension in production is via the python entrypoint setting. Its probably easy to split that into another PR but also I wonder if its fine to keep it here. There are very few users it would affect. Perhaps just making its purpose more clear is sufficient to leave it in so I can continue to reliably test the extension at scale. Any ideas?

Can you retest your case with following settings:

I added a Python wrapper (my fake meltano executable):

```
#!/bin/sh
/some path to/python "$@"
```

in .vscode/settings.json

```
{
  "python.defaultInterpreterPath": "path to your custom meltano/python executable"
}
```

Then do 

image

After doing this, dbt was correctly detected.

I prefer this way as there is no additional setting being introduced. This process can be documented in README.md or Meltano docs and would not confuse any users if they need to set a certain property or not.

> 2: The server portion can be yanked into another PR but I pose a similar question for discussion. The way it is set up right now, the user may know nothing of the server and nothing is negatively affected. Is it valid to leave it as is with a follow-on PR for the changes we discussed around auto-instantiation, project registration, and so on.

I think there is no need to have a custom setting for server address/port. Most extensions that work with language servers don't provide this (eg. [Rust](https://github.com/rust-lang/vscode-rust/blob/8dbe67de99643dadbf9552c933d635e3addb8e82/package.json#L192)). I think the extension should manage the server instead of externalizing it. This way it stays internal to the extension how this feature is implemented, for example if we later on would want to use dbt server.

package.json
```diff
@@ -340,10 +343,13 @@
...
     "dayjs": "^1.8.34",
     "inversify": "^5.0.5",
     "inversify-binding-decorators": "^4.0.0",
+    "node-abort-controller": "^3.0.1",
+    "node-fetch": "^3.2.10",
     "reflect-metadata": "^0.1.13",
     "yaml": "^2.0.0-10"
   }
-}
+}
```
mdesmet, line 355: Please add new line at EOF

src/commands/installDbtOsmosis.ts
```diff
@@ -9,20 +9,21 @@ enum PromptAnswer {
...
 
   private async askForInstall() {
     const answer = await window.showInformationMessage(
-      "dbt Power User can now execute and visualize the result set of your models using an auxilliary package `dbt-osmosis`. Do you want to install `dbt-osmosis` now?",
+      "dbt Power User can now execute and visualize the result set of your models using an auxiliary package `dbt-osmosis`. Do you want to install `dbt-osmosis` now? (reloads window after install)",
       PromptAnswer.YES,
       PromptAnswer.NO
     );
     if (answer === PromptAnswer.YES) {
       this.dbtProjectContainer.installDbtOsmosis();
```
mdesmet, line 25: I think you need to add `await` here.

src/commands/installDbtOsmosis.ts
```diff
@@ -9,20 +9,21 @@ enum PromptAnswer {
...
+      commands.executeCommand("workbench.action.reloadWindow");
     }
   }
 }
```
mdesmet, line 29: Add new line

src/dbt_client/dbtCommandFactory.ts
```diff
@@ -27,10 +27,24 @@ export class DBTCommandFactory {
     return dbtProfilesDir ? ["'--profiles-dir'", `r'${dbtProfilesDir.fsPath}'`] : [];
   }
 
+  private getFirstWorkspacePath(): string {
```
mdesmet, line 30: I think we can also implement this differently. We can first detect any dbt projects and then launch the dbt detection on the first found dbt project.

This is related to simplifying the dbtProjectContainer vs dbtProject dependency wiring. We can get rid of a lot redirection functions by reducing the responsibilities of dbtProjectContainer.

src/webview_view/queryResultPanel.ts
```diff
@@ -0,0 +1,268 @@
...
+    const possible = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+    for (let i = 0; i < 32; i++) {
+        text += possible.charAt(Math.floor(Math.random() * possible.length));
+    }
+    return text;
+}
+
+function getUri(webview: Webview, extensionUri: Uri, pathList: string[]) {
+    return webview.asWebviewUri(Uri.joinPath(extensionUri, ...pathList));
+}
```
mdesmet, line 268: Add new line

src/manifest/dbtProject.ts
```diff
@@ -206,11 +215,12 @@ export class DBTProject implements Disposable {
   }
 
   executeSQL(query: string, title: string) {
-    this.queryResultPanelLoader.showWebview(title).executeQuery(
-      query, 
-      this.projectRoot, 
-      this.dbtProfilesDir, 
-      this.profilesMetaData!.defaultTarget
+    this.queryResultPanel.executeQuery(
```
mdesmet, line 218: Remove no-op changes

src/manifest/dbtProjectContainer.ts
```diff
@@ -23,7 +23,7 @@ export class DBTProjectContainer implements Disposable {
   private dbtWorkspaceFolders: DBTWorkspaceFolder[] = [];
   private _onManifestChanged = new EventEmitter<ManifestCacheChangedEvent>();
   public readonly onManifestChanged = this._onManifestChanged.event;
-  public currentSql?:string;
+  public currentSql?: string;
```
mdesmet, line 26: AFAIK there was no need anymore for a global currentSql. As mentioned before, I would like to reduce the responsibilities of dbtProjectContainer.

package.json
```diff
@@ -49,7 +53,7 @@
           },
           "dbt.profilesDirOverride": {
             "type": "string",
-            "description": "Override the profiles dir for every dbt command. Absolute path or relative to the dbt_project.yml"
+            "description": "Override the profiles dir for every dbt command. Absolute path or relative to the profiles.yml"
```
mdesmet, line 56: Good spot!

```suggestion
            "description": "Override the profiles dir for every dbt command. Absolute path or relative to the directory containing profiles.yml"
```

src/commandProcessExecution.ts
```diff
@@ -63,12 +63,14 @@ export class CommandProcessExecution implements Disposable {
 
       this.commandProcess.once("close", () => {
         if (!stdoutBuffer) {
+          console.warn(stderrBuffer);
```
mdesmet, line 66: Indeed better for debugging.

src/osmosis_client/index.ts
```diff
@@ -0,0 +1,131 @@
+import { window, workspace } from 'vscode';
```
mdesmet, line 1: Move to separate PR

src/webview_view/queryResultPanel.ts
```diff
@@ -0,0 +1,268 @@
...
+        });
+        await this.transmitData(columns, rows, query, result.compiled_sql);
+    };
+
+    public async executeQuery(query: string, projectRootUri: Uri, profilesDir: Uri, target: string, title?: string) {
+        // if (title) { this._panel!.title = title; }
+        this.transmitLoading();
+        let result = await runQuery(query, 100);
+        if (isError(result)) {
+            if (result.error.code !== OsmosisErrorCode.FailedToReachServer) {
```
mdesmet, line 147: Move server related to separate PR

src/dbt_client/index.ts
```diff
@@ -41,7 +41,7 @@ export class DBTClient implements Disposable {
     private queue: DBTCommandQueue,
     private commandProcessExecutionFactory: CommandProcessExecutionFactory,
     private terminal: DBTTerminal
-  ) {}
+  ) { }
```
mdesmet, line 44: Remove all no-op changes.

src/dbt_client/index.ts
```diff
@@ -297,7 +297,11 @@ export class DBTClient implements Disposable {
 
   private async handlePythonExtension(): Promise<void> {
     const pythonEnvironment = await this.pythonEnvironment.getEnvironment();
-    this.pythonPath = pythonEnvironment.getPythonPath();
+    this.pythonPath = getPythonPathFromConfig() || pythonEnvironment.getPythonPath();
     await this.checkAllInstalled();
   }
 }
+
+function getPythonPathFromConfig(): string | undefined {
```
mdesmet, line 305: Move to separate PR.

src/extension.ts
```diff
@@ -6,13 +6,11 @@ import { container } from "./inversify.config";
 export async function activate(context: ExtensionContext) {
```
mdesmet, line 6: Remove no-op changes.

mdesmet (commented):
As discussed above, please split up in smaller pieces.

src/manifest/dbtProject.ts
z3z1ma: Legitimate change

.eslintrc.json
```diff
@@ -14,7 +14,6 @@
         "curly": "warn",
         "eqeqeq": "warn",
         "no-throw-literal": "warn",
-        "semi": "off",
-        "indent": ["error", 2]
+        "semi": "off"
     }
-}
+}
```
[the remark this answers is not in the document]
z3z1ma: Leaving `"indent": [ "error", 2, { "SwitchCase": 1 } ]` for the purpose of proper switch statement indentation

src/dbt_client/dbtCommandFactory.ts
z3z1ma: Because this is working as-is I suggest we move this to a separate PR encapsulating other dbtProjectContainer vs dbtProject changes where we can improve it and we get this one over the line.

src/dbt_client/index.ts
z3z1ma: I think we still need this. I've made arguments already in its favor and I think its better to have an escape hatch for people who need it if it doesn't fundamentally affect 80% of users except those that need/want it. 

I think I am doubly motivated to keep this other than the fact its so minor, in that it helps me test the extension at a reasonable scale; not a toy project, but on multiple real production projects which benefits our iteration cycle with a heavy dose of UAT.

z3z1ma: As a compromise I made the verbiage extremely clear that most users wont need to touch this.

> Path to a python executable or entrypoint. If undefined, we will use the Python extension configured environment. Most users will not need to change this setting without specific reasoning such as custom wrappers and should instead modify the Python extensions selected interpreter.

PR #304, opened 2023-02-05
Added Build command

GiorgiModebadzeW:
## Checklist

added Build command
Screenshot 2023-02-05 at 23 54 52


src/dbt_client/dbtCommandFactory.ts
```diff
@@ -101,6 +101,44 @@ export class DBTCommandFactory {
...
 
+  createBuildModelCommand(
+    projectRoot: Uri,
+    profilesDir: string,
+    params: RunModelParams
+  ): DBTCommand {
+    const { plusOperatorLeft, modelName, plusOperatorRight } = params;
+    const profilesDirParams = this.profilesDirParams(profilesDir);
+
+    const buildModelCommandAdditionalParams = workspace
```
mdesmet, line 112: Please also add this configuration parameter in package.json.

I also wonder if we want to configure different parameters compared to the run feature?

mdesmet (commented):
LGTM

Please also add the feature in the README.

src/dbt_client/dbtCommandFactory.ts
GiorgiModebadzeW: Right now way I use to build is pretty similar to run - its basically 1 to 1 - 
I just default to build instead of run as it also takes away the need for running tests separately - but if you have any usecase in mind where different parameters can be useful I am more than happy to look into it

mdesmet: No I don't have but in your PR you provide ways to configure params through a separate configuration parameter, `buildModelCommandAdditionalParams` but it's not configured in package.json, so users would not be able to discover easily.

mdesmet: Just look for `dbt.runModelCommandAdditionalParams` in package.json

GiorgiModebadzeW: Thanks - did not notice that 
updated + added part in readme

mdesmet (approved):

mdesmet:
Thanks for the valuable contribution!

GiorgiModebadzeW:
thanks for making this extension - its life saver

PR #309, opened 2023-02-07
Lineage graph

jlpjavi:
## Overview
The implementation is based on the parent, children, and test information but presented in graph form. We can think about iterating this further in future versions but I think this is a good enough first version of the feature.

Captura de Pantalla 2023-03-22 a las 23 03 36

Captura de Pantalla 2023-03-21 a las 23 06 53

## Checklist


mdesmet:
How the user will distinguish between the model types. Shouldn't we provide a legend if colors are used?

jlpjavi:
> How the user will distinguish between the model types. Shouldn't we provide a legend if colors are used?

Yes, I'll look into the library to see if the legend can be static, I think is possible

jlpjavi:
> How the user will distinguish between the model types. Shouldn't we provide a legend if colors are used?

Update on this, I have been checking to add a legend and this is only supported after AntV G6 4 version, the thing is if I get the new version the app crashes. I don't know if it has to do with the type of modules that the VSCode extension can import but I couldn't manage to make it work. Maybe @mdesmet you have had some similar problems in the past with other libraries?

Issue #237, opened 2022-11-30
vscode variables and relative paths don't work for profile dir override


juls858:
### Expected behavior
doesn't work:
 - using ${workspaceFolder} should allow extension to function properly
 - relative path `./`
works:
- hard coding the absolute path work

### Actual behavior

An error occured while initializing the dbt project: Runtime Error
  Could not find profile named 'ethereum'

### Steps To Reproduce

1. update profiles dir override value to `${workspaceFolder}`
1. reload window
2. open dbt model

### Log output/Screenshots

![image](https://user-images.githubusercontent.com/14811322/204906950-5c480e5d-c653-4949-bdfb-818e698f1154.png)

VSCode Version

> Version: 1.73.1 (Universal)
> Commit: 6261075646f055b99068d3688932416f2346dd3b
> Date: 2022-11-09T02:08:38.961Z (3 wks ago)
> Electron: 19.0.17
> Chromium: 102.0.5005.167
> Node.js: 16.14.2
> V8: 10.2.154.15-electron.0
> OS: Darwin x64 22.1.0
> Sandboxed: No

### Operating System

macOS 13.0.1 (22A400)

### dbt version

1.3.1

### dbt Adapter

snowflake

### dbt Power User version

v0.10.1

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

mdesmet:
Vscode doesn't provide the variables substitution logic as an API and there is no reliable library for parsing them.

If I remember correctly putting just a `.` works. I think you can't put a `/` at the end. See https://docs.getdbt.com/docs/get-started/connection-profiles#2-use-the-dbt_profiles_dir-environment-variable-to-change-the-default-location

mdesmet:
I did add support for `${workspaceFolder}` in 0.10.3.

Please reopen the issue if you have any issues.

aa-matthias:
Hi @mdesmet - it is still not working here having a workspace/project config like this:

```yml
    "terminal.integrated.env.windows": {
        "DBT_PROFILES_DIR": "./.dbt"
    },
```

moving the .dbt folder to ~ solved the problem so it looks like the DBT_PROFILES_DIR variable is not read correctly.

Issue #199, opened 2022-11-13
Ensure `dbt.pythonPathOverride` works with relative paths, tilde (Meltano)


mdesmet:
![image](https://user-images.githubusercontent.com/9265503/201538278-098a20b6-bb4b-4122-b1f3-b760e7131fd1.png)

aaronsteers:
Related: I found that `${workspaceFolder}` also doesn't seem to work as expected in `dbt.profilesDirOverride` or `dbt.dbtPythonPathOverride`.

I worked around this by using `./...` in place of `${workspaceFolder}/...`.


https://code.visualstudio.com/docs/editor/variables-reference#_predefined-variables

mdesmet:
> Variable substitution is supported inside some key and value strings in launch.json and tasks.json files using ${variableName} syntax.

Variable substitution is a common complaint in extensions. Currently vscode only does it in certain places and doesn't expose this feature.

If you are saying `.` and `..` work as expected we can close this issue. Could you also test `~`?

aaronsteers:
> If you are saying `.` and `..` work as expected we can close this issue. Could you also test `~`?

I did not test `..` but I confirmed that `.` works as expected, in place of `${workspaceFolder}`

> Variable substitution is a common complaint in extensions. Currently vscode only does it in certain places and doesn't expose this feature.

Completely understandable. The VS Code docs are not super clear on this so as a user I was initially not sure what to expect. Your docs do have the caveat though about variables not being expanded, which I appreciate.

PR #331, opened 2023-02-15
Subsitute variables in settings

flexponsive:
## Overview

Expand `${workspaceFolder}` and `${env:VARIABLE}` placeholders in the settings for `dbt.profilesDirOverride` and `dbtPythonPathOverride` using the logic previously only used for environment variable settings. This address #237 and #199.

As was discussed in those issues, variable substitution is a common issue with vscode extensions and the recommended workaround was to use paths relative to the workspace directory. But with version 0.11.4 installed from the vscode marketplace, this did not work for me anymore - only absolute paths worked.

Merging this feature would be a great help for Meltano integration because settings.json can then be added to the repository, ensuring a consistent experience for all developers (see https://github.com/meltano/meltano/issues/2811)

## Checklist

## Notes
- had to bump engine version due to complaint by `vsce package`
- thank you for this great extension, it makes working with dbt a real pleasure!

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   },
   "version": "0.11.4",
   "engines": {
-    "vscode": "^1.74.0"
+    "vscode": "^1.75.1"
   },
   "categories": [
     "Other"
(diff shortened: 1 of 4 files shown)


mdesmet (approved):

flexponsive:
Thanks!

PR #437, opened 2023-05-15
updated icons

### LINKED ISSUE #410: More intuitive icons
### Describe the feature

I find the icons could be improved to be more intuitive. I'm a new user, and always need to hover over the icons to be sure what is what:
image
furthermore, when running parent models or child models, the icon used is the same triangle/play button that's used for execute dbt SQL, rather than the database icon for run:
image

image

Here's some emoji-suggestions that would be intuitive for me, but they should probably use some open-source vectors instead? (Not entirely sure how this space works)
- Execute dbt SQL and compile: the current icons are good
- Build dbt Model: 🛠️
- Run dbt Model: 🏃 or something like image
- Run dbt tests: e.g. 🧪 or image

icon for running parent models or child models should prob be the same as "Run dbt Model"

### Who will benefit?

New users will more quickly be able to understand what the buttons do. (on the contrary existing users might be a bit confused to start with 🙃)

I don't know how to change the code, but I'll happily look for the relevant icons, if I'm told what license terms are needed for the icons!

djmikeale:
## Overview
resolves #410 
 
adds a few new icons for build + run, and uses the existing [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html) for tests + upstream build + downstream build. 
Updating text for test as well

## Checklist


package.json
```diff
@@ -420,7 +414,7 @@
     "terser-webpack-plugin": "^5.3.7",
     "ts-loader": "^8.0.3",
     "typescript": "^4.9.5",
-    "webpack": "^5.82.0",
+    "webpack": "^5.82.1",
```
mdesmet, line 417: Please revert this change and the change to the package-lock file as that is not related to your PR.

mdesmet (approved):
Looks good, when you remove the package-lock and webpack change it is good to merge.

Issue #380, opened 2023-03-20
Add model documentation to DBT power user view


carlgronvald:
### Describe the feature

I think it would be useful to add the documentation of a model to the DBT power user panel view. 

### Describe alternatives you've considered

Documentation on model hover, #96 , could also provide some of the same benefits.

### Who will benefit?

Developers would be able to see documentation while working on a model. This has two functions:
1. Model documentation is more easily understood by developers
2. Because model documentation is shown while developing, there is more of an onus to keep the documentation up to date, as drift between the truth of the model and the documentation is immediately apparent.

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

mdesmet:
This seems very useful. Would you be able to provide a PR?

PR #465, opened 2023-06-24
added documentation tree view

### LINKED ISSUE #77: Feature request: Add the ability to jump to the corresponding yml file of a given model
Hi!

One feature that I think would be useful would be the ability when in a model, to quickly jump to the corresponding schema file (the `.yml` one). This would allow users to check what tests are currently being defined and modify/add existing tests and column descriptions.

From what I can see, all models that have been defined in a schema file have this data saved in the manifest.json file under 
`nodes["<node_name>"].patch_path`. But some nodes just don't have this key if they have not been manually added to a schema file.

b-ned:
Implemented Documentation Pane in the Power User Extension window.

For Columns: Ability to see columns + type + description.
For table: Ability to see materialization and schema.

The code scans the `catalog.json` and `manifest.json` for attributes and throws a warning when there is no documentation found.

Features:
- Clicking on the table name will open the schema file if it exists. If it doesn't it will throw an error.
- Clicking play next to `DOCUMENTATION` will trigger a `dbt docs generate`
- Clicking the button next to the model, will generate the YML if it does not exist.

doc-treeview-demo mov

resolves #77 #380

Note that I created a new DocumentationTreeviewProvider, instead of merging it to the existing one as it is quite different. There might be a cleaner solution, merging the two TreeProviders. I'm open for tips & feedback!

Future improvements:
- Create a function that scans the YML + compares it to the columns of the model (using `project.get_columns_in_relation`)
- For Model: Add tooltip that shows description of model when hovering. (I didn't get the tooltips to work properly, they kept showing the `label` property instead of assigned values. So I left it out for this PR)
- For Columns (Children): Add tooltip that shows the tests, description etc.
- Hide Generate YML button when YML file already exists.

## Checklist

- [ x ] I have run this code and it appears to resolve the stated issue
- [ x ] `README.md` updated and added information about my change


b-ned:
Hi @mdesmet,

The tests keep failing. 

It looks like the test failing started after https://github.com/innoverio/vscode-dbt-power-user/pull/459
However, when looking at the logging of earlier PR's, I can see the same errors (while the tests are marked successful)

I tried to as a demo test to the package.json (`"test": "echo \"Error: no test specified\" && exit 1"`), but with no success..
Any suggestions on how to solve this issue?

mdesmet:
i'll have a look at this. 

BTW I'm not convinced we should be reading the manifest directly in the side pane. Why not extract the information to the event when the manifest changes (similar to the other components) and just display the information?

b-ned:
Thanks!
I'll try to rebuild it with your suggested approach,

b-ned:
Hi @mdesmet,
I rewrote the logic to use the ManifestCacheProjectAddedEvent class. Let me know what you think.
One issue with only using the `manifest.json` is that the column `data_type` is not populated, while in the `catalog.json` it is defined as `type`. I left the column data type out of the feature for now, but for future updates, it might be valuable to create a similar event structure for the catalog file as for the manifest file and add a column data type.

mdesmet:
Looks good! I will have a deeper look this weekend.

Please remove the second commit as it is not relevant.

Which dbt versions did you test with?

b-ned:
So far I've only tested it on dbt 1.4.5

mdesmet (approved):

PR #484, opened 2023-07-24
Added table scale field

Yacobolo:
## Overview
resolves ##438
Added table scaler 
image
image

## Checklist


mdesmet (commented):
Looks great, I like it very much. Shouldn't we save the scale factor in the vscode config, similar like the [queryTemplate](https://github.com/innoverio/vscode-dbt-power-user/blob/master/src/webview_view/queryResultPanel.ts#L189) ?

Yacobolo:
Thanks! Yes we should, I'll make a 2nd revision.

mdesmet (approved):

PR #517, opened 2023-08-10
Option to skip re-parsing the project

jsnb-devoted:
## Overview
We don't allow for users to hardcode the data warehouse credentials so we are using the vscode settings to set the env variables to empty strings:
```
  "terminal.integrated.env.linux": {
    "SNOWFLAKE_ACCOUNT": "",
    "SNOWFLAKE_WAREHOUSE": "",
    "SNOWFLAKE_PASSWORD": "",
```
We still get all the benefit of the features that don't require the data warehouse connection but the empty strings are causing our `dbt` commands to trigger a full parse.

We need some way to skip the regenerating of the manifest. We are satisfied with users doing that on their own manually from the command line.

This is more of a proposal than a PR. I haven't taken the time yet to figure out how to test this out.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,10 @@
             "default": [],
             "description": "Add additional params to the dbt build model command."
           },
+          "dbt.skipParsing": {
+            "type": "boolean",
+            "description": "Option to skip re-parsing the project in the background."
+          },
           "dbt.profilesDirOverride": {
             "type": "string",
             "description": "Override the profiles dir for every dbt command. Absolute path or relative to the directory containing profiles.yml"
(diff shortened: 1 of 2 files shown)


mdesmet:
Another way to look at the issue is how we can avoid full parse. Is there no way to let the extension reuse the credentials as in a normal dbt invocation?

jsnb-devoted:
@mdesmet -- having separate profiles totally solves for this problem for us. The trade off is that you have to have 2 target directory with 2 manifests but we are totally ok with that.
I'm fine to close this and use the target profile for our purposes once that gets released. I suspect there aren't a ton of folks that need this.

PR #521, opened 2023-08-11
Add the `targetPathOverride` config value

### LINKED ISSUE #444: Support manual override of target-path
### Describe the feature

In our repo we would like to use jinja in our `target-path`, but this is not supported by the extension, as described in [this issue](https://github.com/innoverio/vscode-dbt-power-user/issues/189).

If the extension supported a manual override of target-path (i.e. an textbox that lets you optionally specify which folder the extension should use as `target-path`), we would be able to keep jinja code in `target-path`.

_No response_

### Who will benefit?

Those who use jinja code in target-path. This is particularly useful when you have a large repo with long parse times, to keep a separate target-path for each target - so you avoid a full reparse everytime you switch target. According to [dbt](https://www.getdbt.com/blog/analytics-engineering-next-step-forwards/#:~:text=Complexity%20in%20the%20dbt%20ecosystem), a big (and growing) share of dbt deployments have a large number of models.

jsnb-devoted:
## Overview
    
resolves #444 

Creates a new configuration option for overriding the target path. This should help for folks that have Jinja in their dbt_project file `target-path` value and for situations where you want a separate manifest that the extension manages vs the manifest you use for development.

I tested this against the dbt `jaffle_shop` demo repo. If you set the path to `"dbt.targetPathOverride": "${workspaceFolder}/target_2",` the extension will read and write to the `target_2` dir. Confirmed in the debug output that `Using target path /Users/jasonb/code/jaffle_shop/target_2` showed up.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
           },
           "dbt.targetPathOverride": {
             "type": "string",
-            "description": "Override the target-path value for dbt processes run by the extension."
+            "description": "Override the target-path dir for every dbt command. Absolute path or relative to the directory containing manifest.json."
           },
           "dbt.queryLimit": {
             "type": "integer",
--- a/package.json
+++ b/package.json
@@ -72,6 +72,10 @@
             "type": "string",
             "description": "Override the profiles dir for every dbt command. Absolute path or relative to the directory containing profiles.yml"
           },
+          "dbt.targetPathOverride": {
+            "type": "string",
+            "description": "Override the target-path value for dbt processes run by the extension."
+          },
           "dbt.queryLimit": {
             "type": "integer",
             "description": "The maximum number of rows the `Preview SQL Query` command returns.",
--- a/src/manifest/dbtProject.ts
+++ b/src/manifest/dbtProject.ts
@@ -643,13 +643,17 @@ select * from renamed
     const targetPathOverride = workspace
       .getConfiguration("dbt")
       .get<string>("targetPathOverride");
-    if (targetPathOverride !== undefined) {
-      return substituteSettingsVariables(targetPathOverride);
-    }
-    if (projectConfig[DBTProject.TARGET_PATH_VAR] !== undefined) {
-      return projectConfig[DBTProject.TARGET_PATH_VAR] as string;
-    }
-    return "target";
+
+    return (
+      (targetPathOverride
+        ? substituteSettingsVariables(targetPathOverride)
+        : false) ||
+      process.env.DBT_TARGET_PATH ||
+      (projectConfig[DBTProject.TARGET_PATH_VAR]
+        ? (projectConfig[DBTProject.TARGET_PATH_VAR] as string)
+        : false) ||
+      "target"
+    );
   }
 
   private async refresh() {
--- a/src/manifest/dbtProject.ts
+++ b/src/manifest/dbtProject.ts
@@ -123,6 +123,7 @@ export class DBTProject implements Disposable {
     console.log("Using profile directory " + this.dbtProfilesDir);
     this.projectName = projectConfig.name;
     this.targetPath = this.findTargetPath(projectConfig);
+    console.log("Using target path " + this.targetPath);
     this.sourcePaths = this.findSourcePaths(projectConfig);
     this.macroPaths = this.findMacroPaths(projectConfig);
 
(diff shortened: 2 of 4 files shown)


src/manifest/dbtProject.ts
```diff
@@ -639,6 +640,12 @@ select * from renamed
   }
 
   private findTargetPath(projectConfig: any): string {
+    const targetPathOverride = workspace
+      .getConfiguration("dbt")
+      .get<string>("targetPathOverride");
+    if (targetPathOverride !== undefined) {
+      return substituteSettingsVariables(targetPathOverride);
+    }
```
mdesmet, line 648: Could you also add support for getting the target path from DBT_TARGET_PATH environment variable?

src/manifest/parsers/index.ts
```diff
@@ -114,11 +114,16 @@ export class ManifestParser {
   }
 
   private readAndParseManifest(projectRoot: Uri, targetPath: string) {
-    const manifestLocation = path.join(
+    let manifestLocation = path.join(
```
mdesmet, line 117: Maybe change this into a check if the supplied path is relative vs absolute?

jsnb-devoted: Ya - good call. Can do.

src/manifest/dbtProject.ts
jsnb-devoted: Sure I can add that in.

jsnb-devoted: I pattern matched what the `DBT_PROFILE_DIR` logic was doing but I was having a hard time getting vscode to recognize the env variable with `terminal.integrated.env.osx`. I was able to test that the override setting still worked and that if nothing was set it would default to `target`.

src/manifest/parsers/index.ts
```diff
@@ -114,11 +114,11 @@ export class ManifestParser {
...
-    const manifestLocation = path.join(
-      projectRoot.fsPath,
-      targetPath,
-      DBTProject.MANIFEST_FILE,
-    );
+    let manifestLocation = path.join(targetPath, DBTProject.MANIFEST_FILE);
+    if (!path.isAbsolute(targetPath)) {
+      manifestLocation = path.join(projectRoot.fsPath, manifestLocation);
+    }
+
```
mdesmet, line 121: Please rewrite back to a const expression and destructure an array onto `path.join`.

Something like (untested):


```suggestion
    const pathParts = [targetpath];
    if (!path.isAbsolute(targetPath)) {
      pathParts.unshift(projectRoot.fsPath);
    }
    const manifestLocation = path.join(...pathParts, DBTProject.MANIFEST_FILE);

```

src/manifest/dbtProject.ts
```diff
@@ -639,10 +640,20 @@ select * from renamed
...
+    return (
+      (targetPathOverride
+        ? substituteSettingsVariables(targetPathOverride)
+        : false) ||
+      process.env.DBT_TARGET_PATH ||
+      (projectConfig[DBTProject.TARGET_PATH_VAR]
+        ? (projectConfig[DBTProject.TARGET_PATH_VAR] as string)
+        : false) ||
+      "target"
+    );
```
mdesmet, line 656: PTAL at `PythonEnvironment` class and ensure that the env var parsing holds true to the rules set in https://github.com/innoverio/vscode-dbt-power-user#environment-variables.

jsnb-devoted: Thanks -- that makes a little more sense. I got it working with [this commit](https://github.com/innoverio/vscode-dbt-power-user/pull/521/commits/b429fccb27f0b218558ab29a84e910f0658da4de).

Should [this line](https://github.com/innoverio/vscode-dbt-power-user/blob/master/src/manifest/dbtProject.ts#L120)
also use the `PythonEnvironment` instead of `process.env`?

mdesmet: Yeah. good spot!

package.json
```diff
@@ -72,6 +72,10 @@
             "type": "string",
             "description": "Override the profiles dir for every dbt command. Absolute path or relative to the directory containing profiles.yml"
           },
+          "dbt.targetPathOverride": {
+            "type": "string",
+            "description": "Override the target-path value for dbt processes run by the extension."
```
mdesmet, line 77:
```suggestion
            "description": "Override the target-path dir for every dbt command. Absolute path or relative to the directory containing manifest.json."
```

mdesmet (approved):

PR #528, opened 2023-08-15
Implement telemetry

mdesmet:
## Checklist

### DIFF
--- a/src/telemetry/index.ts
+++ b/src/telemetry/index.ts
@@ -1,0 +1,45 @@
+import TelemetryReporter from "@vscode/extension-telemetry";
+import * as vscode from "vscode";
+import { provideSingleton } from "../utils";
+
+@provideSingleton(TelemetryService)
+export class TelemetryService implements vscode.Disposable {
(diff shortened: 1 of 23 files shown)


src/telemetry/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  private isTelemetryEnabled(): boolean {
+    return (
+      vscode.workspace
+        .getConfiguration()
+        .get<boolean>("telemetry.enableTelemetry") !== false
+    );
+  }
```
lionsardesai, line 52: This should not be overridden. isTelemetryEnabled flag should already be handled as part of the telemetry package.
I think originally the idea was for us to have an additional flag so that people can turn off telemetry just for this plugin and not need to turn it off for everything. The config name needs to be under dbt. asked @anandgupta42  to update his branch with an example.

lionsardesai: reference from vscode docs: https://code.visualstudio.com/api/extension-guides/telemetry

PR #554, opened 2023-08-31
Expose exceptions in diagnostics

mdesmet:
## Checklist

### DIFF
--- a/dbt_integration.py
+++ b/dbt_integration.py
@@ -333,8 +333,9 @@ class DbtProject:
         try:
             self.parse_project(init=reinit)
             self.write_manifest_artifact()
-        except Exception:
+        except Exception as e:
             self.config = _config_pointer
+            raise Exception(str(e))
 
     def write_manifest_artifact(self) -> None:
         """Write a manifest.json to disk"""
--- a/src/diagnostics/index.ts
+++ b/src/diagnostics/index.ts
@@ -1,0 +1,12 @@
+import { Disposable, languages } from "vscode";
+import { provideSingleton } from "../utils";
+
+@provideSingleton(DbtPowerUserDiagnostics)
+export class DbtPowerUserDiagnostics implements Disposable {
+  public readonly rebuildManifestDiagnostics =
+    languages.createDiagnosticCollection("dbt");
+
+  dispose() {
+    this.rebuildManifestDiagnostics.dispose();
+  }
+}
--- a/src/manifest/dbtProject.ts
+++ b/src/manifest/dbtProject.ts
@@ -5,9 +5,11 @@ import { join } from "path";
 import { PythonBridge, pythonBridge, PythonException } from "python-bridge";
 import {
   commands,
+  Diagnostic,
   Disposable,
   Event,
   EventEmitter,
+  Range,
   RelativePattern,
   Uri,
   ViewColumn,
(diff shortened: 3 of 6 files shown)


src/manifest/dbtProject.ts
```diff
@@ -308,10 +312,30 @@ export class DBTProject implements Disposable {
...
+        this.telemetry.sendTelemetryEvent(
+          "pythonBridgeCannotParseProjectUserError",
+          {
+            error: exc.exception.message,
+            adapter: this.adapterType,
+          },
+        );
+        return;
+      }
+      // if we get here, it
```
anandgupta42, line 331: The comment seems in-complete

src/manifest/dbtProject.ts
```diff
@@ -308,10 +312,30 @@ export class DBTProject implements Disposable {
...
       );
+      this.diagnostics.rebuildManifestDiagnostics.clear();
     } catch (exc) {
-      this.telemetry.sendTelemetryError("pythonBridgeCannotParseProject", exc, {
-        adapter: this.adapterType,
-      });
+      if (exc instanceof PythonException) {
+        this.diagnostics.rebuildManifestDiagnostics.set(
+          Uri.joinPath(this.projectRoot, DBTProject.DBT_PROJECT_FILE),
+          [new Diagnostic(new Range(0, 0, 999, 999), exc.exception.message)],
```
anandgupta42, line 320: It's not clear where the range is coming from. Is it to limit the message size? If yes, I am not sure if the range would capture the exact problem within it.  Also, adding a comment explaining the rationale behind the range will be helpful for posterity purposes.

src/diagnostics/index.ts
```diff
@@ -0,0 +1,12 @@
+import { Disposable, languages } from "vscode";
+import { provideSingleton } from "../utils";
+
+@provideSingleton(DbtPowerUserDiagnostics)
+export class DbtPowerUserDiagnostics implements Disposable {
+  public readonly rebuildManifestDiagnostics =
+    languages.createDiagnosticCollection("dbt");
```
anandgupta42, line 7: nit: the collection name "dbt" seems too generic . Not sure what the scope of this is and if it may lead to global collisions. If yes, then let's make it specific to our extension like "dpu-dbt"

dbt_integration.py
```diff
@@ -333,8 +333,9 @@ def safe_parse_project(self, reinit: bool = False) -> None:
         try:
             self.parse_project(init=reinit)
             self.write_manifest_artifact()
-        except Exception:
+        except Exception as e:
```
anandgupta42, line 336: nit: should we prefix "parser error" or something similar to give some hint from which operation the error was triggered?

anandgupta42: or may be not since the intention of the error is to show it in as a diagnostic message.

anandgupta42 (approved):
Minor feedback else LGTM.

src/manifest/dbtProject.ts
mdesmet: WE need to associate it with a file but that is hard in our case, so we take the project config.

I will add the comment.

src/diagnostics/index.ts
mdesmet: This is not how it works. the list is just an instance and need to be managed as such. the name just shows up as the source.

Clear would only impact that instance.

image

src/manifest/dbtProject.ts
mdesmet: Completed

dbt_integration.py
mdesmet: Indeed it is rather a diagnostic error. I think the other operations are more user actions and less about diagnostics

PR #576, opened 2023-09-10
Add enableGenerateModelCodeLens option.

yassun7010:
## Overview
It is not desirable for all users to have CodeLens for the Generate model in schema.yml.

My project has a policy of not creating views for sources, so this PR add a setting option to prevent members from clicking on CodeLens.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,8 @@ Telemetry is used for error and usage reporting in order to make the extension b
 
 You can select a file name template and prefix in the settings.
 
+If this feature is not desired, it can be disabled by setting "dbt.enableGenerateModelCodeLens" to false.
+
 ### <a id="doceditor">Edit the documentation from your model</a>
 
 ![Edit the documentation from your model](./media/images/documentation-editor.png)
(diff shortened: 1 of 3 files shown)


yassun7010:
Thanks for the advice!

I share your view point of not wanting to bring complexity into the configuration file.

My motivation for this PR is that I prefer to prohibit the cause of the problem rather than detecting the problem after it has occurred.

Also, current config schema already has the "prefixGenerateModel" and "fileNameTemplateGenerateModel", and I feel it has already somewhat complex.

A detection mechanism using linting/pre-commit and ci/cd checks would be possible by making/using features such as dry-run and diff.
However, I also feel that such a feature would lead to other complications.

I have thought about it with respect to the simplicity of this tool, and based on future extensibility, the following configuration may be better than adding "enableGenerateModelCodeLens" option.

```json
{
    "dbt.codeLensFeatures": ["GenerateModel"] // default
}
```

After all, It is important that the developer's policies are respected, so you may close this PR if you feel it does not fit the tool's policies.

anandgupta42:
Closing the pull request as it has been open for a while. This is part of our regular housekeeping process.

PR #586, opened 2023-09-17
Diagnostics improved

mdesmet:
## Checklist


src/manifest/dbtProject.ts
```diff
@@ -225,12 +233,12 @@ export class DBTProject implements Disposable {
...
         project: DBTProject.hashProjectRoot(this.projectRoot.fsPath),
       });
-      this.diagnostics.pythonBridgeDiagnostics.clear();
+      this.pythonBridgeDiagnostics.clear();
     } catch (exc: any) {
       if (exc instanceof PythonException) {
         // python errors can be about anything, so we just associate the error with the project file
         //  with a fixed range
-        this.diagnostics.rebuildManifestDiagnostics.set(
+        this.pythonBridgeDiagnostics.set(
```
lionsardesai, line 241: should this be this.rebuildManifestDiagnostics.set ?

mdesmet: No because it is linked to the exception thrown in that code block (it is about the python bridge init)

PR #592, opened 2023-09-19
feat: Download and Copy button added to the query results view

### LINKED ISSUE #591: Export button for query result
### Describe the feature

User should be able to copy the query result to the clipboard in CSV format and download it as CSV as well

Copy Pasting Data

### Who will benefit?

User who manually analyze the result will find this useful

anandgupta42:
## Overview

resolves #591 

## Checklist

### DIFF
--- a/query_panel/index.html
+++ b/query_panel/index.html
@@ -84,8 +84,12 @@
           />
 
           <div v-show="hasData">
-            <button @click="copyResultsToClipboard">Copy Results</button>
-            <button @click="downloadAsCSV">Download as CSV</button>
+            <vscode-button @click="copyResultsToClipboard"
+              >Copy Results</vscode-button
+            >
+            <vscode-button @click="downloadAsCSV"
+              >Download as CSV</vscode-button
+            >
             <div id="query-results" :style="getTableStyles()"></div>
           </div>
 
--- a/query_panel/index.html
+++ b/query_panel/index.html
@@ -1,108 +1,153 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
-
-<!-- These are injected at HTML render time
+  <!-- These are injected at HTML render time
     __ROOT__            [path to resource directory]
     __THEME__           ["light" | "dark"]
     __NONCE__           [security]
     __CSPSOURCE__       [webview Content Secure Policy source]
  -->
 
-<head>
+  <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 
-    <meta http-equiv="Content-Security-Policy"
-        content="default-src 'none'; img-src __CSPSOURCE__ https:; style-src __CSPSOURCE__ 'unsafe-inline'; script-src 'strict-dynamic' 'unsafe-eval' 'nonce-__NONCE__';" />
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="default-src 'none'; img-src __CSPSOURCE__ https:; style-src __CSPSOURCE__ 'unsafe-inline'; script-src 'strict-dynamic' 'unsafe-eval' 'nonce-__NONCE__';"
+    />
 
     <title>Query Runner</title>
 
     <!-- These are made available for main.js -->
     <script nonce="__NONCE__" type="importmap">
-        {
-          "imports": {
-            "vue": "__ROOT__/assets/vue.esm-browser.prod.js",
-            "tabulator": "__ROOT__/assets/tabulator.min.js",
-            "prism": "__ROOT__/assets/prism.js"
-          }
+      {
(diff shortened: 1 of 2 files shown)


query_panel/index.html
```diff
@@ -1,108 +1,154 @@
...
+          id="view-1"
+        >
+          <img
+            v-if="isLoading"
+            id="spinner"
+            src="__ROOT__/assets/spinner.gif"
+          />
+
+          <div v-show="hasData">
+            <button @click="copyResultsToClipboard">Copy Results</button>
```
mdesmet, line 87: Shouldn't this be vscode-button?

query_panel/main.js
```diff
@@ -34,6 +34,75 @@ const app = createApp({
...
+                  return `"${fieldData}"`; // Wrap in double quotes
+                }
+                return JSON.stringify(fieldData, replacer);
+              })
+              .join(","),
+          ),
+        ].join("\r\n");
+        return csv;
+      } catch (error) {
+        console.error("Error converting data to CSV:", error);
```
mdesmet, line 63: Shouldn't we add a telemetry error event?

query_panel/main.js
```diff
@@ -34,6 +34,75 @@ const app = createApp({
...
+    downloadAsCSV() {
+      const data = this.table.getData(); // Get the data the same way you do for copying
+      if (!data || data.length === 0) {
+        console.error("No data available for downloading.");
+        return;
+      }
+      let csvContent = this.dataToCsv(data);
+      let blob = new Blob([csvContent], { type: "text/csv" });
+      let url = window.URL.createObjectURL(blob);
+      let a = document.createElement("a");
```
mdesmet, line 76: all lets can be consts.

query_panel/main.js
```diff
@@ -34,6 +34,75 @@ const app = createApp({
...
+        return ""; // Return an empty string if there's an error
+      }
+    },
+    downloadAsCSV() {
+      const data = this.table.getData(); // Get the data the same way you do for copying
+      if (!data || data.length === 0) {
+        console.error("No data available for downloading.");
+        return;
+      }
+      let csvContent = this.dataToCsv(data);
```
mdesmet, line 73: This method can return empty string, why?

query_panel/main.js
```diff
@@ -34,6 +34,75 @@ const app = createApp({
...
+      let blob = new Blob([csvContent], { type: "text/csv" });
+      let url = window.URL.createObjectURL(blob);
+      let a = document.createElement("a");
+      a.href = url;
+      a.download = `power_user_data_${new Date().toISOString()}.csv`; // Filename with a timestamp
+      a.click();
+    },
+    copyTextToClipboard(text) {
+      const el = document.createElement("textarea");
```
mdesmet, line 82: Why use dynamic html and not use s standard vscode-textarea and vue data element.

query_panel/main.js
anandgupta42: it returns empty in case of error

mdesmet: I don't think exporting a empty file in case of an error is very useful. I would simply show an error message to the user that the file can't be generated with the error.

query_panel/main.js
```diff
@@ -31,9 +31,86 @@ const app = createApp({
...
+    async copyResultsToClipboard() {
+      try {
+        if (!this.table) {
+          throw new Error("Table is not initialized.");
+        }
+        const data = this.table.getData();
+        if (!data || data.length === 0) {
+          throw new Error("No data available for copying.");
+        }
+        const csv = this.dataToCsv(data);
```
lionsardesai, line 104: we dont need to dump csv when copying to clipboard

query_panel/main.js
```diff
@@ -31,9 +31,80 @@ const app = createApp({
...
+                  return `"${fieldData}"`; // Wrap in double quotes
+                }
+                return JSON.stringify(fieldData, replacer);
+              })
+              .join(","),
+          ),
+        ].join("\r\n");
+        return csv;
+      } catch (error) {
+        console.error("Error converting data to CSV:", error);
```
mdesmet, line 64: Can we display an error to the user here. Otherwise the user has to switch to developer tools console to see why the button doesn't work.

II would also get rid of the additional throw.

query_panel/main.js
```diff
@@ -31,9 +31,79 @@ const app = createApp({
...
+                }
+                return JSON.stringify(fieldData, replacer);
+              })
+              .join(","),
+          ),
+        ].join("\r\n");
+        return csv;
+      } catch (error) {
+        throw err;
+      }
```
mdesmet, line 65: This is no-op, please remove the error capture and throwing, as the error is capured already underneath.

mdesmet (approved):

PR #597, opened 2023-09-20
fix: column comparision issue

### LINKED ISSUE #595: vs-code-power-user : documentation editor case sensitivity issue with sync columns with database
### Expected behavior

The documentation editor "synchronize columns with database" feature should correctly match field names regardless of case-sensitivity. 

### Actual behavior

The synchronize columns with database feature in the documentation editor seems to be case-sensitive and causes issues if an existing yaml document with lower-case field names is synchronized, the doc editor ignores the existing content (description) and adds the field as UPPER as if it were new.

### Steps To Reproduce

create yaml model definition file with field and description

```
models:
  - name: int_table_subtotal
    description: subtotals
    columns:
      - name: myfield
        description: my awesome description
```

Use the documentation editor to edit this model.

When you re-sync the columns, the 'myfield' details will get wiped out and replaced with blank 'MYFIELD' column info.

### Log output/Screenshots

_No response_

### Operating System

windows 10

### dbt version

core 1.5.4

### dbt Adapter

snowflake 1.5.2

### dbt Power User version

v0.17.9

anandgupta42:
## Overview
resolves #595 

I have tested this on jaffle shop. Changed case of payments_id column of stg_payments model and it didn't clear the description as described in the issue.

## Checklist

### DIFF
--- a/docs_edit_panel/main.js
+++ b/docs_edit_panel/main.js
@@ -21,7 +21,8 @@ const app = createApp({
     updateColumns(columns) {
       this.docs.columns = columns.map((column) => {
         const existingColumn = this.docs?.columns.find(
-          (existingColumn) => column.name === existingColumn.name,
+          (existingColumn) =>
+            column.name.toLowerCase() === existingColumn.name.toLowerCase(),
         );
         return {
           name: column.name,


lionsardesai (commented):
does save also behave properly?

docs_edit_panel/main.js
```diff
@@ -21,7 +21,8 @@ const app = createApp({
     updateColumns(columns) {
       this.docs.columns = columns.map((column) => {
         const existingColumn = this.docs?.columns.find(
-          (existingColumn) => column.name === existingColumn.name,
+          (existingColumn) =>
+            column.name.toLowerCase() === existingColumn.name.toLowerCase(),
```
mdesmet, line 25: I think this should obey quoted identifier settings of dbt to work consistently.

For example A != "a" in snowflake.

mdesmet:
> does save also behave properly?

It will indeed be converted to the format that comes back from the database, generating potentially more changes in the yaml.

@anandgupta42 : WDYT?

anandgupta42:
> > does save also behave properly?
> 
> It will indeed be converted to the format that comes back from the database, generating potentially more changes in the yaml.
> 
> @anandgupta42 : WDYT?

AFA I understand, the best practice is for the docs to reflect the state of the database. In that case, does it not make sense to let the database be the source of the truth?

mdesmet (approved):

Issue #594, opened 2023-09-20
Parse additional manifest.json metadata for sources and macros


jsnb-devoted:
### Describe the feature

Once https://github.com/AltimateAI/vscode-dbt-power-user/pull/585 goes thru there will be hover capabilities for refs and sources. The `ref` hover box is rich with information because the `NodeParser` parses the descriptions/columns from the manifest. Sources and Macros only parse the paths so that go-to definition can be supported. If we get descriptions and columns for sources then those hover boxes can provide that metadata.

### Who will benefit?

This saves folks time of having to go-to the model and open the schema.yml file to see the descriptions/schema info for a source.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

mdesmet:
Hey @jsnb-devoted, can you have a look at this? This would be a really great addition and probably not difficult to add?

jsnb-devoted:
I guess I already did it! I’ll close the issue.
https://github.com/AltimateAI/vscode-dbt-power-user/pull/599

PR #599, opened 2023-09-23
Add descriptions and column metadata to the source metadata

jsnb-devoted:
## Overview
Resolves half of #594 

Parsing the descriptions and columns for sources and adding that info into the Hover provider for sources. I'm just pattern matching what is happening in the model parsing and it all lined up. 

Now the description and columns show up in the source hover box too:
Screenshot 2023-09-23 at 10 03 59 AM

Macros has similar metadata (descriptions and even function arguments/types -- see [macro properties dbt docs](https://docs.getdbt.com/reference/macro-properties)) but even the most popular packages like dbt_utils don't fill out those properties. I think adding that would be a good follow up. It might encourage package developers to actually fill out the properties.

## Checklist

### DIFF
--- a/src/hover_provider/sourceHoverProvider.ts
+++ b/src/hover_provider/sourceHoverProvider.ts
@@ -18,6 +18,7 @@ import { DBTProjectContainer } from "../manifest/dbtProjectContainer";
 import { ManifestCacheChangedEvent } from "../manifest/event/manifestCacheChangedEvent";
 import { isEnclosedWithinCodeBlock, provideSingleton } from "../utils";
 import { TelemetryService } from "../telemetry";
+import { GenerateHoverMarkdownString } from "./utils";
 
 @provideSingleton(SourceHoverProvider)
 export class SourceHoverProvider implements HoverProvider, Disposable {
@@ -117,38 +118,7 @@ export class SourceHoverProvider implements HoverProvider, Disposable {
       .get(sourceName)
       ?.tables.find((table) => table.name === tableName);
     if (node) {
-      const content = new MarkdownString();
-      content.supportHtml = true;
-      content.isTrusted = true;
-      content.appendMarkdown(
-        `<span style="color:#347890;">(source)&nbsp;</span><span><strong>${node.name}</strong></span>`,
-      );
-      if (node.description !== "") {
-        content.appendMarkdown(`</br><span>${node.description}</span>`);
-      }
-      content.appendText("\n");
-      content.appendText("\n");
-      content.appendMarkdown("---");
-      content.appendText("\n");
-      content.appendText("\n");
-      for (const colKey in node.columns) {
-        const column = node.columns[colKey];
-        content.appendMarkdown(
-          `<span style="color:#347890;">(column)&nbsp;</span><span>${column.name} &nbsp;</span>`,
-        );
-        if (column.data_type !== null) {
-          content.appendMarkdown(
-            `<span>-&nbsp;${column.data_type.toUpperCase()}</span>`,
-          );
-        }
-        if (column.description !== "") {
-          content.appendMarkdown(
-            `<br/><span><em>${column.description}</em></span>`,
-          );
-        }
-        content.appendMarkdown("</br>");
-      }
-      return content;
+      return GenerateHoverMarkdownString(node, "source");
     }
     return undefined;
   }
--- a/src/hover_provider/sourceHoverProvider.ts
+++ b/src/hover_provider/sourceHoverProvider.ts
@@ -123,6 +123,31 @@ export class SourceHoverProvider implements HoverProvider, Disposable {
       content.appendMarkdown(
         `<span style="color:#347890;">(source)&nbsp;</span><span><strong>${node.name}</strong></span>`,
       );
+      if (node.description !== "") {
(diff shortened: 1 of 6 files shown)


src/hover_provider/sourceHoverProvider.ts
```diff
@@ -123,6 +123,31 @@ export class SourceHoverProvider implements HoverProvider, Disposable {
...
+      content.appendMarkdown("---");
+      content.appendText("\n");
+      content.appendText("\n");
+      for (const colKey in node.columns) {
+        const column = node.columns[colKey];
+        content.appendMarkdown(
+          `<span style="color:#347890;">(column)&nbsp;</span><span>${column.name} &nbsp;</span>`,
+        );
+        if (column.data_type !== null) {
+          content.appendMarkdown(
```
anandgupta42, line 140: I think we can consider refactoring the main function to be more readable and maintainable. Something in the lines of :

`private createTableContent(node: any): MarkdownString {
  const content = new MarkdownString();
  content.appendMarkdown(
    `<span style="color:#347890;">(source)&nbsp;</span><span><strong>${node.name}</strong></span>`,
  );
  if (node.description !== "") {
    content.appendMarkdown(`</br><span>${node.description}</span>`);
  }
  content.appendText("\n\n");
  content.appendMarkdown("---");
  content.appendText("\n\n");
  return content;
}

private appendColumnContent(content: MarkdownString, column: any) {
  content.appendMarkdown(
    `<span style="color:#347890;">(column)&nbsp;</span><span>${column.name} &nbsp;</span>`,
  );
  if (column.data_type !== null) {
    content.appendMarkdown(`<span>-&nbsp;${column.data_type.toUpperCase()}</span>`);
  }
  if (column.description !== "") {
    content.appendMarkdown(`<br/><span><em>${column.description}</em></span>`);
  }
  content.appendMarkdown("</br>");
}

private getSourceHover(
  sourceName: string,
  currentFilePath: Uri,
  tableName: string,
): MarkdownString | undefined {
  const projectRootpath = this.dbtProjectContainer.getProjectRootpath(currentFilePath);
  if (!projectRootpath) return;

  const sourceMap = this.sourceMetaMap.get(projectRootpath.fsPath);
  if (!sourceMap) return;

  const node = sourceMap.get(sourceName)?.tables.find(table => table.name === tableName);
  if (!node) return;

  const content = this.createTableContent(node);

  for (const colKey in node.columns) {
    const column = node.columns[colKey];
    this.appendColumnContent(content, column);
  }

  return content;
}
`

Note this is a sample code just to give you some idea of what I am thinking

src/hover_provider/sourceHoverProvider.ts
```diff
@@ -123,6 +123,31 @@ export class SourceHoverProvider implements HoverProvider, Disposable {
       content.appendMarkdown(
         `<span style="color:#347890;">(source)&nbsp;</span><span><strong>${node.name}</strong></span>`,
       );
+      if (node.description !== "") {
+        content.appendMarkdown(`</br><span>${node.description}</span>`);
+      }
+      content.appendText("\n");
```
mdesmet, line 129: We are adding a lot of new lines, however if either description or columns are not present, this may not be presented very beautifully. 

Could you ensre that the lines are only added if the relevant section is present?

mdesmet (commented):
LGTM

src/hover_provider/sourceHoverProvider.ts
mdesmet: As this issue is also present in the model code. Maybe you can refactor this out to a separate function and use this function in both places?

jsnb-devoted: The newlines actually don't render as newlines the way it looks. They are just there so that the `---` is shown as a line. Like this:

---

When the descriptions or columns aren't present the hover box looks like:
Screenshot 2023-09-26 at 8 36 13 AM
I can still remove it if columns aren't present. It is straight forward enough to change.

In any case I can definitely consolidate the code into a single function before getting this merged. They are nearly identical.

mdesmet: Great! Thanks for clarifying.

src/hover_provider/utils.ts
```diff
@@ -0,0 +1,40 @@
+import { MarkdownString } from "vscode";
+import { NodeMetaType, SourceMetaType } from "../domain";
+
+export function GenerateHoverMarkdownString(
```
jsnb-devoted, line 4: @mdesmet I am not a javascript developer so keep me honest here. Not sure if the union typing here is a best practice but this is one function doing the markdown for models and source.

@anandgupta42 you want this further refactored to separate the column functionality in a separate function?

mdesmet, line 4: functions should start with lowercase character

mdesmet (commented):
Thanks, looks fine to me. Just a small remark on lowercasing the function.

src/hover_provider/utils.ts
anandgupta42: @jsnb-devoted this looks good for now

anandgupta42 (approved):

mdesmet (approved):

PR #611, opened 2023-09-28
Adding an allow list configuration option

Ben-A-Canva:
## Overview
Adds an optional folder allowlist configuration.

## Motivation
Introduce an 'allowlist' feature to enable users to specify project folders explicitly, ensuring that only projects within these folders are included in the build process. This provides greater control over project compilation and improves workspace performance by preventing the accidental compilation of unrelated projects.

## Issue
With our monorepo structure, we are unable to use the extension because it attempts to spin up several hundred projects at once causing device OOM's and reboots. Adding an allowlist will help us to specifically target the projects we work on.

## Checklist

### DIFF
--- a/src/manifest/dbtWorkspaceFolder.ts
+++ b/src/manifest/dbtWorkspaceFolder.ts
@@ -65,7 +65,7 @@ export class DBTWorkspaceFolder implements Disposable {
       )
       // TODO: also filter out projects within the target folder of another project
       //  This is somewhat difficult as we would need to parse the target-path variable of the project.
-      .map((uri) => Uri.file(uri.path.split("/")!.slice(0, -1).join("/")))      ;
+      .map((uri) => Uri.file(uri.path.split("/")!.slice(0, -1).join("/")));
     if (projectFiles.length > 20) {
       window.showWarningMessage(
         `dbt Power User detected ${projectFiles.length} projects in your work space, this will negatively affect performance.`,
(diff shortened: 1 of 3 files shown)


src/manifest/dbtWorkspaceFolder.ts
```diff
@@ -52,12 +52,25 @@ export class DBTWorkspaceFolder implements Disposable {
         `**/{${DBTProject.DBT_MODULES.join(",")}}`,
       ),
     );
-    const projectFiles = dbtProjectFiles
+    var projectFiles = dbtProjectFiles
```
mdesmet, line 55: Please never do `var`, rather keep on using `const` and apply the filter directly onto dbtProjectFiles inlining the same logic.

mdesmet (commented):
Please also add a motivation on why this change is needed, ie what problems does it solve.

src/manifest/dbtWorkspaceFolder.ts
Ben-A-Canva: yup, good callout. I'm not used to typescript 😁

anandgupta42:
LGTM. @Ben-A-Canva, Can you please update the readme also?

mdesmet (approved):
LGTM, but please add docs as @anandgupta42 asked.

anandgupta42 (approved):

anandgupta42:
LGTM @mdesmet the doc has been added. We can merge this.

Ben-A-Canva:
> LGTM, but please add docs as @anandgupta42 asked.

Added an entry to the readme file. Is there anything else I need to update for docs? 👀

PR #614, opened 2023-09-29
Adding perspective library in query panel

AdiGajbhiye:
## Overview
Adding perspective library in query panel
## Checklist


anandgupta42:
-export is not working
- the legacy UI button has to be moved to the right

mdesmet:
@AdiGajbhiye : is there any way to make the height adapt to the evaiable height of the panel?

mdesmet:
It seems that the on click handler is removed when hiding the perspective sidebar.

mdesmet (approved):
Nice work!

I like the native export integration. I don't like the way it is implemented but I validated that there is no override mechanism in perspective, so we are good to go.

PR #633, opened 2023-10-09
feat: Plain sql to model

lionsardesai:
## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -262,10 +262,7 @@
       {
         "command": "dbtPowerUser.sqlToModel",
         "title": "Associate model references",
-        "icon": {
-          "light": "./media/images/run-light.svg",
-          "dark": "./media/images/run-dark.svg"
-        }
+        "icon": "$(circuit-board)"
       },
       {
         "command": "dbtPowerUser.compileCurrentModel",
--- a/package.json
+++ b/package.json
@@ -261,7 +261,7 @@
       },
       {
         "command": "dbtPowerUser.sqlToModel",
-        "title": "Convert to Model",
+        "title": "Convert to dbt Model",
         "icon": "$(circuit-board)"
       },
       {
--- a/package.json
+++ b/package.json
@@ -383,7 +383,7 @@
         },
         {
           "command": "dbtPowerUser.sqlToModel",
-          "when": "resourceLangId =~ /^sql$|^jinja-sql$/",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -367,30 +372,35 @@
           "group": "navigation@3"
         },
         {
-          "command": "dbtPowerUser.testCurrentModel",
+          "command": "dbtPowerUser.sqlToModel",
```
mdesmet, line 375: move last as this is probably least used navigation item.

src/commands/sqlToModel.ts
```diff
@@ -0,0 +1,84 @@
...
+    if (event === undefined) {
+      return undefined;
+    }
+    const rriter = event.nodeMetaMap.values();
+    for (const rrvalue of rriter) {
+      allmodels.push(rrvalue);
+    }
+
+    const fileText = activedoc.document.getText();
+    const compiled_sql = await project.compileQuery(fileText);
```
mdesmet, line 60: Exceptions are handled in this method but please verify this.

src/commands/sqlToModel.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    const retobj = await this.altimate.runModeller({
```
mdesmet, line 62: Add exception handling

src/commands/sqlToModel.ts
```diff
@@ -0,0 +1,84 @@
...
+    const allmodels: NodeMetaData[] = [];
+    const project = this.dbtProjectContainer.findDBTProject(currentFilePath);
+    if (project === undefined) {
+      return undefined;
+    }
+    // uncomment this line in case you need to pass everything
+    // const reciter = this.eventMap.values();
+
+    // for now, restrict it to this project only.
+    // assuming user has narrowed down to a project.
```
mdesmet, line 49: You can remove above comment as that would not work anyway.

src/commands/sqlToModel.ts
lionsardesai: double checked. exceptions caught and error messages raised.

src/commands/sqlToModel.ts
lionsardesai: added catch block

package.json
lionsardesai: moved to last.

mdesmet (approved):

PR #638, opened 2023-10-10
Pass changed document text to sqlfmt through stdin instead of file path

ljb7977:
## Overview

Slack: https://getdbt.slack.com/archives/C05KPDGRMDW/p1696953977216509

### Problem
With the formatOnSave option turned on, the first time I pressed save the file it doesn't get formatted but it does in the second save.

### Cause
The existing implementation passes the file path to `sqlfmt` rather than the changed content. Since the file is not yet saved when I hit the first save, `sqlfmt` reads the previous version of the file and no formatting occurs, and the unformatted new content is saved to the file. And when I hit the second save, finally the file gets formatted because `sqlfmt` now can read the unformatted content from the file.

### Solution
Pass modified content of the file to `sqlfmt` instead of the file path
- CommandProcessExecution.complete is extended to get a string and pass it into spawned process as stdin.

## Checklist

### DIFF
--- a/src/commandProcessExecution.ts
+++ b/src/commandProcessExecution.ts
@@ -52,7 +52,7 @@ export class CommandProcessExecution {
     }
   }
 
-  async complete(): Promise<string> {
+  async complete(stdin?: string): Promise<string> {
     return new Promise<string>((resolve, reject) => {
       const commandProcess = this.spawn();
       let stdoutBuffer = "";
@@ -78,6 +78,11 @@ export class CommandProcessExecution {
         console.warn(error);
         reject(`${error}`);
       });
+
+      if (stdin) {
+        commandProcess.stdin.write(stdin);
+        commandProcess.stdin.end();
+      }
     });
   }
 
--- a/src/commandProcessExecution.ts
+++ b/src/commandProcessExecution.ts
@@ -6,14 +6,29 @@ import { EnvironmentVariables } from "./domain";
 
 @provide(CommandProcessExecutionFactory)
 export class CommandProcessExecutionFactory {
-  createCommandProcessExecution(
-    command: string,
-    args?: string[],
-    cwd?: string,
-    token?: CancellationToken,
-    envVars?: EnvironmentVariables,
-  ) {
-    return new CommandProcessExecution(command, args, cwd, token, envVars);
+  createCommandProcessExecution({
+    command,
+    args,
+    stdin,
+    cwd,
+    token,
+    envVars,
+  }: {
+    command: string;
+    args?: string[];
+    stdin?: string;
+    cwd?: string;
+    token?: CancellationToken;
+    envVars?: EnvironmentVariables;
+  }) {
+    return new CommandProcessExecution(
+      command,
+      args,
+      stdin,
+      cwd,
+      token,
+      envVars,
+    );
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/document_formatting_edit_provider/dbtDocumentFormattingEditProvider.ts
```diff
@@ -27,12 +30,25 @@ export class DbtDocumentFormattingEditProvider
...
+
+    // channel.appendLine(">----- file content -----<");
+    // channel.appendLine("path: " + document.uri.fsPath);
+    // const content = readFileSync(document.uri.fsPath);
+    // channel.appendLine("file content");
+    // channel.appendLine(content.toString());
+
+    const edit = await this.executeSqlFmt(document);
+    // channel.appendLine(JSON.stringify(edit));
+    // channel.appendLine("\n\n\n");
```
ljb7977, line 50: These comments will be removed later. I left them temporarily for debugging

mdesmet:
Can you add why you want to make this change? What problem are we trying to solve here?

src/document_formatting_edit_provider/dbtDocumentFormattingEditProvider.ts
```diff
@@ -52,11 +68,11 @@ export class DbtDocumentFormattingEditProvider
       .filter((s) => s !== "");
 
     const sqlFmtArgs = [
+      "-",
```
mdesmet, line 71: Why we need this dash?

mdesmet (commented):
LGTM, please remove the  comments and the output channel.

src/document_formatting_edit_provider/dbtDocumentFormattingEditProvider.ts
ljb7977: This is an argument to input stdin to sqlfmt. ([ref](https://docs.sqlfmt.com/getting-started/using-sqlfmt#:~:text=sqlfmt%20can%20also%20format%20code%20passed%20through%20standard%20input%20(stdin)%20by%20passing%20%2D%20as%20the%20files%20argument.%20The%20formatted%20code%20will%20be%20output%20to%20stdout%20(all%20other%20output%20from%20sqlfmt%20is%20routed%20to%20stderr)%3A)). Instead of passing document.uri.fsPath to sqlfmt, putting this here enables stdin input to sqlfmt and the .complete function passes document's content to sqlfmt through stdin.

ljb7977:
Removed the comment lines and updated PR description :)

src/commandProcessExecution.ts
```diff
@@ -52,7 +52,7 @@ export class CommandProcessExecution {
     }
   }
 
-  async complete(): Promise<string> {
+  async complete(stdin?: string): Promise<string> {
```
mdesmet, line 55: Maybe more future proof would be to pass this to the `constructor` instead. That way we can also use it in `spawn` method if necessary.

src/commandProcessExecution.ts
```diff
@@ -6,14 +6,29 @@ import { EnvironmentVariables } from "./domain";
...
+    token,
+    envVars,
+  }: {
+    command: string;
+    args?: string[];
+    stdin?: string;
+    cwd?: string;
+    token?: CancellationToken;
+    envVars?: EnvironmentVariables;
+  }) {
```
ljb7977, line 23: Refactored here to get those arguments as object since most of them are optional.

src/commandProcessExecution.ts
ljb7977: Sounds good. I pushed a commit for that!

PR #640, opened 2023-10-11
Feat: level wise gradual cll loading

AdiGajbhiye:
## Overview
Feat: level wise gradual cll loading
## Checklist


mdesmet:
The progress indicators are now duplicated. That seems confusing to the user, can we get rid of the duplication?

mdesmet (approved):

PR #641, opened 2023-10-11
Fix: no rows in query panel

AdiGajbhiye:
## Checklist

### DIFF
--- a/query_panel/main.js
+++ b/query_panel/main.js
@@ -72,7 +72,7 @@ const app = createApp({
   data() {
     return {
       count: 0,
-      data: null,
+      cacheData: null,
       rawCode: "",
       compiledCode: "",
       error: {},
--- a/query_panel/main.js
+++ b/query_panel/main.js
@@ -51,7 +51,7 @@ class Grid {
     }
     console.log("this is load -> ", schema, result);
     const table = await this.worker.table(schema);
-    await table.update(result.rows);
+    await table.replace(result.rows);
     await this.elem.load(table);
     await this.elem.restore({
       columns: [], // reset columns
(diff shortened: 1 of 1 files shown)


query_panel/main.js
```diff
@@ -16,16 +16,26 @@ async function updateConfig(config) {
 const DEFAULT_HEIGHT = 455;
 
 class Grid {
+  table;
```
mdesmet, line 19: `constructor(private table)`?

query_panel/main.js
```diff
@@ -43,8 +53,7 @@ window.addEventListener("DOMContentLoaded", async function () {
 const app = createApp({
   data() {
     return {
-      count: 0,
-      data: null,
+      eventData: null,
```
mdesmet, line 56: We have been changing the name of this field lots of times. I'm not sure if the new name is better, why not keep the name data?

query_panel/main.js
mdesmet: Why do we need to even save the table reference?

query_panel/main.js
```diff
@@ -50,9 +50,10 @@ class Grid {
       schema[result.columnNames[i]] = this.mapType(result.columnTypes[i]);
     }
     const table = await this.worker.table(schema);
-    table.replace(result.rows);
+    await table.replace(JSON.parse(JSON.stringify(result.rows)));
```
lionsardesai, line 53: why was this needed? should we be catching anything if this fails?

AdiGajbhiye:
closing in favor of https://github.com/AltimateAI/vscode-dbt-power-user/pull/657

PR #644, opened 2023-10-12
Add schema to perspective

mdesmet:
## Checklist

### DIFF
--- a/query_panel/main.js
+++ b/query_panel/main.js
@@ -24,8 +24,33 @@ class Grid {
     this.elem = document.querySelector("perspective-viewer");
   }
 
-  async load(rows) {
-    const table = await this.worker.table(rows);
+  mapType(agateType) {
+    switch (agateType) {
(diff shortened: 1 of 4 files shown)


anandgupta42 (approved):
Code makes sense. I have not tested it.

lionsardesai (approved):

query_panel/main.js
```diff
@@ -24,8 +24,33 @@ class Grid {
...
+      case "Text":
+        return "string";
+      case "Integer":
+        return "integer";
+      case "Boolean":
+        return "boolean";
+      case "Date":
+        return "date";
+      case "DateTime":
+        return "datetime";
```
lionsardesai, line 38: for these can we map to strings?

mdesmet: I think it is not necessary. I think datetime is OK to show in user timezone. date is not shown in user timezone.

PR #660, opened 2023-10-19
fix: connection is closed errors

### LINKED ISSUE #648: dbt_utils.union_relations macro will not run.
### Expected behavior

Model will run when executed via dbt Power User.

### Actual behavior

An error is thrown when the model is run.  
`An error occured while trying to execute your query: Runtime Error 250002 (08003): Connection is closed`

### Steps To Reproduce

I click ctrl-Enter as I do with all other queries.  They run as expected.  A model that includes dbt_util.union_relations throws the error indicated above.  This was not happening before the recent update to dbt Power User.

### Log output/Screenshots

`An error occured while trying to execute your query: Runtime Error 250002 (08003): Connection is closed`

### Operating System

Windows 10

### dbt version

1.6.0

### dbt Adapter

Snowflake

### dbt Power User version

v0.21.1

lionsardesai:
should resolve #648.
nested named connections were opened with same name and potentially get closed before the outer context has finished execution. Alternate is having different named connections but better to prevent nesting since the inner calls can reuse the outer connection. This should fix the connection is closed errors we are seeing.

## Checklist

### DIFF
--- a/dbt_integration.py
+++ b/dbt_integration.py
@@ -83,7 +83,7 @@ def to_dict(obj):
         return {
             "rows": [to_dict(row) for row in obj.rows],
             "column_names": obj.column_names,
-            "column_types": list(map(lambda x: x.__class__.__name__, obj.column_types))
+            "column_types": list(map(lambda x: x.__class__.__name__, obj.column_types)),
         }
     if isinstance(obj, str):
         return obj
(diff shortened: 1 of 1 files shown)


mdesmet (approved):

pkanter:
Thank you for the fix!

PR #674, opened 2023-10-24
feat: Bulk get schema + Altimate Scan

lionsardesai:
add logic to get full db catalog.

TODO - currently assuming all projects look at disjoint catalogs. There might be some way to optimize the calls but not sure if i can get that info from just hostname/db name etc.

Note - using get_catalog because it is supported in older versions as well. the newer version lets you get only certain schemas. This might be a little too heavy and we are relying on adapter-level optimizations (eg snowflake is restricting to the information_schema of that database). However, this is the same call that `dbt docs generate` makes so it wont be slower than that call. It does avoid all the other stuff that generate does so ideally this should be much faster.

Additionally this PR also builds on the catalog and has the starting pieces of the deep scan aka altimate scan.

## Checklist


package.json
```diff
@@ -264,6 +264,11 @@
         "title": "Convert to dbt Model",
         "icon": "$(circuit-board)"
       },
+      {
+        "command": "dbtPowerUser.altimateScan",
+        "title": "Deep Scan",
```
anandgupta42, line 269: Should this not be Altimate Scan?

src/commands/command_utils.ts
```diff
@@ -0,0 +1,93 @@
+import { Uri, Diagnostic, Range, DiagnosticSeverity } from "vscode";
+import { NodeMetaData } from "../domain";
+import { DBTProject } from "../manifest/dbtProject";
+
+export function findModelProblems(
```
mdesmet, line 5: This is not scalable. This should go in the respective class that does the documentation checks.

src/commands/altimateScan.ts
```diff
@@ -0,0 +1,150 @@
...
+        project,
+        projectEventMap,
+        projectDiagnostics,
+      );
+      for (const [filePath, fileDiagnostics] of Object.entries(
+        projectDiagnostics,
+      )) {
+        project.projectHealth.set(Uri.file(filePath), fileDiagnostics);
+      }
+    }
```
mdesmet, line 58: This logic should go in a separate class. As we are going to add more checks (probably ran in parallel).

src/commands/command_utils.ts
```diff
@@ -0,0 +1,93 @@
...
+        allDiagnostic[schemaPath] = schemaDiagnostics = [];
+      }
+      schemaDiagnostics.push(
+        new Diagnostic(
+          new Range(0, 0, 0, 0),
+          err_message,
+          DiagnosticSeverity.Warning,
+        ),
+      );
+      console.log(err_message);
```
mdesmet, line 90: always use camel casing in javascript.

src/commands/command_utils.ts
```diff
@@ -0,0 +1,93 @@
...
+        name +
+        "\n" +
+        "The column named `" +
+        existingCol +
+        "` is documented but might be out of sync with the existing model.";
+      // If we are here, the patch_path is guaranteed to be defined since
+      // we catch missing doc errors before we enter this function.
+      const schemaPath = Uri.joinPath(
+        project.projectRoot,
+        projectModel.patch_path.split("://")[1],
```
mdesmet, line 77: Will this work on Windows. Please check other occurences of this pattern?

src/commands/command_utils.ts
```diff
@@ -0,0 +1,93 @@
...
+  for (const column of modelDict) {
+    if (!existingColumnsLowered.includes(column.column_name.toLowerCase())) {
+      const err_message =
+        "Column undocumented for model: " +
+        name +
+        "\n" +
+        "Expected an entry for: " +
+        column.column_name;
+
+      // TODO - this should point to the model file ?
```
mdesmet, line 48: Please point this indeed to the model file. Should be doable based on the manifest.

dbt_integration.py
```diff
@@ -59,6 +61,17 @@
     from dbt.adapters.base import BaseRelation  # type: ignore
     from dbt.contracts.connection import AdapterResponse
 
+import logging
+
+# Set up logging to file
+logging.basicConfig(filename="python_log.log", level=logging.DEBUG)
+logger = logging.getLogger(__name__)
```
mdesmet, line 68: Are we going to introduce the logger here?

I think we shouldn't log by default.

dbt_integration.py
```diff
@@ -558,6 +571,38 @@ def get_columns(self, node: "ManifestNode") -> List["ColumnInfo"]:
...
+    def get_catalog(self) -> Dict[str, Any]:
+        """Get catalog from adapter"""
+        catalog_table: agate.Table = agate.Table([])
+        catalog_data: List[PrimitiveDict] = []
+        exceptions: List[Exception] = []
+        logger.debug("getting connection")
+        try:
+            with self.adapter.connection_named("generate_catalog"):
+                logger.debug("Generating catalog...")
+                catalog_table, exceptions = self.adapter.get_catalog(self.dbt)
```
mdesmet, line 583: What about a model that refers to another database?

dbt_integration.py
```diff
@@ -558,6 +571,38 @@ def get_columns(self, node: "ManifestNode") -> List["ColumnInfo"]:
...
+                )
+                for row in catalog_table
+            ]
+            logger.debug("parsed catalog data")
+            logger.debug(catalog_data)
+
+        except Exception as e:
+            logger.debug("exception during get catalog event")
+            logger.debug(e)
+            return catalog_data
```
mdesmet, line 603: Shouldn't we expose the exception as a string to the user?

dbt_integration.py
lionsardesai: No this is only until i finish debugging. this should be gone when its ready. keeping the comment as a todo

src/commands/command_utils.ts
lionsardesai: This pattern is added by our parser (i think its of the format <package>://<path>) so it should be independent of os. Also being done in doc editor.

dbt_integration.py
lionsardesai: if docs generate works, this should work too. adapter's get_catalog is what is called for docs generate.

README.md
```diff
@@ -237,3 +239,7 @@ This feature relies on a defined schema in a YML file. It extracts data from bot
 ### <a id="logtailer">dbt logs force tailing</a>
 
 ![dbt logs](./media/images/dbt-log.gif)
+
+### <a id="projectscan">Altimate scan</a>
```
anandgupta42, line 243: We should some details about the scans performed and what each of the scan mean

anandgupta42:
Overall looks good. There are a few things that we need to take care

- report the list of files skipped during the scan due to parsing or any other errors [ in future we should extend it to show a complete report]
- send out count of the errors found per type and also the count of the number of files skipped

anandgupta42:
Can you either imrpove the tool tip to talk about what the scan does or add a line in the panel itself to explain what it does?

Screenshot 2023-10-28 at 11 30 22 AM

src/commands/tests/initCatalog.ts
```diff
@@ -0,0 +1,41 @@
+import { DBTProject } from "../../manifest/dbtProject";
+import { AltimateScanAgent } from "../agent/agent";
+import { BaseAltimateScanAgent } from "../agent/baseAltimateScanAgent";
+import { AltimateScanStep } from "./step";
+
+export class InitCatalog implements AltimateScanStep {
+  async run(agent: AltimateScanAgent) {
+    // This one has to await
+    // CAUTION - not sure why this isnt giving me circular dependency warning.
+    await (agent as BaseAltimateScanAgent).runStep(this);
```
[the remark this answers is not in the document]
lionsardesai: i only want to await the init catalog step. the tests can all run in parallel. so i am not awaiting all the runs except this one. may be there is a better design for this?

src/commands/altimateScan.ts
```diff
@@ -0,0 +1,115 @@
...
+    this.telemetry.sendTelemetryEvent("altimateScan:Start");
+    let totalProblems: number = 0;
+    window.withProgress(
+      {
+        location: ProgressLocation.Notification,
+        title: "Scanning for problems...",
+        cancellable: false,
+      },
+      async () => {
+        const projects = this.dbtProjectContainer.findAllDBTProjects();
```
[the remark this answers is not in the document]
lionsardesai: There is no project selection yet. all projects loaded are scanned. should we restrict it by project?

src/commands/command_utils.ts
lionsardesai: double checked. also updated this with create file path that is used inside our manifest parsers

dbt_integration.py
lionsardesai: this is removed

dbt_integration.py
lionsardesai: TODO for a future page where we can select a project before scanning.

src/manifest/dbtProject.ts
```diff
@@ -600,6 +601,51 @@ export class DBTProject implements Disposable {
     );
   }
 
+  async getCatalog(): Promise<{ [key: string]: string }[]> {
+    await this.blockUntilPythonBridgeIsInitalized();
+    if (!this.pythonBridgeInitialized) {
+      window.showErrorMessage(
+        "Could not execute query, because the Python bridge has not been initalized. If the issue persists, please open a Github issue.",
```
anandgupta42, line 608: @lionsardesai Can you use the error message with the slack link instead?

src/commands/tests/initCatalog.ts
```diff
@@ -0,0 +1,41 @@
...
+        Object.create(null),
+      );
+      return modelDict;
+    } catch (err) {
+      console.log(
+        "Error in getting schema for project: " + project.getProjectName(),
+        err,
+      );
+      console.log(err);
+      return {};
```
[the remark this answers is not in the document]
lionsardesai: done. exceptions are now caught in the dbtproject.ts file and error message shown immediately. it returns empty catalog so all this should not be needed.

src/commands/tests/initCatalog.ts
```diff
@@ -0,0 +1,41 @@
...
+    // CAUTION - not sure why this isnt giving me circular dependency warning.
+    await (agent as BaseAltimateScanAgent).runStep(this);
+  }
+
+  public async getProjectCatalog(project: DBTProject) {
+    try {
+      const cata = await project.getCatalog();
+      const modelDict: { [key: string]: any[] } = cata.reduce(
+        (mdict: { [key: string]: any[] }, model) => {
+          const modelKey: string = JSON.stringify({
```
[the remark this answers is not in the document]
lionsardesai: i need a string value to use as key so using stringify so i can unwrap later.

src/commands/altimateScan.ts
```diff
@@ -0,0 +1,115 @@
...
+// online tests rely on a connection to database
+const onlineAltimateScanSteps = [
+  // check for unmaterialized models
+  new UnmaterializedModelTest(),
+  // then check for undocumented models columns
+  new UndocumentedModelColumnTest(),
+  // then check for stale columns in models
+  new StaleModelColumnTest(),
+  // feel free to add more tests
+];
```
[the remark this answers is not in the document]
lionsardesai: thsi should be done.

src/commands/tests/step.ts
```diff
@@ -0,0 +1,5 @@
+import { AltimateScanAgent } from "../agent/agent";
+
+export interface AltimateScanStep {
+  run(agent: AltimateScanAgent): void;
```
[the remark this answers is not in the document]
lionsardesai: removing as outdated. not using agent.

src/commands/altimateScan.ts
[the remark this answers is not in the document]
lionsardesai: closing after discussion. this is a TODO for future.

src/commands/tests/initCatalog.ts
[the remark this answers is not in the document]
lionsardesai: closing as outdated.

README.md
lionsardesai: done

mdesmet (approved):

PR #686, opened 2023-10-31
feat: Query explanation

lionsardesai:
add query explanation UI

## Checklist

### DIFF
--- a/src/commands/runModel.ts
+++ b/src/commands/runModel.ts
@@ -53,7 +53,7 @@ export class RunModel {
     }
   }
 
-  executeQueryOnActiveWindow() {
+  _getQueryWithName() {
     if (!window.activeTextEditor) {
       return;
     }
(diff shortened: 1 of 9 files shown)


src/webview_provider/queryResultPanel.ts
```diff
@@ -85,10 +98,12 @@ export class QueryResultPanel implements WebviewViewProvider {
 
   private _disposables: Disposable[] = [];
   private _panel: WebviewView | undefined;
+  private adapter: string = "";
```
mdesmet, line 101: empty string is not really clean. Should be rather undefined.

src/manifest/dbtProject.ts
```diff
@@ -552,7 +552,7 @@ export class DBTProject implements Disposable {
           "Encountered an unknown issue: " + exc + ".",
         ),
       );
-      return "Detailed error information:\n" + exc;
+      return "Exception:\n\nDetailed error information:\n" + exc;
```
mdesmet, line 555: Why this change?

src/commands/runModel.ts
```diff
@@ -53,7 +53,7 @@ export class RunModel {
     }
   }
 
-  executeQueryOnActiveWindow() {
+  _getQueryWithName() {
```
mdesmet, line 56: why not use a private method?

lionsardesai: done.. python habits.

src/commands/runModel.ts
```diff
@@ -67,7 +67,25 @@ export class RunModel {
...
+    const { query, queryName } = ret;
+    this.executeSQL(window.activeTextEditor!.document.uri, query, queryName);
+  }
+
+  getSummaryOnActiveWindow() {
+    const ret = this._getQueryWithName();
+    if (ret === undefined) {
+      return;
+    }
+    const { query, queryName } = ret;
```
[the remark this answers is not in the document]
lionsardesai: removed

src/webview_provider/queryResultPanel.ts
lionsardesai: setting it to unknown to match dbtProject.ts

src/manifest/dbtProject.ts
lionsardesai: Adding this because this function still returns a string after catching an exception. All exceptions have Exception: in the beginning so matching those. if string starts with that, i'm not generating summary for it.

lionsardesai: returning undefined if exception encountered after discussion with michiel.

mdesmet (approved):

PR #706, opened 2023-11-09
fix: wait for projects to load.

lionsardesai:
wait for projects to load before populating walkthrough items.

## Checklist


src/dbtPowerUserExtension.ts
```diff
@@ -82,8 +82,8 @@ export class DBTPowerUserExtension implements Disposable {
 
   async activate(context: ExtensionContext): Promise<void> {
     this.dbtProjectContainer.setContext(context);
-    this.dbtProjectContainer.initializeWalkthrough();
```
mdesmet, line 85: Should be fixed so it works dynamically, ie when projects are being added/removed

anandgupta42 (approved):

Issue #601, opened 2023-09-25
Unable to change target environment for query development purpose


VijaybabuNakkonda:
### Expected behavior

In Dev:
profiles.yml: I have profiles.yml target set to dev
models: a random model file with query select current_database()
result: Database_Dev
In QA:
profiles.yml: I have profiles.yml target set to QA
models: a random model file with query select current_database()
result: Database_QA

### Actual behavior

In Dev:
profiles.yml: I have profiles.yml target set to dev
models: a random model file with query select current_database()
result: Database_Dev
In QA:
profiles.yml: I have profiles.yml target set to QA
models: a random model file with query select current_database()
result: Database_**Dev**

### Steps To Reproduce

In Dev:
profiles.yml: I have profiles.yml target set to dev
models: a random model file with query select current_database()
result: Database_Dev
In QA:
profiles.yml: I have profiles.yml target set to dev
models: a random model file with query select current_database()
result: Database_**Dev**
(But if you switch the virtual environment or restart vscode, it picks up right database)

### Operating System

windows 10

### dbt version

1.6.0

### dbt Adapter

snowflake

### dbt Power User version

v0.18.0

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

mdesmet:
Could you also describe how you change the target?

mdesmet:
From slack: by changing the target in `profiles.yml`

VijaybabuNakkonda:
yes, thats right, by changing target value from profiles.yml

danieldiamond:
hey @mdesmet, experiencing the same bug so I'll share some more meaningful context:

Note: I am using `devcontainers`

I am authenticating with `externalbrowser`. I can confirm with `dbt debug` and `dbt build` that I can run dbt commands. yet when try out the query results tab it results in the error `251006: Password is empty`
```
{
  "code": -1,
  "message": "Database Error\n  251006: Password is empty",
[...]

PR #708, opened 2023-11-09
feat: allow users to change profiles.yml without having to reload window

lionsardesai:
fixes: #601 

Scenarios tested: 
- switch from a valid target to another valid target - in both cases project should reload the new credentials/adapter type as specified in the target.
- switch from an invalid target (project still loaded) to a valid one 
- switch from an invalid profile (project loaded without configs) to a valid profile and a valid target

Scenarios not tested (and probably not supported)
- switching to an adapter type that is not installed in the environment
- dbt was not installed before and now it is
- changing profiles root directory (either by using extension settings or using dbt defaults)

## Checklist

### DIFF
--- a/dbt_integration.py
+++ b/dbt_integration.py
@@ -61,7 +61,6 @@ if TYPE_CHECKING:
     from dbt.adapters.base import BaseRelation  # type: ignore
     from dbt.contracts.connection import AdapterResponse
 
-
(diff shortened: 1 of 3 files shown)


mdesmet (approved):

VijaybabuNakkonda:
This is great @mdesmet . Thanks for you action.!!

PR #711, opened 2023-11-13
fix: AI-1119 updated desc of key/instance

lionsardesai:
Display Name not working - vscode auto generates the name of the property based on the property text. cant seem to override using display name as in other places. if this request does get accepted, we might be able to do something.

This PR updates the description only.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,13 +56,15 @@
       {
         "title": "Altimate AI",
         "properties": {
-          "dbt.altimateInstanceName": {
-            "type": "string",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -56,13 +56,15 @@
       {
         "title": "Altimate AI",
         "properties": {
-          "dbt.altimateInstanceName": {
+          "dbt.altimateAiKey": {
```
mdesmet, line 59: Changing this key will break existing users. Are we sure we wanna do that?

mdesmet (approved):

PR #713, opened 2023-11-14
feat: lineage handle other node types

AdiGajbhiye:
## Checklist


src/webview_provider/newLineagePanel.ts
```diff
@@ -601,6 +608,11 @@ export class NewLineagePanel implements LineagePanelView {
         upstreamCount,
         downstreamCount,
         nodeType,
+        materialization: node?.config?.materialized,
```
mdesmet, line 611: I don't think materialized is always available if config is provided

mdesmet:
image
ephemeral seems to be missing icon.

mdesmet (commented):
LGTM except missing icon for ephemeral

mdesmet:
What about? I think we should restrict this to only models and snapshots as only those can be compiled.

```
if (node.config.materialized !== "seed") {
            compiledSql = await project.compileNode(node.name);
            if (!compiledSql) {
              return;
            }
          }
```

lionsardesai:
Few issues happening due to the last few changes.
before:
image
now:
image
Lineage not going all the way to the seed node.

src/webview_provider/newLineagePanel.ts
```diff
@@ -365,18 +367,19 @@ export class NewLineagePanel implements LineagePanelView {
     try {
       await Promise.all([
         ...Object.values(visibleTables).map(async (node) => {
-          let compiledSql: string | undefined;
-          if (node.config.materialized === "ephemeral") {
+          const nodeType = node.uniqueId.split(".")?.[0];
+          if (
+            !(nodeType === "model" || nodeType === "snapshot") ||
+            (nodeType === "model" && node.config.materialized === "ephemeral")
```
lionsardesai, line 373: added this because if we try to compile ephemeral models, we are getting the error that one of our users was also facing. for dbt-core 1.3, the error is: `Asked to compile class 'dbt.contracts.graph.compiled.CompiledModelNode' node, but it has no compiled form`

lionsardesai: also since the rest of the code deals with schema+compiled sql, by adding this condition we are skipping schema get for seed and source type of nodes

src/webview_provider/newLineagePanel.ts
```diff
@@ -365,18 +367,19 @@ export class NewLineagePanel implements LineagePanelView {
...
             return;
           }
-          if (node.config.materialized !== "seed") {
-            compiledSql = await project.compileNode(node.name);
-            if (!compiledSql) {
-              return;
-            }
+          const compiledSql = await project.compileNode(node.name);
+          if (!compiledSql) {
+            return;
```
lionsardesai, line 382: skip this check for seed node. compiledSql will be empty for seed but it will skip the schema get below

lionsardesai:
Another issue happening in this branch is if i have a model that has a source as a parent, clicking on the plus sign kills the webview (blank white page appears).

lionsardesai:
Other items from anand's list:
- Handle case of no lineage better - currently we dont give a message to the user if no lineage was found. We should also not grey out things if there is nothing to highlight.
- Sources columns are missing - view details is empty / sync with DB is missing/not working
- Disable view details for ephemeral models since they wont have columns in DB.
- test on dbt 1.3 (I have pushed a project [here](https://github.com/AltimateAI/sample_dbt_projects/tree/main/jaffle_shop_altimate) that should have the set up needed to run things in dbt 1.3. You will need python <= 3.10 for this to work. create a new virtual env using the pyproject.toml file in this project. All the errors shown here are also based off of that project.

mdesmet (approved):

PR #717, opened 2023-11-17
fix: state saving not working if someone ignored the walkthrough prompt

### LINKED ISSUE #407: Add ability to disable notifications
### Describe the feature

I want to disable irrelevant notifications, that I see after every VSCode startup.
image

Why these errors irrelevant for me?
* `An error occured while initializing the dbt project: Runtime Error Credentials in profile "dbt_[clientname]", target "dev" invalid: ['project_id'] is not valid under any of the given schemas`
I'm using one project directory as a template for creating new projects
* `An error occured while initializing the dbt project: Runtime Error dbt cannot run because no profile was specified for this dbt project. To specify a profile for this project, add a line like the this to your dbt_project.yml file: profile: [profile name] Here, [profile name] should be replaced with a profile name defined in your profiles.yml file. You can find profiles.yml here: /home/ivantoriya-ubuntu20_04_4/.dbt/profiles.yml`
same as above + I'm storing profiles in a different way
* `dbt Power User detected 17 projects in your work space, this will negatively affect performance.`
My performance not affected 

* It can't be done in VSCode settings.
* I don't want to change my setup, because of that.

### Who will benefit?

Me, my team, and someone else with a similar issue.

lionsardesai:
Fix #407 

## Checklist

### DIFF
--- a/src/manifest/dbtProjectContainer.ts
+++ b/src/manifest/dbtProjectContainer.ts
@@ -184,17 +184,17 @@ export class DBTProjectContainer implements Disposable {
   }
 
   setToWorkspaceState(key: string, value: any) {
-    this.context?.workspaceState.update(key, value);
+    this.context!.workspaceState.update(key, value);
   }
   getFromWorkspaceState(key: string): any {
-    return this.context?.workspaceState.get(key);
+    return this.context!.workspaceState.get(key);
   }
   setToGlobalState(key: string, value: any) {
-    this.context?.globalState.update(key, value);
+    this.context!.globalState.update(key, value);
   }
 
   getFromGlobalState(key: string): any {
-    return this.context?.globalState.get(key);
+    return this.context!.globalState.get(key);
   }
 
   get extensionId(): string {
--- a/src/manifest/dbtProjectContainer.ts
+++ b/src/manifest/dbtProjectContainer.ts
@@ -134,9 +134,8 @@ export class DBTProjectContainer implements Disposable {
     );
     if (answer === PromptAnswer.YES) {
       commands.executeCommand("dbtPowerUser.openSetupWalkthrough");
-    } else if (answer === PromptAnswer.IGNORE) {
-      this.setToGlobalState("showSetupWalkthrough", false);
-    }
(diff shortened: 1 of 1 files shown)


src/manifest/dbtProjectContainer.ts
```diff
@@ -192,7 +194,7 @@ export class DBTProjectContainer implements Disposable {
   }
 
   getFromGlobalState(key: string): any {
-    this.context?.globalState.get(key);
+    return this.context?.globalState.get(key);
```
mdesmet, line 197: Shouldn't the context always be available here?

src/manifest/dbtProjectContainer.ts
```diff
@@ -134,12 +134,14 @@ export class DBTProjectContainer implements Disposable {
     );
     if (answer === PromptAnswer.YES) {
       commands.executeCommand("dbtPowerUser.openSetupWalkthrough");
+    } else if (answer === PromptAnswer.IGNORE) {
+      this.setToGlobalState("showSetupWalkthrough", false);
     }
```
mdesmet, line 139:
```suggestion
    if (answer === PromptAnswer.YES) {
      commands.executeCommand("dbtPowerUser.openSetupWalkthrough");
    } 
    this.setToGlobalState("showSetupWalkthrough", false);
```

mdesmet (approved):
LGTM

src/manifest/dbtProjectContainer.ts
mdesmet: Also `getFromGlobalState` should accept type similar like globaState.get

mdesmet: make also use of this method in walkthrough.

src/manifest/dbtProjectContainer.ts
lionsardesai: doing this will set the flag even if user clicks on the x mark to close the prompt instead of clicking on the button ignore. X mark should mean ignore the prompt and not take action so doing this.

PR #721, opened 2023-11-20
fix: handle source node

AdiGajbhiye:
## Checklist

### DIFF
--- a/src/manifest/parsers/graphParser.ts
+++ b/src/manifest/parsers/graphParser.ts
@@ -85,6 +85,7 @@ export class GraphParser {
       const nodeName = restNodeName.join(".");
       switch (nodeType) {
         case "source": {
+          // TODO: match source label to graph children nodes key
           const [sourceName, tableName] = nodeName.split(".");
           const url = sourceMetaMap
             .get(sourceName)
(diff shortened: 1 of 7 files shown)


src/manifest/parsers/graphParser.ts
```diff
@@ -89,11 +89,7 @@ export class GraphParser {
           const url = sourceMetaMap
             .get(sourceName)
             ?.tables.find((table) => table.name === tableName)?.path!;
-          return new Source(
-            `${tableName} (${sourceName})`,
-            parentNodeName,
-            url,
-          );
+          return new Source(sourceName, parentNodeName, url);
```
mdesmet, line 92: Why this change?

src/webview_provider/newLineagePanel.ts
```diff
@@ -252,19 +300,47 @@ export class NewLineagePanel implements LineagePanelView {
...
     }
+    const project = this.getProject();
+    if (!project) {
+      return;
+    }
     if (nodeType === DBTProject.RESOURCE_TYPE_SOURCE) {
       const { sourceMetaMap } = event;
       const splits = tableKey.split(".");
-      const schema = splits[2];
+      const sourceName = splits[2];
```
suryaiyer95, line 310: Is there a case where length might be > 4?

mdesmet:
Lineage is not working with sources

image

mdesmet (approved):

PR #733, opened 2023-11-27
Fix: source cll

AdiGajbhiye:
## Checklist

### DIFF
--- a/src/altimate.ts
+++ b/src/altimate.ts
@@ -1,7 +1,7 @@
 import { window, workspace } from "vscode";
 import { provideSingleton } from "./utils";
 import fetch from "node-fetch";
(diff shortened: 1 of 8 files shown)


src/altimate.ts
```diff
@@ -1,34 +1,44 @@
...
 
 interface Schemas {
   [key: string]: { [key: string]: unknown };
 }
 
+export type ModelNode = Pick<
+  NodeMetaData,
+  "database" | "schema" | "name" | "alias" | "uniqueId" | "columns"
+>;
+
```
mdesmet, line 25: Please make this an explicit interface. No need to use NodeMetadata here.

mdesmet (commented):
LGTM except small remark on Altimate interface.

PR #740, opened 2023-12-01
fix: cll edge cases

AdiGajbhiye:
## Overview

- fix terminating conditions for cll to avoid unnecessary API request
- not making API requests when no targets
- for upstream expansion, instead of getting parent models, getting ancestor models(not ephemeral)
- fix error message for ephemeral case
- fix unnecessary models were being sent when see more nodes present
- fix table list item in see more side panel
- auto expanding table lineage for ephemeral models
- fix identifier for source table

## Checklist

### DIFF
--- a/src/manifest/dbtProject.ts
+++ b/src/manifest/dbtProject.ts
@@ -62,7 +62,13 @@ interface FileNameTemplateMap {
   [key: string]: string;
 }
 
-interface ResolveReferenceResult {
+interface ResolveReferenceNodeResult {
+  database: string;
+  schema: string;
+  alias: string;
+}
+
+interface ResolveReferenceSourceResult {
   database: string;
   schema: string;
   alias: string;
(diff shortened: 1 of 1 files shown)


src/manifest/dbtProject.ts
```diff
@@ -625,11 +627,11 @@ export class DBTProject implements Disposable {
...
-    return this.getColumsOfRelation(
-      node.database,
-      node.schema,
-      node.alias || tableName,
-    );
+    let _tableName = node.alias || tableName;
+    if (node.resource_type === DBTProject.RESOURCE_TYPE_SOURCE) {
+      _tableName = node.identifier;
+    }
+    return this.getColumsOfRelation(node.database, node.schema, _tableName);
```
mdesmet, line 634: This should only be called for sources. Please adapt it in that way.

mdesmet: Ideally the `ResolveReferenceResult` should be split per Source vs node as they don't have same properties.

mdesmet (commented):
LGTM except the logic in `getColumnsOfSource`

src/manifest/dbtProject.ts
AdiGajbhiye: For now, I have directly used `node.identifier`. Will change `ResolveReferenceResult` in future PR

mdesmet (approved):

PR #749, opened 2023-12-05
feat: Add packagesInstallPath fetch method

gaurpulkit:
## Checklist

### DIFF
--- a/src/manifest/dbtProject.ts
+++ b/src/manifest/dbtProject.ts
@@ -334,6 +334,9 @@ export class DBTProject implements Disposable {
       await this.python?.lock(
         (python) => python!`to_dict(project.safe_parse_project(${init}))`,
       );
+      // Add dbt packagesInstallPath in dbt modules
+      const packagesInstallPath = await this.findPackagesInstallPath();
+      DBTProject.DBT_MODULES.push(packagesInstallPath);
       if (init) {
         this.adapterType = (await this.python?.lock(
           (python) => python!`project.config.credentials.type`,
@@ -1009,6 +1012,14 @@ select * from renamed
         : false) ||
       "target"
     );
+  }
+
+  async findPackagesInstallPath() {
+    const packages_install_path = await this.python?.lock(
+      (python) => python!`to_dict(project.config.packages_install_path)`,
+    );
+    console.log(`Packages install path: ${packages_install_path}`);
+    return packages_install_path;
   }
 
   private async refresh() {
--- a/src/manifest/dbtProject.ts
+++ b/src/manifest/dbtProject.ts
@@ -93,7 +93,7 @@ export class DBTProject implements Disposable {
   private macroPaths: string[];
   private python?: PythonBridge;
   private pythonBridgeInitialized = false;
-  public initializationException: any | null = null;
+  private initializationException?: Error;
 
   private _onProjectConfigChanged =
     new EventEmitter<ProjectConfigChangedEvent>();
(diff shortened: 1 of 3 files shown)


src/manifest/dbtProject.ts
```diff
@@ -334,6 +334,9 @@ export class DBTProject implements Disposable {
       await this.python?.lock(
         (python) => python!`to_dict(project.safe_parse_project(${init}))`,
       );
+      // Add dbt packagesInstallPath in dbt modules
+      const packagesInstallPath = await this.findPackagesInstallPath();
+      DBTProject.DBT_MODULES.push(packagesInstallPath);
```
mdesmet, line 339: If we can reliably retrieve the packagesInstallPath no need for a constant. Please remove the usage of the constant and use `await this.findPackagesInstallPath();`
Also as this involves a lock it might be better to cache all configs in a local variable in the The dbtProject instead. We probably need more configs.

src/manifest/dbtWorkspaceFolder.ts
```diff
@@ -127,6 +162,11 @@ export class DBTWorkspaceFolder implements Disposable {
         projectConfig,
         this._onManifestChanged,
       );
+      try {
+        await dbtProject.initializePythonBridge();
+      } catch (exc: any) {
+        dbtProject.initializationException = exc;
```
mdesmet, line 168: See above comment, this shouldn't use a property of `dbtProject`

src/manifest/dbtWorkspaceFolder.ts
```diff
@@ -83,7 +79,34 @@ export class DBTWorkspaceFolder implements Disposable {
...
+    );
+    await Promise.all(
+      this.dbtProjects.map(async (project) => {
+        await project.initializeDBTProject();
+      }),
+    );
+
+    // Exception handling after filtering
+    this.dbtProjects.map((project) => {
+      project.handlePythonBridgeException();
```
mdesmet, line 108: This code should also be moved here. No need to have a public method and a variable for the exception

src/manifest/dbtWorkspaceFolder.ts
```diff
@@ -83,7 +79,34 @@ export class DBTWorkspaceFolder implements Disposable {
...
+      }),
+    );
+    await Promise.all(
+      this.dbtProjects.map(async (project) => {
+        await project.initializeDBTProject();
+      }),
+    );
+
+    // Exception handling after filtering
+    this.dbtProjects.map((project) => {
```
mdesmet, line 107: use `forEach`

src/manifest/dbtWorkspaceFolder.ts
```diff
@@ -152,20 +192,23 @@ export class DBTWorkspaceFolder implements Disposable {
...
-  private unregisterDBTProject(uri: Uri) {
+  private async unregisterDBTProject(uri: Uri) {
     const projectToDelete = this.dbtProjects.find(
       (dbtProject) => dbtProject.projectRoot.fsPath === uri.fsPath,
     );
     if (projectToDelete === undefined) {
       return;
     }
+    // Close python bridge
+    await projectToDelete.closePythonBridge();
```
mdesmet, line 203: Move this call to dispose of dbtProject and wrap this in try catc.

src/manifest/dbtProject.ts
```diff
@@ -1011,6 +1040,14 @@ select * from renamed
     );
   }
 
+  async findPackagesInstallPath() {
+    const packages_install_path = await this.python?.lock(
+      (python) => python!`to_dict(project.config.packages_install_path)`,
+    );
+    console.log(`Packages install path: ${packages_install_path}`);
```
mdesmet, line 1047: Remove this line as this log line is not super useful, we would not know the packages install path belongs to which project for example.

Issue #756, opened 2023-12-07
query limit should always be a positive value


anandgupta42:
### Expected behavior

We need to make sure that query limit should always be a positive value. Add a check to prevent users from putting negative value.

### Actual behavior

The user can put negative value

### Steps To Reproduce

1. set the value as -500 before executing the query

![Screenshot 2023-12-06 at 9 36 13 PM](https://github.com/AltimateAI/vscode-dbt-power-user/assets/93243293/0e5b6816-d863-4099-a1bd-2c6b8f4db38f)

### Operating System

all versions

### dbt version

all versions

### dbt Adapter

all adapters

### dbt Power User version

0.27.2

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #758, opened 2023-12-08
fix: set 1 as min query limit value

saravmajestic:
## Overview
This PR makes the query limit input to reset to 1 if user has entered negative value.

resolves: #756 

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "bugs": {
     "url": "https://github.com/AltimateAI/vscode-dbt-power-user/issues"
   },
-  "version": "0.28.0",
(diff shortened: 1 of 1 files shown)


mdesmet:
This is not the only way to set the limit. We should throw an error if the limit is set to an invalid value through vscode settings as well.

mdesmet:
@saravmajestic:

I meant using onDidChangeConfiguration: [Event](https://code.visualstudio.com/api/references/vscode-api#Event)<[ConfigurationChangeEvent](https://code.visualstudio.com/api/references/vscode-api#ConfigurationChangeEvent)>

Issue #679, opened 2023-10-28
add support for exposures in table lineage view


anandgupta42:
### Describe the feature

add support for exposures in table lineage view

### Who will benefit?

All the users who have exposures defined in their dbt project and also use the lineage panel

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #759, opened 2023-12-08
feat: show exposures node in lineage panel

saravmajestic:
## Overview
As part of this PR, `exposures` from dbt will be displayed in new lineage panel view. Also `View details` will show basic details of the exposure in sidebar

Resolves: #679 

Screenshot:
image

## Checklist


new_lineage_panel/src/TableDetails.tsx
```diff
@@ -73,23 +74,6 @@ const ColumnCard: FunctionComponent<{
   );
 };
 
-const PurposeSection: FunctionComponent<{
```
saravmajestic, line 76: Extracted this to separate component

tsconfig.json
```diff
@@ -10,12 +10,15 @@
     "types": ["reflect-metadata", "node"],
     "lib": ["es6", "es2020", "dom"],
     "sourceMap": true,
-    "rootDir": "src",
-    "strict": true /* enable all strict type-checking options */
+    "rootDir": ".",
```
saravmajestic, line 13: TODO: verify if this is not breaking any build

AdiGajbhiye:
UI looks good. Maybe one thing that we can do is hide cards that don't have content, like tags in the above screenshot.
@ppradnesh @anandgupta42  what do you think?

.vscode/settings.json
```diff
@@ -1,11 +1,11 @@
...
\ No newline at end of file
+  "files.exclude": {
+    "out": false // set this to true to hide the "out" folder with the compiled JS files
+  },
+  "search.exclude": {
+    "out": true // set this to false to include "out" folder in search results
+  },
+  // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
+  "typescript.tsc.autoDetect": "off"
+}
```
mdesmet, line 11: Revert change

types/dbt/exposure.ts
```diff
@@ -0,0 +1,14 @@
...
+  label?: string;
+  maturity?: "high" | "medium" | "low";
+  name: string;
+  owner: { email: string; name: string };
+  tags: [string];
+  unique_id: string;
+  url?: string;
+  type: "dashboard" | "notebook" | "analysis" | "ml" | "application";
+  config: { enabled: boolean };
+};
```
mdesmet, line 14: Move to domain.ts

mdesmet (commented):
LGTM except some minor things.

types/dbt/exposure.ts
saravmajestic: `domain.ts` has types specific to altimate extension types and also could not be shared with other packages in the repo. Reason for creating separate types directory is to have abstraction and ability to share between packages in repo. Also this particular file is to define types of dbt core apis, which is needed in all packages. Please let me know if you think it is still to be moved to domain.ts. @mdesmet

.vscode/settings.json
saravmajestic: done

types/dbt/exposure.ts
mdesmet: I get what you are trying to do but it is premature. The other classes are actually also dbt core related data classes. Please move it to domain and also name the class similar to the other classes, ie. without Dbt prefix.

Let's first think this refactoring through instead of just applying it to one class.

types/domain.ts
```diff
@@ -0,0 +1,9 @@
+import { DbtExposure } from "./dbt/exposure";
+
+export interface ExposureMetaData extends DbtExposure {
+  path: string;
+  unique_id: string;
```
mdesmet, line 5: This is also present in DbtExposure?

What's the logic behind splitting this, as I think some of the data here is coming from dbt as well?

saravmajestic: These are extra attributes added on top of dbt type from manifest

src/domain.ts
```diff
@@ -72,6 +71,24 @@ interface TestMetaData {
...
+export interface ExposureMetaData {
+  description?: string;
+  depends_on: { macros: [string]; nodes: [string]; sources: [string] };
+  label?: string;
+  maturity?: "high" | "medium" | "low";
+  name: string;
+  owner: { email: string; name: string };
+  tags: [string];
+  url?: string;
+  type: "dashboard" | "notebook" | "analysis" | "ml" | "application";
```
mdesmet, line 83: Can we type that to string. We don't wanna fail because dbt adds a new type.

src/domain.ts
```diff
@@ -72,6 +71,24 @@ interface TestMetaData {
   column_name?: string;
 }
 
+export interface ExposureMetaData {
+  description?: string;
+  depends_on: { macros: [string]; nodes: [string]; sources: [string] };
+  label?: string;
+  maturity?: "high" | "medium" | "low";
```
mdesmet, line 78: Can we type that to string. We don't wanna fail because dbt adds a new type.

mdesmet (commented):
LGTM code wise. Can you add exposures to our dbt project than I will test it.

mdesmet (approved):

Issue #485, opened 2023-07-25
Add integration tests for dbt integration


mdesmet:
### Describe the feature

Test dbt integration with multiple dbt versions to ensure stability

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

saravmajestic:
@anandgupta42 @mdesmet @AdiGajbhiye based on my understanding, we have to test 2 aspects in the extension (please add if i am missing anything)
- loading the dbt project and verifying if all data is parsed as expected and ready to be sent to webview
- web view interactions, for example: in lineage panel, whether graph is displayed, clicking the nodes does the required action etc

We can use the proposed testing flow as [here](https://code.visualstudio.com/api/working-with-extensions/testing-extension) for the first aspect. 

Vscode as such does not support webview testing: https://github.com/microsoft/vscode-test/issues/46#issuecomment-548589003. So as you mentioned, we can try exploring other libraries like https://github.com/redhat-developer/vscode-extension-tester/wiki/WebView, but it might be complicated to switch between frames and handling the interactions and this library is not matured enough.

I propose that we can follow regular web automation testing for webviews. Since we are using react for webviews, we can use jest and react-testing-library to test all the interactions by providing mock data for apis. 

Let me know what you guys think

PR #760, opened 2023-12-08
feat: testing framework

saravmajestic:
## Overview
- This PR implements a basic automation test using vscode-test. More details [here](https://code.visualstudio.com/api/working-with-extensions/testing-extension#the-test-runner-script) 
- This PR enables testing only vscode api and container related functionality. For testing webview panels, we need to handle this in the react apps because vscode-test does not support webview testing: https://github.com/microsoft/vscode-test/issues/46#issuecomment-548589003

### Major chanegs
- made changes to get inversify container from extension instead of creating new instance
- `node-fetch` has to be dynamically imported since it is esm only module now and does not support commonjs

Ticket: https://github.com/AltimateAI/vscode-dbt-power-user/issues/485

## Checklist


.vscode/launch.json
```diff
@@ -3,34 +3,32 @@
...
-			],
-			"outFiles": [
-				"${workspaceFolder}/out/test/**/*.js"
-			],
-			"preLaunchTask": "npm: test-compile"
-		}
-	]
-}
\ No newline at end of file
+  "version": "0.2.0",
```
saravmajestic, line 6: prettier formatting changes

package.json
```diff
@@ -807,16 +807,17 @@
...
-    "test-compile": "tsc -p ./",
     "lint": "eslint src --ext ts",
     "lint:fix": "eslint src --ext ts --fix",
     "deploy-vscode": "vsce publish",
     "build-vsix": "vsce package",
     "deploy-openvsx": "ovsx publish",
     "panel:lineage": "npm run build --prefix ./new_lineage_panel",
-    "install:panels": "npm install --prefix ./new_lineage_panel"
+    "install:panels": "npm install --prefix ./new_lineage_panel",
+    "pretest": "rm -Rf ./.vscode-test/extensions;rm -Rf dist; rm -Rf ./new_lineage_panel/dist; npm run panel:lineage; tsc -p ./",
```
AdiGajbhiye, line 819: maybe we could have additional clean commands like, `clean:extension`, `clean:panel`, etc. and use those.

PR #763, opened 2023-12-10
fix: support dbt path resolver in multi root workspace.

yassun7010:
## Overview
In our project, we use [multi-root-workspace](https://code.visualstudio.com/docs/editor/multi-root-workspaces).

And we have installed `dbt` on 'dbt' workspace folder, not installed on workspace root.

In the current implementation, `dbt-power-user` searchs `dbt` of workspace root, so editor raises error in `dbt_project.yml`.

the error message of my editor is here:

image

Ideally, I would like it to use the python paths in the current edit workspace folder, but this feature is too large, so I have changed the search logic to python environments when `dbt` is not installed in workspace root.

## Checklist

### DIFF
--- a/src/manifest/pythonEnvironment.ts
+++ b/src/manifest/pythonEnvironment.ts
@@ -2,6 +2,9 @@ import { Disposable, Event, extensions, Uri, workspace } from "vscode";
 import { EnvironmentVariables } from "../domain";
 import { provideSingleton, substituteSettingsVariables } from "../utils";
 import { TelemetryService } from "../telemetry";
+import { PythonExtension } from "@vscode/python-extension";
+import * as path from "path";
+import * as fs from "fs";
 
 interface PythonExecutionDetails {
   getPythonPath: () => string;
@@ -78,9 +81,25 @@ export class PythonEnvironment implements Disposable {
     const api = extension.exports;
 
     return (this.executionDetails = {
-      getPythonPath: () =>
-        api.settings.getExecutionDetails(workspace.workspaceFile)
-          .execCommand[0],
+      getPythonPath: () => {
+        const pythonPath: string = api.settings.getExecutionDetails(
+          workspace.workspaceFile,
+        ).execCommand[0];
+
+        if (!findDBTPath(pythonPath)) {
+          for (const workspaceFolder of workspace.workspaceFolders || []) {
+            const candidatePythonPath = api.settings.getExecutionDetails(
+              workspaceFolder.uri,
+            ).execCommand[0];
+
+            if (findDBTPath(candidatePythonPath)) {
+              return candidatePythonPath;
+            }
+          }
+        }
+
+        return pythonPath;
+      },
       onDidChangeExecutionDetails: api.settings.onDidChangeExecutionDetails,
       getEnvVars: () => {
         const configText = workspace.getConfiguration();
@@ -123,4 +142,9 @@ export class PythonEnvironment implements Disposable {
       },
     });
   }
+}
+
+function findDBTPath(pythonPath: string): boolean {
+  const dbtPath = path.join(path.dirname(pythonPath), "dbt");
+  return fs.existsSync(dbtPath);
 }
--- a/src/manifest/pythonEnvironment.ts
+++ b/src/manifest/pythonEnvironment.ts
@@ -1,10 +1,10 @@
-import { Disposable, Event, extensions, Uri, workspace } from "vscode";
+import { Disposable, Event, extensions, Uri, window, workspace } from "vscode";
 import { EnvironmentVariables } from "../domain";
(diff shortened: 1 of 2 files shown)


mdesmet:
Can we create an issue first with clear reproduction steps?

The Python Path is set through selecting the Python interpreter, so not sure if this change is required.

yassun7010:
Sorry, I create https://github.com/AltimateAI/vscode-dbt-power-user/issues/770 .

src/manifest/pythonEnvironment.ts
```diff
@@ -124,3 +143,8 @@ export class PythonEnvironment implements Disposable {
     });
   }
 }
+
+function findDBTPath(pythonPath: string): boolean {
+  const dbtPath = path.join(path.dirname(pythonPath), "dbt");
+  return fs.existsSync(dbtPath);
```
mdesmet, line 149: Can you explain what this code is doing?
Why dbt would exist on this path?

yassun7010: Sure.

In [PEP405](https://peps.python.org/pep-0405/#creating-virtual-environments), 

```
Third-party packages installed into the virtual environment will have their Python modules placed in the site-packages directory, and their executables placed in bin/ or Scripts.
```

On the implementation of venv, third-party package (and `python` ) executables placed in `bin/` directory of the workspace python virtual environment.

Using this relationship, this code is searching for dbt executables.

If the dbt executable path is not found in all virtual environments, this will behave the same as before.

mdesmet: Is this always valid for any installations of python, is the bin folder always where executables are placed?

yassun7010: Unfortunately, this rule is not guaranteed for all virtual environment implementations.

In my environment (Mac M2, Mac OS 14.2), it is placed on bin with venv (this means that poetry, pdm also show the same behavior), virtualenv.


Again, if the dbt execution path is not found, the code returns the same as before.

mdesmet: Can we not use the dbt detection mechanism instead that we already have?

mdesmet: See

https://github.com/AltimateAI/vscode-dbt-power-user/blob/cf6aa0fb5c0f40064ce8d2e3c7c7daefe0e12e6e/src/dbt_client/dbtCommandFactory.ts#L46-L54

mdesmet: I think that will also require fixing as it just uses getFirstWorkspacePath()

yassun7010: getFirstWorkspacePath() does not solve this problem.

getFirstWorkspacePath() is executed when the editor is opened, but not when the workspace is switched by changing the editing file.

An alternative would be to use onDidChangeWorkspaceFolders to detect the event and switch the entire dbt command-related environment. This change is likely to be a big change.

mdesmet: OK I think the code piece doesn't matter as this is just current working directory configuration. What I'm saying is that you can use this command to check if dbt is installed or not instead of just checking if dbt is present.

yassun7010: Yes.

I've experimented further with `vscode-shandy-sqlfmt`, and found that
that `cwd: this.getFirstWorkspacePath()` isn't enough,
maybe doing `venv/bin/activate && which dbt` or `python -c "import dbt"`  to determine the installation of `dbt`.

mdesmet: The cwd command is not important, that is just the current working directory, it is not relevant for checking if dbt is available. You should use the pythonPath and execute the command I shared above.

yassun7010: I am working on this issue and encountering a troubling dependency.

`PythonEnviroment` is required to run `DBTCommand`, but since this fix requires `DBTCommand` to determine `PythonEnviroment`.

Should we add special logic to check dbt installation to determine PythonEnviroment?

yassun7010: I modified the logic of the detection using VerifyDbtInstalledCommand.

anandgupta42:
@yassun7010 We are testing this PR and hoping to close it by early next week.

mdesmet (approved):

suryaiyer95 (commented):
Looks pretty good!

anandgupta42:
Hey @yassun7010 - We need more information before we can ship this feature. Can you please reach out to us in the[ dbt community slack channel](https://getdbt.slack.com/archives/C05KPDGRMDW) or drop me a message at anand [at] altimate [dot] ai

yassun7010:
@anandgupta42 

I contacted you via slack dm.

Is there a channel etc.. for this conversation?

PR #776, opened 2023-12-13
Feat: session id and other fixes in lineage

AdiGajbhiye:
## Overview

- add session id in cll request
- adding extension version in every API request
- skip unnecessary cll API request
- showing dialog when no lineage present
- remove auto expanding 1st level when opening file when node already exists
- hide tests when not available
- fix open file ux
- fix credentials missing error message on freemium features
- when validateSQL action, show only 1 compiled SQL tab and close it when SQL is valid

## Checklist


new_lineage_panel/src/App.tsx
```diff
@@ -26,13 +26,18 @@ import {
...
+  highlightTableConnections,
+  layoutElementsOnCanvas,
+} from "./graph";
 import { TableDetails } from "./TableDetails";
 import { Button, Card, CardBody, Input, Label, Tooltip } from "reactstrap";
 import AlertCircleIcon from "./assets/icons/alert-circle.svg?react";
 import styles from "./styles.module.scss";
 import { TABLES_SIDEBAR, COLUMNS_SIDEBAR, EXPOSURE_SIDEBAR } from "./constants";
 import ExposureDetails from "./exposure/ExposureDetails";
+import { v4 as uuidv4 } from "uuid";
```
saravmajestic, line 40: `window.crypto.randomUUID()` instead of uuid library

new_lineage_panel/src/App.tsx
```diff
@@ -172,79 +183,54 @@ function App() {
       aiEnabled: boolean;
     }) => {
       setShowSidebar(false);
-      if (!args) {
-        return;
-      }
-      const { node, aiEnabled: _aiEnabled } = args;
-      aiEnabled = _aiEnabled;
+      if (!args) return;
+      aiEnabled = args.aiEnabled;
```
saravmajestic, line 187: should `aiEnabled` static variable? can this be set as ref or state variable? this type of variable declarations go against react patterns

new_lineage_panel/src/App.tsx
```diff
@@ -409,7 +395,10 @@ function App() {
           sidebarScreen,
           setSidebarScreen,
           selectedColumn,
-          setSelectedColumn,
+          setSelectedColumn: (v) => {
+            setSelectedColumn(v);
+            sessionId = uuidv4();
```
saravmajestic, line 400: this one too. better to be a state/ref/context variable instead of static reference

new_lineage_panel/src/App.tsx
```diff
@@ -73,9 +78,15 @@ export const endProgressBar = () => {
 export const previewFeature = () => {
   vscode.postMessage({ command: "previewFeature", args: {} });
 };
+export const showNoLineage = (params: { table: string; name: string }) => {
+  vscode.postMessage({ command: "showNoLineage", args: { params } });
+};
 
 export let isDarkMode = false;
 
+let sessionId = uuidv4();
```
saravmajestic, line 87: Would be better to keep in separate file if it needs to be static. Otherwise, whenever this variable or `getSessionId` is referred, App.tsx will create circular dependency

new_lineage_panel/src/TableDetails.tsx
```diff
@@ -297,12 +299,26 @@ const TableDetails = () => {
...
+    try {
+      const result = await Promise.all([
+        _bfsTraversal(true),
+        _bfsTraversal(false),
+      ]);
+      if (result.every((isLineage) => !isLineage)) {
+        showNoLineage(_column);
+      }
+    } catch (e) {
+      console.error(e);
```
saravmajestic, line 314: nit: extra info on this log could be useful

new_lineage_panel/src/TableDetails.tsx
```diff
@@ -297,12 +299,26 @@ const TableDetails = () => {
...
+      setSelectedColumn({ table: "", name: "" });
+    }
     endProgressBar();
   };
   if (isLoading || !data || !selectedTable) return <ComponentLoader />;
+  const tabs = ["Column"];
```
saravmajestic, line 320: `useMemo` could be used here for performance improvement. Otherwise, this tabs will be recalculated everytime state changes in this component

saravmajestic (changes requested):

new_lineage_panel/src/TableDetails.tsx
AdiGajbhiye: Should be fine as there is no heavy calculation being performed here. And this portion will probably change in the future.

new_lineage_panel/src/App.tsx
AdiGajbhiye: this should be fine as it is mostly constant through out the app

mdesmet (approved):

anandgupta42 (approved):
Couple of test added and the rest will be manually tested.

PR #777, opened 2023-12-13
adding network logs

AdiGajbhiye:
## Checklist

### DIFF
--- a/src/telemetry/index.ts
+++ b/src/telemetry/index.ts
@@ -17,12 +17,9 @@ export class TelemetryService implements vscode.Disposable {
     properties?: { [key: string]: string },
     measurements?: { [key: string]: number },
   ) {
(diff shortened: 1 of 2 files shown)


mdesmet (approved):

src/telemetry/index.ts
```diff
@@ -8,14 +8,18 @@ export class TelemetryService implements vscode.Disposable {
     "50598369-dd83-4f9a-9a65-ca1fa6f1785c",
   );
 
+  private instanceName = vscode.workspace
+    .getConfiguration("dbt")
+    .get<string>("altimateInstanceName");
+
```
mdesmet, line 14: maybe just inline this that we always latest setting, otherwise this would need a restart to be updated.

mdesmet (approved):

Issue #745, opened 2023-12-04
Query results could use dbt show to support databases with different syntax without having to override the query template


Anteus:
### Describe the feature

For example Teradata only supports select top {{limit}} * instead of limit at the end. Could probably use dbt show --select "<model_name>" instead to use the current adapter's supported query.

### Who will benefit?

Users of other databases that doesn't support the limit syntax

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #793, opened 2023-12-22
fix: use dbt macro get_limit_subquery_sql for limit query if available

saravmajestic:
## Overview
- This PR enables adding limit to queries based on dbt macro `get_limit_subquery_sql` (ex: https://github.com/oracle/dbt-oracle/blob/5c6816b5058f6d6ed76e3362eafccf9501f92c37/dbt/include/oracle/macros/show.sql#L17-L23)
- also removed queryTemplate usage from frontend (query panel) and handled the logic in dbtcoreintegration.

Conditions:
- if dbt version is>= 1.5 then get the query template using dbt `get_limit_subquery_sql` macro
- if above step didnt return any macro, check if user has provided query template, we will use the same
- if any failure happened, use default query template

resolves: #745 

## Checklist


src/manifest/dbtProject.ts
```diff
@@ -958,7 +959,54 @@ select * from renamed
     }
     this.queryResultPanel.getSummary(compiledSql, this.adapterType);
   }
+  private getLimitQuery(queryTemplate: string, query: string, limit: number) {
+    return queryTemplate
+      .replace("{query}", () => query)
+      .replace("{limit}", () => limit.toString());
+  }
+  private async getQuery(
```
saravmajestic, line 967: I would prefer these helper functions to be extracted to separate file. @mdesmet your thoughts?

src/manifest/dbtProject.ts
```diff
@@ -44,6 +44,7 @@ import { PythonEnvironment } from "./pythonEnvironment";
 import { TelemetryService } from "../telemetry";
 import { ValidateSqlParseErrorResponse } from "../altimate";
 import * as crypto from "crypto";
+import { DefaultQueryTemplate } from "../webview_provider/constants";
```
mdesmet, line 47: If it is part of multiple code parts, maybe move it to domain.ts.

src/manifest/dbtProject.ts
```diff
@@ -958,7 +959,54 @@ select * from renamed
     }
     this.queryResultPanel.getSummary(compiledSql, this.adapterType);
   }
+  private getLimitQuery(queryTemplate: string, query: string, limit: number) {
```
mdesmet, line 962: Please add spacing between functions.

src/manifest/dbtProject.ts
mdesmet: Just moving these helper functions into a separate file benefit us how? If we want to refactor this, it should be rather something about making query execution easier. Maybe moving all code related to query execution into a single class with a clear API and make it a collaborator of the DbtProject could work but let's keep this refactoring out of this PR.

mdesmet, line 967: Add spacing between functions.

src/manifest/dbtProject.ts
```diff
@@ -958,7 +959,54 @@ select * from renamed
...
+        "executeMacroGetLimitSubquerySQLError",
+        err,
+        { adapter: this.adapterType },
+      );
+    }
+    return {
+      queryTemplate: DefaultQueryTemplate,
+      limitQuery: this.getLimitQuery(DefaultQueryTemplate, query, limit),
+    };
+  }
```
mdesmet, line 1009: Add spacing between functions

src/manifest/dbtProject.ts
```diff
@@ -958,7 +959,54 @@ select * from renamed
...
+  private async getQuery(
+    query: string,
+    limit: number,
+  ): Promise<{ queryTemplate: string; limitQuery: string }> {
+    const queryTemplate = workspace
+      .getConfiguration("dbt")
+      .get<string>("queryTemplate");
+
+    if (queryTemplate && queryTemplate !== DefaultQueryTemplate) {
```
mdesmet, line 975: Why do we explicitly check if `!==  DefaultQueryTemplate`, If users set it why don't we just use it? Isn't it null in case of unset, otherwise why do we need to put it in the other place as 

```
const queryTemplate = workspace
      .getConfiguration("dbt")
      .get<string>("queryTemplate", DefaultQueryTemplate);
```

src/manifest/dbtProject.ts
```diff
@@ -958,7 +959,54 @@ select * from renamed
...
+      console.log("Using user provided query template", queryTemplate);
+      const limitQuery = this.getLimitQuery(queryTemplate, query, limit);
+
+      return { queryTemplate, limitQuery };
+    }
+
+    try {
+      const dbtVersion = await this.getDBTVersion();
```
mdesmet, line 983: I think this can be cached. We shouldn't call getDbtVersion() on every query execution. We just need to reevaluate getDbtVersion() every time the python environment is changed.

src/manifest/dbtProject.ts
saravmajestic: @mdesmet i also assumed that we will get null, if this value is not set. but i am getting default value that is set in package.json. Because of that i had to add this check

src/manifest/dbtProject.ts
saravmajestic: should that be part of this PR?

src/manifest/dbtProject.ts
saravmajestic: my bad. I am actually getting empty value now. fixed the code.

src/manifest/dbtProject.ts
```diff
@@ -578,6 +579,56 @@ select * from renamed
...
+        "executeMacroGetLimitSubquerySQLError",
+        err,
+        { adapter: this.adapterType },
+      );
+    }
+    return {
+      queryTemplate: DefaultQueryTemplate,
+      limitQuery: this.getLimitQuery(DefaultQueryTemplate, query, limit),
+    };
+  }
```
mdesmet, line 630: This shold not be in dbtProject but in dbtCoreIntegration.

mdesmet (changes requested):

src/domain.ts
```diff
@@ -176,3 +176,6 @@ export enum RunModelType {
 export interface EnvironmentVariables {
   [key: string]: string | undefined;
 }
+
+export const DefaultQueryTemplate =
```
mdesmet, line 180: If consts use all caps. This looks like a class

src/manifest/dbtProject.ts
mdesmet: Let's not tackle this as part of this PR.

src/manifest/dbtProject.ts
saravmajestic: fix it, pls re review

src/domain.ts
saravmajestic: renamed.

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -143,12 +144,100 @@ export class DBTCoreProjectIntegration
...
+      },
+    );
+
     return this.python!.lock<ExecuteSQLResult>(
-      (python) => python`to_dict(project.execute_sql(${query}))`,
+      (python) => python`to_dict(project.execute_sql(${limitQuery}))`,
     );
   }
 
+  async findLimitQuery(query: string, limit: number) {
```
mdesmet, line 216: should be private method.

anandgupta42:
@saravmajestic Please update the docs also

src/domain.ts
```diff
@@ -176,3 +176,6 @@ export enum RunModelType {
 export interface EnvironmentVariables {
   [key: string]: string | undefined;
 }
+
+export const DEFAULT_QUERY_TEMPLATE =
+  "select * from ({query}\n) as query limit {limit}";
```
mdesmet, line 181: As this is only relevant for dbt core we should move it there.

src/webview_provider/queryResultPanel.ts
```diff
@@ -273,10 +274,7 @@ export class QueryResultPanel implements WebviewViewProvider {
       .get<boolean>("enableNewQueryPanel", true);
     const queryTemplate = workspace
       .getConfiguration("dbt")
-      .get<string>(
-        "queryTemplate",
-        "select * from ({query}) as query limit {limit}",
-      );
```
mdesmet, line 279: Is there any way you can move htis logic out of the webview and into the dbtProject so we get it from the dbtCoreIntegration instead?

saravmajestic: @mdesmet do you mean just this getting query template from configuration or whole logic by getting macro from dbt as well? if it is just the config, I am not sure what is the benefit of moving it to core integration, because this is just getting from config file. If it is about whole logic of macro, that is not possible because it expects a query and limit as arguments which is not available in this context. pls lmk what you think.

query_panel/main.js
```diff
@@ -481,10 +435,6 @@ const app = createApp({
           this.clearData();
           this.focusWelcomePane();
           break;
-        case "renderSummary":
```
saravmajestic, line 484: since summary is moved to datapilot, this is not needed anymore

query_panel/main.js
```diff
@@ -221,69 +220,24 @@ const app = createApp({
...
-        this.summary = data.summary;
-        this.compiledCode = newCompiledCode;
-      }
-    },
     focusSummaryTab() {
       document.querySelector("#panel-manager").activeid = "tab-6";
     },
     updateDispatchedCode(raw_stmt, compiled_stmt) {
       this.rawCode = raw_stmt;
-      let newCompiledCode = compiled_stmt;
```
saravmajestic, line 267: This logic is moved to dbtCorIntegration.ts

mdesmet:
@saravmajestic : can you also take care of the codeQL errors?

mdesmet (approved):

PR #798, opened 2023-12-26
feat: adding test ids

AdiGajbhiye:
## Checklist

### DIFF
--- a/new_lineage_panel/src/App.tsx
+++ b/new_lineage_panel/src/App.tsx
@@ -374,6 +374,7 @@ function App() {
             setMoreTables({});
             vscode.postMessage({ command: "init" });
           }}
+          data-testid="reset-btn"
         >
           Reset
         </Button>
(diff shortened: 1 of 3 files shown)


new_lineage_panel/src/App.tsx
```diff
@@ -374,6 +374,7 @@ function App() {
             setMoreTables({});
             vscode.postMessage({ command: "init" });
           }}
+          data-testid="reset-btn"
```
saravmajestic, line 377: can you use other locators like role as button with name attribute? As per best practices, better to use attributes visible to user rather than these abstracted ones
https://playwright.dev/docs/best-practices#use-locators
https://testing-library.com/docs/queries/bytestid/

saravmajestic (commented):
Can you check the comment on testid?

anandgupta42 (approved):
LGTM

PR #801, opened 2024-01-02
feat: docs generator v2

saravmajestic:
## Overview

- includes common react webview panel which can be extended for other panels. Details on new web view is [here](https://github.com/AltimateAI/vscode-dbt-power-user/tree/feat/docs-v2/webview_panels)
- includes docs gen changes
- includes datapilot side panel 

### How to test:
- Go to documentation editor tab and click `show new ux` button to load new UI

## Checklist


src/manifest/dbtProjectContainer.ts
```diff
@@ -184,6 +184,10 @@ export class DBTProjectContainer implements Disposable {
     return this.context!.extensionUri;
   }
 
+  get extensionPath() {
+    return this.context!.extensionPath;
+  }
+
```
mdesmet, line 190: In other places we use extensionUri, why we need path? Will it work on windows?

src/webview_provider/altimateWebviewProvider.ts
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  private setupWebviewOptions(context: WebviewViewResolveContext) {
+    this._panel!.description = this.panelDescription;
+    this._panel!.webview.options = <WebviewOptions>{
+      enableScripts: true,
+      localResourceRoots: [
+        Uri.file(
+          path.join(
+            this.dbtProjectContainer.extensionPath,
```
mdesmet, line 176: Fishy, should use uri.fesPath or won't work on windows.

src/webview_provider/altimateWebviewProvider.ts
```diff
@@ -0,0 +1,258 @@
...
+          await commands.executeCommand("dbtPowerUser.datapilot-webview.focus");
+          // Adding timeout to let the datapilot complete rendering
+          setTimeout(() => {
+            sharedStateManager.postMessage(message);
+          }, 500);
+          break;
+        case "updateConfig":
+          console.log(
+            "Updating config",
+            (params as UpdateConfigProps).key,
```
mdesmet, line 104: extract params to typed variable.

src/webview_provider/types.ts
```diff
@@ -0,0 +1,5 @@
+export type UpdateConfigProps = {
+  key: string;
+  value: string | boolean | number;
+  isPreviewFeature?: boolean;
+};
```
mdesmet, line 5: If only used in alitmateWebViewProvider. keep it only there and don't export.

src/webview_provider/sharedStateManager.ts
```diff
@@ -0,0 +1,23 @@
...
+  addListener(listener: (message: any) => void) {
+    this.listeners.push(listener);
+  }
+
+  postMessage(message: any) {
+    this.listeners.forEach((listener) => listener(message));
+  }
+}
+
+export const sharedStateManager = SharedStateManager.getInstance();
```
mdesmet, line 23: Why not use vscode event emitters?

src/webview_provider/altimateWebviewProvider.ts
saravmajestic: ok, i ll try with uri.fsPath

src/webview_provider/types.ts
saravmajestic: This is used in datapilot panel as well and will be used in other panels in future if needed

src/webview_provider/sharedStateManager.ts
saravmajestic: I will try that

src/webview_provider/types.ts
mdesmet: but if it is part of the panel it is inherited and will be present on the object that is using it. So I don't immediately see the need to export it and create another file.

src/webview_provider/altimateWebviewProvider.ts
```diff
@@ -0,0 +1,258 @@
...
+              "dbtPowerUser.datapilot-webview.focus",
+            );
+          }
+          break;
+        case "datapilot:message":
+          await commands.executeCommand("dbtPowerUser.datapilot-webview.focus");
+          // Adding timeout to let the datapilot complete rendering
+          setTimeout(() => {
+            sharedStateManager.postMessage(message);
+          }, 500);
```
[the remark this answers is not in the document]
mdesmet: The right way is to let the panel send a message to let it tell you the panel is ready to receive messages.

src/webview_provider/sharedStateManager.ts
saravmajestic: I can build event emitters in 2 ways:
- define a static variable in `AltimateWebviewProvider` and listen to the events in sub classes 
- define a service for event emitter, inject it in AltimateWebviewProvider and sub classes and listen to the events

It is required to use same EventEmitter instance in all classes to listen to the fired events.

I prefer to use static variable as eventemitter will not be mutated and keeps it simple. If we go with service, then need to inject in all classes and use it. Lmk your thoughts @mdesmet

src/webview_provider/altimateWebviewProvider.ts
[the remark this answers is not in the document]
saravmajestic: @mdesmet here is the [commit](https://github.com/AltimateAI/vscode-dbt-power-user/pull/801/commits/de319de462ec0e99432144fa68a1d69760e54024) to address eventemitter and webview ready handler

src/webview_provider/sharedStateManager.ts
mdesmet: I think you just need to extract your eventmitter in one class and inject it as a singleton through inversify.

mdesmet (approved):

PR #828, opened 2024-01-16
fix: safe creation of packages path

suryaiyer95:
## Overview

We are seeing exceptions related to loading of the ALTIMATE_PACKAGES_PATH. To resolve, we are moving it inside the function where the contextmanager loads it. It is wrapped in a try catch block so even if it fails other functionalities are not affected

## Checklist


anandgupta42 (approved):

anandgupta42 (approved):
LGTM

PR #844, opened 2024-01-19
fix: python process exits when unhandled exception exists

suryaiyer95:
## Checklist

### DIFF
--- a/dbt_integration.py
+++ b/dbt_integration.py
@@ -320,9 +320,10 @@ class DbtProject:
                 self.config.load_dependencies(),
                 self.adapter.connections.set_query_header,
             )
+            self.dbt = project_parser.load()
         except Exception as e:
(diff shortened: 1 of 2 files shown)


dbt_integration.py
```diff
@@ -314,11 +314,14 @@ def init_project(self):
...
+            project_parser = ManifestLoader(
+                self.config,
+                self.config.load_dependencies(),
+                self.adapter.connections.set_query_header,
+            )
+        except Exception as e:
+            raise Exception(str(e))
         self.dbt = project_parser.load()
         project_parser.save_macros_to_adapter(self.adapter)
         self.dbt.build_flat_graph()
```
anandgupta42, line 327: why did we not include all these also in the try catch block?

mdesmet (approved):

PR #851, opened 2024-01-23
feat: don't show analysis nodes in autocomplete

svipulrao98:
## Overview
 Analysis nodes should not appear in auto-complete

  Example:
    resolves [AI-1448](https://altimateai.atlassian.net/browse/AI-1448)

## Checklist

[AI-1448]: https://altimateai.atlassian.net/browse/AI-1448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/autocompletion_provider/modelAutocompletionProvider.ts
+++ b/src/autocompletion_provider/modelAutocompletionProvider.ts
@@ -108,9 +108,7 @@ export class ModelAutocompletionProvider
         Array.from(models)
           .filter(
             ([key, model]) =>
-              !model.uniqueId.startsWith(
-                `${DBTProject.RESOURCE_TYPE_ANALYSIS}.`,
-              ),
+              model.resource_type !== DBTProject.RESOURCE_TYPE_ANALYSIS,
           )
           .map(([key, model]) => ({
             label: `(${model.package_name}) ${key}`,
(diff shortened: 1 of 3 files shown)


src/autocompletion_provider/modelAutocompletionProvider.ts
```diff
@@ -104,15 +105,20 @@ export class ModelAutocompletionProvider
...
-            model.package_name === projectName
-              ? key
-              : `${model.package_name}, ${key}`,
-          kind: CompletionItemKind.Value,
-          detail: "Model",
-        })),
+        Array.from(models)
+          .filter(
+            ([key, model]) =>
+              !model.uniqueId.includes(`${DBTProject.RESOURCE_TYPE_ANALYSIS}.`),
```
AdiGajbhiye, line 111: use `startsWith`

src/autocompletion_provider/modelAutocompletionProvider.ts
```diff
@@ -104,15 +105,22 @@ export class ModelAutocompletionProvider
...
-            model.package_name === projectName
-              ? key
-              : `${model.package_name}, ${key}`,
-          kind: CompletionItemKind.Value,
-          detail: "Model",
-        })),
+        Array.from(models)
+          .filter(
+            ([key, model]) =>
+              !model.uniqueId.startsWith(
```
mdesmet, line 111: Please use `model.resource_type === DBTProject.RESOURCE_TYPE_ANALYSIS` instead of relying on uniqueid

mdesmet: @AdiGajbhiye : Maybe also take that into account in other places.

svipulrao98: right now we don't have `resource_type` in model [ref](https://github.com/AltimateAI/vscode-dbt-power-user/blob/master/src/domain.ts#L18-L30)

should we add it in this PR?

mdesmet: Yeah it is better than relying on the uniqueId

mdesmet (approved):

PR #853, opened 2024-01-24
feat: respect env variables

svipulrao98:
## Checklist

[AI-1455]: https://altimateai.atlassian.net/browse/AI-1455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mdesmet:
The migration is missing from the parameter to the env var. Can you try to implement that as well?

dbt_integration.py
```diff
@@ -169,17 +169,19 @@ def wrapper(
...
+        str(project_dir)
+        if (project_dir / "profiles.yml").exists()
+        else str(Path.home() / ".dbt")
     )
 
 
 def target_dir() -> Path:
     if "DBT_TARGET_PATH" in os.environ:
-        return Path(os.environ["DBT_TARGET_PATH"]).resolve()
+        return str(Path(os.environ["DBT_TARGET_PATH"]).resolve())
```
mdesmet, line 184: That won't work for windows

dbt_integration.py
```diff
@@ -169,17 +169,19 @@ def wrapper(
...
-def default_profiles_dir(project_dir: Path) -> Path:
+def default_profiles_dir(project_dir) -> str:
     if "DBT_PROFILES_DIR" in os.environ:
-        return Path(os.environ["DBT_PROFILES_DIR"]).resolve()
+        return str(Path(os.environ["DBT_PROFILES_DIR"]).resolve())
     return (
-        project_dir if (project_dir / "profiles.yml").exists() else Path.home() / ".dbt"
+        str(project_dir)
+        if (project_dir / "profiles.yml").exists()
+        else str(Path.home() / ".dbt")
```
mdesmet, line 178: That won't work for windows

svipulrao98: yes, I was thinking of handling it in the python level itself, or just avoiding `str` altogether

PR #859, opened 2024-01-26
feat: variable to set initialization

svipulrao98:
## Checklist

### DIFF
--- a/src/dbt_client/dbtCoreIntegration.ts
+++ b/src/dbt_client/dbtCoreIntegration.ts
@@ -140,6 +140,7 @@ export class DBTCoreProjectIntegration
   static DBT_PROFILES_FILE = "profiles.yml";
 
   private profilesDir?: string;
+  private hasManifest: boolean = false;
   private targetPath?: string;
   private adapterType?: string;
   private version?: number[];
@@ -185,6 +186,7 @@ export class DBTCoreProjectIntegration
   }
 
   async refreshProjectConfig(): Promise<void> {
+    this.hasManifest = false;
     await this.createPythonDbtProject();
     await this.python.ex`project.init_project()`;
     this.targetPath = await this.findTargetPath();
@@ -318,6 +320,7 @@ export class DBTCoreProjectIntegration
         (python) => python`to_dict(project.safe_parse_project())`,
       );
       this.rebuildManifestDiagnostics.clear();
+      this.hasManifest = true;
     } catch (exc) {
       if (exc instanceof PythonException) {
         // dbt errors can be about anything, so we just associate the error with the project file
@@ -554,7 +557,7 @@ export class DBTCoreProjectIntegration
     );
   }
 
-  private throwBridgeErrorIfAvailable() {
+  throwBridgeErrorIfAvailable() {
     const allDiagnostics = [
       this.pythonBridgeDiagnostics,
       this.rebuildManifestDiagnostics,
@@ -567,6 +570,11 @@ export class DBTCoreProjectIntegration
           throw new Error(firstError.message);
         }
       }
+    }
+    if (!this.hasManifest) {
+      throw new Error(
+        "The dbt project is still initializing, please try again later.",
+      );
     }
   }
 
--- a/src/dbt_client/dbtIntegration.ts
+++ b/src/dbt_client/dbtIntegration.ts
@@ -259,6 +259,7 @@ export interface DBTProjectIntegration extends Disposable {
   ): Promise<{ [key: string]: string }[]>; // TODO: this should be typed
   getColumnsOfModel(modelName: string): Promise<{ [key: string]: string }[]>; // TODO: this should be typed
   getCatalog(): Promise<{ [key: string]: string }[]>; // TODO: this should be typed
+  throwBridgeErrorIfAvailable(): void;
 }
 
 @provide(DBTCommandExecutionInfrastructure)
--- a/src/manifest/dbtProject.ts
+++ b/src/manifest/dbtProject.ts
@@ -374,7 +374,7 @@ export class DBTProject implements Disposable {
   async compileQuery(query: string): Promise<string | undefined> {
     this.telemetry.sendTelemetryEvent("compileQuery");
     try {
-      return this.dbtProjectIntegration.unsafeCompileQuery(query);
+      return await this.dbtProjectIntegration.unsafeCompileQuery(query);
     } catch (exc: any) {
       if (exc instanceof PythonException) {
         window.showErrorMessage(
@@ -428,6 +428,10 @@ export class DBTProject implements Disposable {
 
   async getColumnsOfSource(sourceName: string, tableName: string) {
     return this.dbtProjectIntegration.getColumnsOfSource(sourceName, tableName);
+  }
+
+  throwPythonBridgeError() {
+    return this.dbtProjectIntegration.throwBridgeErrorIfAvailable();
   }
 
   async getCatalog(): Promise<{ [key: string]: string }[]> {
--- a/src/webview_provider/newLineagePanel.ts
+++ b/src/webview_provider/newLineagePanel.ts
@@ -84,7 +84,6 @@ export class NewLineagePanel implements LineagePanelView {
 
   eventMapChanged(eventMap: Map<string, ManifestCacheProjectAddedEvent>): void {
     this.eventMap = eventMap;
-    this.init();
   }
 
   changedActiveColorTheme() {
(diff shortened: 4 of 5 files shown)


src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -204,6 +205,10 @@ export class DBTCoreProjectIntegration
     );
   }
 
+  updateInitializationState(): void {
+    this.isInitialized = true;
+  }
```
mdesmet, line 210: Why does this need to be interface method and can't it just be a internal thing.

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -343,6 +348,7 @@ export class DBTCoreProjectIntegration
             adapter: this.getAdapterType() || "unknown", // TODO: this should be moved to dbtProject
           },
         );
+        this.updateInitializationState();
```
mdesmet, line 351: This will only error out becasue of errors in dbtCore integration.

src/dbt_client/dbtIntegration.ts
```diff
@@ -220,6 +220,8 @@ export interface DBTProjectDetection extends Disposable {
 export interface DBTProjectIntegration extends Disposable {
   // initialize execution infrastructure
   initializeProject(): Promise<void>;
+  // update initialization
+  updateInitializationState(): void;
```
mdesmet, line 224: Remove.

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -145,6 +145,7 @@ export class DBTCoreProjectIntegration
 {
   static DBT_PROFILES_FILE = "profiles.yml";
 
+  private isInitialized?: boolean = false;
```
mdesmet, line 148: If you set it to false, why you need the `?`?

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -343,6 +344,7 @@ export class DBTCoreProjectIntegration
             adapter: this.getAdapterType() || "unknown", // TODO: this should be moved to dbtProject
           },
         );
+        this.isInitialized = true;
```
mdesmet, line 347: maybe rename to `hasManifest` as this code will be hit a lot more, also what happens when project config is changed. Then manifest and config are erased but i don't see any code handling this.

svipulrao98: I will add a check, if `hasManifest` is true, then we don't `rebuildManifestDiagnostics` during `rebuildManifest`

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -556,7 +560,7 @@ export class DBTCoreProjectIntegration
     );
   }
 
-  private throwBridgeErrorIfAvailable() {
+  private async throwBridgeErrorIfAvailable() {
```
AdiGajbhiye, line 563: No need for async here. As you are not awaiting anything in function body.

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -570,6 +574,11 @@ export class DBTCoreProjectIntegration
         }
       }
     }
+    if (this.hasManifest) {
```
AdiGajbhiye, line 577: should be `!this.hasManifest`, right?

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -1,4 +1,5 @@
 import {
+  commands,
```
AdiGajbhiye, line 2: is it getting used anywhere?

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -570,6 +573,12 @@ export class DBTCoreProjectIntegration
         }
       }
     }
+    if (!this.hasManifest) {
+      const initErrorMessage =
+        "Operation cannot proceed while the manifest is being built.";
```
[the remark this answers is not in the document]
svipulrao98: thanks!

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -570,6 +573,12 @@ export class DBTCoreProjectIntegration
         }
       }
     }
+    if (!this.hasManifest) {
+      const initErrorMessage =
+        "The dbt Project is still initializing, please try again later.";
+      window.showErrorMessage(initErrorMessage);
```
mdesmet, line 579: It's not logic to both throw and show an error message. the error message must be handled somewhere upstream already.

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -570,6 +573,11 @@ export class DBTCoreProjectIntegration
         }
       }
     }
+    if (!this.hasManifest) {
+      const initErrorMessage =
```
mdesmet, line 577: No need for variable

src/webview_provider/newLineagePanel.ts
```diff
@@ -264,39 +264,45 @@ export class NewLineagePanel implements LineagePanelView {
...
-      };
-    });
-    if (Object.keys(node.columns).length > columnsFromDB.length) {
-      // Flagging events where columns fetched from db are less than the number of columns in the manifest
-      this.telemetry.sendTelemetryEvent("columnLineagePossibleStaleSchema");
+      return true;
+    } catch (exc) {
+      const error = String(exc);
+      window.showErrorMessage("Cannot add columns from DB: " + error);
+      throw new Error(error);
```
mdesmet, line 304: Why throw again?

Also this doesn't seem the right error boundary as this method is used in loops.

src/webview_provider/newLineagePanel.ts
```diff
@@ -297,12 +297,10 @@ export class NewLineagePanel implements LineagePanelView {
...
         this.telemetry.sendTelemetryEvent("columnLineagePossibleStaleSchema");
       }
-      return true;
     } catch (exc) {
-      const error = String(exc);
-      window.showErrorMessage("Cannot add columns from DB: " + error);
-      throw new Error(error);
+      window.showErrorMessage("Cannot add columns from DB: " + exc);
     }
+    return true;
```
mdesmet, line 303: @AdiGajbhiye : Shouldn't the exception clause not return false, and return true should be within the try block?

svipulrao98: @mdesmet yes you are right, we should return false, but upon doing so we are just [returning here](https://github.com/AltimateAI/vscode-dbt-power-user/blob/9e2afe35858dbb522b19ff89ddfecde6551a15a8/src/webview_provider/newLineagePanel.ts#L491), so the panel goes blank. might result in some breaking changes, so I kept this change for now
image

mdesmet: @AdiGajbhiye : Can you lean in on this one?

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -570,6 +573,11 @@ export class DBTCoreProjectIntegration
         }
       }
     }
+    if (!this.hasManifest) {
+      throw new Error(
+        "The dbt Project is still initializing, please try again later.",
```
mdesmet, line 578:
```suggestion
        "The dbt project is still initializing, please try again later.",
```

src/webview_provider/newLineagePanel.ts
```diff
@@ -264,37 +264,41 @@ export class NewLineagePanel implements LineagePanelView {
...
-        name: c.column,
-        data_type: c.dtype,
-        description: "",
-      };
-    });
-    if (Object.keys(node.columns).length > columnsFromDB.length) {
-      // Flagging events where columns fetched from db are less than the number of columns in the manifest
-      this.telemetry.sendTelemetryEvent("columnLineagePossibleStaleSchema");
+    } catch (exc) {
+      window.showErrorMessage("Cannot add columns from DB: " + exc);
```
mdesmet, line 301: We should add telemetry here.

src/manifest/dbtProject.ts
```diff
@@ -419,7 +419,11 @@ export class DBTProject implements Disposable {
   }
 
   async unsafeCompileQuery(query: string) {
-    return this.dbtProjectIntegration.unsafeCompileQuery(query);
+    try {
+      return this.dbtProjectIntegration.unsafeCompileQuery(query);
+    } catch (exc) {
+      throw new Error(String(exc));
+    }
```
mdesmet, line 426: No need to retrhow exception. it needs to be captured in the calling code.

src/webview_provider/newLineagePanel.ts
```diff
@@ -268,7 +268,12 @@ export class NewLineagePanel implements LineagePanelView {
   }
 
   private async addModelColumnsFromDB(project: DBTProject, node: NodeMetaData) {
-    const columnsFromDB = await project.getColumnsOfModel(node.name);
+    let columnsFromDB;
+    try {
+      columnsFromDB = await project.getColumnsOfModel(node.name);
+    } catch (error) {
+      return false;
+    }
```
mdesmet, line 276: If you ignore the exception here it will be lost. This seems wrong.

src/webview_provider/newLineagePanel.ts
```diff
@@ -551,7 +557,14 @@ export class NewLineagePanel implements LineagePanelView {
...
     }
-    const dbColumnAdded = await this.addModelColumnsFromDB(project, node);
+    let dbColumnAdded = false;
+    try {
+      await this.addModelColumnsFromDB(project, node);
+    } catch (error) {
+      window.showErrorMessage(extendErrorWithSupportLinks(String(error)));
+      this.telemetry.sendTelemetryError("columnLineageGetModelColumns");
+    }
+    dbColumnAdded = true;
```
mdesmet, line 567: This is called in a loop so, will show multiple error messages. please fix

PR #860, opened 2024-01-26
feat: added logging service

AdiGajbhiye:
WIP

fix: logging service

WIP

WIP

feat: combined logging and terminal

feat: added strip ansi

feat: optional telemetry

WIP

fix: as per comments

WIP

## Checklist

### DIFF
--- a/src/dbt_client/dbtTerminal.ts
+++ b/src/dbt_client/dbtTerminal.ts
@@ -1,11 +1,22 @@
-import { EventEmitter, Terminal, window } from "vscode";
-import { provideSingleton, stripANSI } from "../utils";
+import { Disposable, EventEmitter, Terminal, window } from "vscode";
+import {
+  extendErrorWithSupportLinks,
+  provideSingleton,
+  stripANSI,
+} from "../utils";
+import { CustomException } from "./exception";
+import { TelemetryService } from "../telemetry";
 
 @provideSingleton(DBTTerminal)
 export class DBTTerminal {
+  private disposables: Disposable[] = [];
   private terminal?: Terminal;
   private readonly writeEmitter = new EventEmitter<string>();
-  private outputChannel = window.createOutputChannel(`Log - dbt`);
+  private outputChannel = window.createOutputChannel(`Log - dbt`, {
+    log: true,
+  });
+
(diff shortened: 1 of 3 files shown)


src/dbt_client/dbtTerminal.ts
```diff
@@ -15,12 +26,57 @@ export class DBTTerminal {
...
+  warn(e: CustomException, sendTelemetry: boolean = false) {
+    const errorMessage = e.getErrorMessage();
+    this.outputChannel?.warn(stripANSI(errorMessage));
+    console.warn(errorMessage);
+    if (sendTelemetry) {
+      this.telemetry.sendTelemetryError(e.name, e.error, e.extra);
+    }
+  }
+
+  error(e: CustomException, sendTelemetry = false) {
```
mdesmet, line 63: by default sendTelemetry of errors.

src/dbt_client/dbtTerminal.ts
```diff
@@ -15,12 +26,57 @@ export class DBTTerminal {
...
+    const errorMessage = e.getErrorMessage();
+    this.outputChannel?.error(stripANSI(errorMessage));
+    window.showErrorMessage(extendErrorWithSupportLinks(e.message));
```
mdesmet, line 66: Showing error messages goes beyond what a terminal is supposed to do. Maybe think of a better name. Why not keep this part where it was?

src/webview_provider/docsEditPanel.ts
```diff
@@ -272,23 +278,23 @@ export class DocsEditViewPanel implements WebviewViewProvider {
...
-                    this.telemetry.sendTelemetryError(
-                      "docsEditPanelLoadPythonError",
-                      exc,
+                    this.terminal.error(
+                      new CustomPythonException(
+                        "docsEditPanelLoadPythonError",
+                        `An error occured while fetching metadata for ${modelName} from the database`,
+                        exc,
+                      ),
+                      true,
```
mdesmet, line 287: These error messages were written from the PoV of a error message following a user action. They are not necessarily written for logging (they might miss a lot of things like which project etc).

I think it is maybe better to keep these two concerns separate.

mdesmet (approved):

PR #797, opened 2023-12-26
feat: defer to prod

saravmajestic:
## Overview
Implements defer to prod feature and 2 other modules
- added 3 configs `deferToProduction, favorState, manifestPathForDeferral` for storing the defer state and applying the same while running `dbt build`
- Bigquery cost estimate module
- Project health checker
- created common react webview panel which can be extended for other panels. Details on new web view is [here](https://github.com/AltimateAI/vscode-dbt-power-user/tree/feat/defer/webview_panels)

#### Defer
- added the extra parameters (if provided by user in insights panel) to the following commands
     - build
     - run
     - test
     - testModel

## Checklist

- [x] I have run this code and it appears to resolve the stated issue
- [ ] `README.md` updated and added information about my change

PR #867, opened 2024-01-30
feat: defer to prod and insights panel

saravmajestic:
## Overview
Original PR: https://github.com/AltimateAI/vscode-dbt-power-user/pull/797/files

Implements defer to prod feature and 2 other modules
- added config `deferConfigPerProject` with 3 sub configs `deferToProduction, favorState, manifestPathForDeferral` for storing the defer state per project and applying the same while running dbt commands listed below
- Bigquery cost estimate module
- Project health checker

#### Defer
- added the extra parameters (if provided by user in insights panel) to the following commands
     - build
     - run
     - test
     - testModel
     
#### Pending 
- [ ] We should also add a new action to "build project" in the run button dropdown. That way they can run the project with defer prod with cost saving
- [ ] verify the defer config save in projects which is added to workspace with multiple projects from different directories
- [ ] if above line item is not working, then we need to add a parameter to dbtProject that refers to the workspace where it has been found in
- [x] Bigquery cost estimate results to show up in right side of insights panel (@saravmajestic have to do this, just need to copy the changes before the refactor PR merge)

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -168,6 +168,32 @@
             },
             "default": [],
             "description": "Add additional params to `sqlfmt`"
+          },
+          "dbt.deferConfigPerProject": {
+            "type": "object",
+            "description": "Run subset of models without building their parent models",
+            "patternProperties": {
+              ".*": {
+                "type": "object",
+                "properties": {
+                  "deferToProduction": {
+                    "type": "boolean",
+                    "description": "Run subset of models without building their parent models",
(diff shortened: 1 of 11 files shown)


saravmajestic:
Closing in favor of: https://github.com/AltimateAI/vscode-dbt-power-user/pull/881

PR #873, opened 2024-01-31
feat: fallback to read profiles if python fails

svipulrao98:
## Overview
Read custom package reader if python fails to detect

## Checklist


src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -32,7 +32,8 @@ import { CommandProcessExecutionFactory } from "../commandProcessExecution";
 import { PythonBridge, PythonException } from "python-bridge";
 import * as path from "path";
 import { DBTProject } from "../manifest/dbtProject";
-import { existsSync } from "fs";
+import { existsSync, readFileSync } from "fs";
+import * as yaml from "js-yaml";
```
mdesmet, line 36: We are already using yaml library to read yaml in other places. let's not introduce another library for the sake of it.

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -85,6 +86,27 @@ export class DBTCoreProjectDetection
...
+  private getPackageInstallPathFallback(projectDirectory: Uri): string {
+    const dbtProjectFile = path.join(
+      projectDirectory.fsPath,
+      "dbt_project.yml",
+    );
+    if (existsSync(dbtProjectFile)) {
+      const dbtProjectConfig: any = yaml.load(
+        readFileSync(dbtProjectFile, "utf8"),
+      );
+      const packagesInstallPath = dbtProjectConfig["packages-install-path"];
```
mdesmet, line 98: Can you also check if this path is existing as it might be jinja based?

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -85,6 +86,32 @@ export class DBTCoreProjectDetection
...
+  private getPackageInstallPathFallback(projectDirectory: Uri): string {
+    const dbtProjectFile = path.join(
+      projectDirectory.fsPath,
+      "dbt_project.yml",
+    );
+    if (existsSync(dbtProjectFile)) {
+      const dbtProjectConfig: any = parse(readFileSync(dbtProjectFile, "utf8"));
+      const packagesInstallPath = dbtProjectConfig["packages-install-path"];
+      if (packagesInstallPath) {
+        let absolutePath;
```
mdesmet, line 98: Maybe you can name this one packageInstallPath and above packageInstallPathFromYaml.

Anyway absolutepath is a bit of a generic name.

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -85,6 +86,32 @@ export class DBTCoreProjectDetection
...
+            projectDirectory.fsPath,
+            packagesInstallPath,
+          );
+        }
+        if (existsSync(packageInstallPathFromYaml)) {
+          return packageInstallPathFromYaml;
+        }
+      }
+    }
+    return path.join(projectDirectory.fsPath, "dbt_packages");
```
saravmajestic, line 112: should we also check if `dbt_packages` directory exists?

svipulrao98: okay, should we create one if it doesn't exist?

mdesmet: No need for creation. If no project file exists, there is no project anyway, so not sure how we would even land there.

PR #880, opened 2024-02-01
Feat/target path profiles dir 

mdesmet:
## Checklist

### DIFF
--- a/dbt_integration.py
+++ b/dbt_integration.py
@@ -101,8 +101,7 @@ def validate_sql(
 ):
     try:
         ALTIMATE_PACKAGE_PATH = os.path.join(
-            os.path.dirname(os.path.abspath(__file__)),
-            "altimate_packages"
+            os.path.dirname(os.path.abspath(__file__)), "altimate_packages"
         )
         with add_path(ALTIMATE_PACKAGE_PATH):
             from altimate.validate_sql import validate_sql_from_models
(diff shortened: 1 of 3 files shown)


src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -527,6 +511,21 @@ export class DBTCoreProjectIntegration
...
+    const profilesDir = await this.python.lock(
+      (python) =>
+        python`default_profiles_dir(Path(${this.projectRoot.fsPath}))`,
+    );
+    return profilesDir;
+  }
+
+  private async getTargetDirectory(): Promise<string> {
+    const targetDir = await this.python.lock(
+      (python) => python`target_path(${this.projectRoot.fsPath})`,
```
mdesmet, line 524: This line before was `python`target_path(Path(${this.projectRoot.fsPath}))`,`, That's the only place where we were using a pathlib PosixPath. It was not necessary and for some reason it didn't throw any errors in local testing.

dbt_integration.py
```diff
@@ -169,25 +167,48 @@ def wrapper(
...
-def find_package_paths(project_directories, profiles_dir_override):
+def default_profiles_dir(project_dir):
+    if "DBT_PROFILES_DIR" in os.environ:
+        profiles_dir = os.path.expanduser(os.environ["DBT_PROFILES_DIR"])
+        if os.path.isabs(profiles_dir):
+            return os.path.normpath(profiles_dir)
+        return os.path.join(project_dir, profiles_dir)
+    project_profiles_file = os.path.normpath(os.path.join(project_dir, "profiles.yml"))
+    return (
+        project_dir
```
suryaiyer95, line 179: We need to convert this to string right?

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -527,6 +511,21 @@ export class DBTCoreProjectIntegration
     return macroPaths;
   }
 
+  private async findProfilesDirectory(): Promise<string> {
+    const profilesDir = await this.python.lock(
+      (python) =>
+        python`default_profiles_dir(Path(${this.projectRoot.fsPath}))`,
```
suryaiyer95, line 517: I think, this is the one causing the issue.

dbt_integration.py
mdesmet: Nope this should all be strings. os.path operations operatate on strings.

suryaiyer95 (commented):
Tested cases and passed

1. Target path override
- dbt_project.yml 
     - target-path: 'compiled_files'
     - target-path: '../compiled_files'
-  DBT_TARGET_PATH environment variable


2. Profiles dir override
- DBT_PROFILES_DIR
      - '../../.db/profiles.yml'
      - './profiles.yml'
      - '~/.dbt'

anandgupta42 (approved):
LGTM

PR #887, opened 2024-02-05
feat: auto expand lineage

AdiGajbhiye:
## Overview

- added auto expansion widget
- change table node ui
- code refactor

Screenshot 2024-02-07 at 1 21 54 PM

## Checklist


saravmajestic:
@AdiGajbhiye can you record a video explaining the changes so it will be easy to review?

new_lineage_panel/src/ActionWidget.tsx
```diff
@@ -0,0 +1,392 @@
...
+  const [nodeCount, setNodeCount] = useState(0);
+  const [maxRange, setMaxRange] = useState([0, 0]);
+
+  useEffect(() => {
+    setLeftExpansion(Math.max(minRange[0], 0));
+    setRightExpansion(Math.max(minRange[1], 0));
+  }, [minRange, setLeftExpansion, setRightExpansion]);
+
+  useEffect(() => {
+    (async () => {
```
saravmajestic, line 96: it is better to avoid this pattern of self invoking async functions inside useEffect to avoid repeatedly calling the async function. pls extract into useCallback function and invoke it here

new_lineage_panel/src/ActionWidget.tsx
```diff
@@ -0,0 +1,392 @@
...
+        selectedTable,
+        level - leftExpansion,
+        level + rightExpansion
+      );
+      setNodeCount(newNodes.length - startingNodesNum);
+    })();
+  }, [flow, minRange, leftExpansion, rightExpansion, selectedTable]);
+
+  useEffect(() => {
+    (async () => {
```
saravmajestic, line 116: here also, pls extract this method

new_lineage_panel/src/ActionWidget.tsx
```diff
@@ -0,0 +1,392 @@
...
+import { LineageContext, aiEnabled } from "./App";
+import { useReactFlow } from "reactflow";
+import {
+  calculateMinLevel,
+  expandTableLineageLevelWise,
+  highlightTableConnections,
+  layoutElementsOnCanvas,
+} from "./graph";
+import classNames from "classnames";
+import { BetterPopover } from "./components/Modal/BetterPopover";
```
saravmajestic, line 34: try to use constant directory naming pattern. `components` start with lowercase, but Modal with uppercase

new_lineage_panel/src/ActionWidget.tsx
```diff
@@ -0,0 +1,392 @@
...
+
+const InfoIcon: FunctionComponent<{ id: string; message: string }> = ({
```
saravmajestic, line 36: This is one big file with too many functions!

new_lineage_panel/src/ActionWidget.tsx
```diff
@@ -0,0 +1,392 @@
...
+          <div className="w-100 d-flex gap-xl justify-content-between align-items-center">
+            <div className="normal-text">Parents</div>
+            <div className="normal-text">Children</div>
+          </div>
+          <Button
+            color={nodeCount === 0 ? "secondary" : "primary"}
+            size="sm"
+            disabled={nodeCount === 0}
+            onClick={async () => {
+              if (!selectedTable) return;
```
saravmajestic, line 220: Good to see the early return!

new_lineage_panel/src/ActionWidget.tsx
```diff
@@ -0,0 +1,392 @@
...
+          >
+            Add {nodeCount} tables
+          </Button>
+        </div>
+      )}
+    />
+  );
+};
+
+export const ActionWidget: FC<{
```
saravmajestic, line 247: is it necessary to invoke this as FC?

new_lineage_panel/src/ActionWidget.tsx
```diff
@@ -0,0 +1,392 @@
...
+                      setNonSelectCheck(e.target.checked);
+                    }}
+                  />
+                  <Label check for="non-select-check">
+                    Non-Select
+                  </Label>
+                  <InfoIcon
+                    id="non_select_lineage"
+                    message={
+                      "Non-Select linkages are shown if columns appear " +
```
saravmajestic, line 314: Could be replaced with string literals?

new_lineage_panel/src/components/Column.tsx
```diff
@@ -0,0 +1,111 @@
+import { FunctionComponent, ReactElement, useMemo } from "react";
+import BooleanIcon from "../assets/icons/boolean.svg?react";
```
saravmajestic, line 2: you could follow [this pattern](https://github.com/AltimateAI/vscode-dbt-power-user/blob/master/webview_panels/src/assets/icons/index.ts#L1) to avoid multiple imports

new_lineage_panel/src/components/Column.tsx
```diff
@@ -0,0 +1,111 @@
...
+      <Icon width={size} height={size} />
+    </div>
+  );
+};
+
+export const NodeTypeIcon: FunctionComponent<{ nodeType: string }> = ({
+  nodeType,
+}) => (
+  <div>
+    {nodeType === "seed" && <SeedIcon />}
```
saravmajestic, line 65: performance tip: this code will execute all nodeType checks for every render. You could either return icon on match or create a json object with nodetype and icon and just return in one line
first approach:
```
export const NodeTypeIcon: FunctionComponent<{ nodeType: string }> = ({
  nodeType,
}) => {
if (nodeType === 'seed'){
return <seedIcon />
}
...
...
}
```

second approach:
```
const IconMap = {
seed: <SeedIcon />,
...
...
}
export const NodeTypeIcon: FunctionComponent<{ nodeType: string }> = ({
  nodeType,
}) => (
  <div>
{IconMap[nodeType]}
</div>
```

new_lineage_panel/src/components/Column.tsx
```diff
@@ -0,0 +1,111 @@
...
+  <>
+    <div className={styles.table_node_pill} id={id}>
+      <div className={styles.icon}>{icon}</div>
+      <div>{text}</div>
+    </div>
+    <UncontrolledTooltip target={id}>{label}</UncontrolledTooltip>
+  </>
+);
+
+export const NODE_TYPE_STYLES = {
```
saravmajestic, line 91: you could merge NODE_TYPE_STYLES, NODE_TYPE_SHORTHAND and NodeTypeIcon into single json map, so it will be single place to manage and if we need to add more properties, we can keep adding to same json, instead of creating separate constants

new_lineage_panel/src/styles.module.scss
```diff
@@ -343,3 +343,49 @@ $size: 1.2em;
...
+  border-radius: 4px;
+  border: var(--table-node-border-color) solid 1px;
+  display: flex;
+  align-items: center;
+  background-color: var(--bg-color);
+  color: var(--text-color);
+}
+
+.expand_nav_left {
+  @extend .expand-nav;
```
saravmajestic, line 357: this is new for me. I will explore what this is

new_lineage_panel/src/graph.ts
```diff
@@ -635,3 +716,41 @@ export const bfsTraversal = async (
   }
   return isLineage;
 };
+
+export const moveTableFromSeeMoreToCanvas = (
```
saravmajestic, line 720: some comments for each of these methods will be useful. I dont think anyone except @AdiGajbhiye can fix issues in this flow :)

saravmajestic: Will be really useful if we have unit tests atleast, so we could have documented different usecases handled in these methods @anandgupta42

new_lineage_panel/src/App.tsx
```diff
@@ -158,19 +126,22 @@ function App() {
...
+  const [rightExpansion, setRightExpansion] = useState(0);
   const [collectColumns, setCollectColumns] = useState<
     Record<string, string[]>
   >({});
-  const [confidence, setConfidence] = useState<{
-    confidence: string;
-    operator_list?: string[];
-  }>({ confidence: "high" });
+  const [confidence, setConfidence] = useState<Confidence>({
+    confidence: "high",
```
saravmajestic, line 135: if the value for this confidence will be set of strings like high or low only, using it as an enum could be better

saravmajestic (changes requested):
Added my comments

mdesmet (approved):

PR #889, opened 2024-02-06
fix: use float instead of int in query results to avoid overflow

saravmajestic:
## Overview
- Fixes integer overflow in query results panel new UI
- perspective.js does not handle int64 and workaround suggested is to convert the int to float and use without precision. https://github.com/finos/perspective/issues/1346
- So for any integer type in schema, we will use float type with 0 precision digits and without grouping to avoid thousands separator

## Checklist

### DIFF
--- a/query_panel/main.js
+++ b/query_panel/main.js
@@ -69,7 +69,7 @@ class Grid {
     await this.elem.load(table);
     await this.elem.restore({
       columns: [], // reset columns
-      settings: false,
+      settings: true,
       title: "query result",
       plugin_config: { editable: false },
     });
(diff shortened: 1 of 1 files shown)


query_panel/main.js
```diff
@@ -67,7 +69,7 @@ class Grid {
     await this.elem.load(table);
     await this.elem.restore({
       columns: [], // reset columns
-      settings: true,
+      settings: false,
```
mdesmet, line 72: Why this change?

saravmajestic: This is not intentional. good catch!

mdesmet (approved):

PR #896, opened 2024-02-09
feat: default insights panel

gaurpulkit:
## Checklist

### DIFF
--- a/webview_panels/src/modules/defer/DeferToProduction.tsx
+++ b/webview_panels/src/modules/defer/DeferToProduction.tsx
@@ -221,6 +221,10 @@ const ManifestSelection = ({
   );
 };
 
+const DefaultInsightsPanel = (): JSX.Element => {
+  return <></>;
+};
+
 const DeferToProduction = (): JSX.Element => {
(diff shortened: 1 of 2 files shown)


webview_panels/src/modules/defer/DeferToProduction.tsx
```diff
@@ -240,13 +244,17 @@ const DeferToProduction = (): JSX.Element => {
...
   });
+  const [showPanel, setShowPanel] = useState(false);
 
   const onMesssage = useCallback(
     (event: MessageEvent<IncomingMessageProps>) => {
       const { command, args } = event.data;
       switch (command) {
         case "updateDeferConfig":
-          setDeferState(args as unknown as DeferToProductionProps);
+          if (args) {
```
saravmajestic, line 254: here you are checking falsy value for args and if there is no project you are sending `""` from webview provider. so this if loop will not be executted right? then the deferstate variable will not be unset.

mdesmet:
@saravmajestic , @gaurpulkit : is this PR still relevant?

PR #901, opened 2024-02-13
feat: Ai 1456 custom packages

svipulrao98:
## Checklist


mdesmet:
@svipulrao98 : can you rebase this PR?

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -82,6 +85,25 @@ export class DBTCoreProjectDetection
...
+      const packagesInstallPath = dbtProjectConfig["packages-install-path"];
+      if (packagesInstallPath) {
+        if (path.isAbsolute(packagesInstallPath)) {
+          return packagesInstallPath;
+        } else {
+          return path.join(projectDirectory.fsPath, packagesInstallPath);
+        }
+      }
+    }
+    return path.join(projectDirectory.fsPath, "dbt_packages");
```
mdesmet, line 104: On line 109 we already do the same. I think we don't need to return this. and just only do the work if the fallback returns something.

PR #904, opened 2024-02-14
feat: query result sharing

AdiGajbhiye:
## Overview
- Added new functionality to enable query sharing, allowing users to upload query preview results to the Altimate frontend.
- Added a new boolean property dbt.enableQuerySharing to the package.json file with a default value of false. 

Screenshot 2024-02-21 at 1 07 07 PM

## Checklist


src/webview_provider/queryResultPanel.ts
```diff
@@ -185,6 +187,61 @@ export class QueryResultPanel implements WebviewViewProvider {
...
+            );
+            break;
+          case InboundCommand.InviteUser:
+            const { role, email } = message;
+            const company = workspace
+              .getConfiguration("dbt")
+              .get<string>("altimateInstanceName");
+            if (!company) {
+              window.showErrorMessage(
+                "Please instance name and API key for inviting user.",
```
saravmajestic, line 223:
```suggestion
                "Please add instance name and API key for inviting user.",
```

saravmajestic: why dont you use `handlePreviewFeatures` check instead of checking just instance name?

saravmajestic (commented):
Just minor text change. codewise it looks fine. I dont have any idea about the functionality.

PR #905, opened 2024-02-15
fix: query analysis ux issues

saravmajestic:
## Overview
Fixes following issues
- compile sql for followup api request
- icons (ask followup qn submit button, sync with database refresh icon) colors fix
- apply theme for markdown renderer based on current vscode theme
- wrap the followup qns button

## Checklist

### DIFF
--- a/webview_panels/src/modules/app/appSlice.ts
+++ b/webview_panels/src/modules/app/appSlice.ts
@@ -1,15 +1,15 @@
 import { createSlice, PayloadAction } from "@reduxjs/toolkit";
-import { AppStateProps } from "./types";
+import { AppStateProps, Themes } from "./types";
 
 export const initialState = {
-  theme: "",
+  theme: Themes.Dark,
 } as AppStateProps;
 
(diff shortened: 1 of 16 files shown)


webview_panels/src/modules/app/appSlice.ts
```diff
@@ -1,15 +1,15 @@
 import { createSlice, PayloadAction } from "@reduxjs/toolkit";
-import { AppStateProps } from "./types";
+import { AppStateProps, Themes } from "./types";
 
 export const initialState = {
-  theme: "",
+  theme: Themes.Dark,
```
anandgupta42, line 5: @saravmajestic is this intentional?

saravmajestic: @anandgupta42 yes, i have to apply theme for the markdown based on vscode theme. This initial value value will be overwritten after the page is loaded with right value

anandgupta42 (approved):
LGTM

PR #909, opened 2024-02-16
fix: add logging with extra args

saravmajestic:
## Overview
Problem: In the current logging service, if we need to pass extra arguments/meta data for logs, use JSON.stringify and add to the log message. This will be difficult to read and parse later.

Solution: Use the same parameters as outputChannel to accept arbitrary array of data and write to channel.

## Checklist


AdiGajbhiye (approved):
lgtm

anandgupta42 (approved):
LGTM

PR #918, opened 2024-02-20
fix: make horizontal scrollbar visible in query panel

saravmajestic:
### Problem
Horizontal scrollbar in query results panel is not visible and make it harder to see the columns out of viewport

### Solution
Horizontal scrollbar is already available in perspective viewer table but the height is limited by the border. In this PR, border width is reduced to make the scrollbar more prominent

### How to verify
- open query results panel and run a query which can return more columns in result
- horizontal scrollbar should be visible as in screenshot and user should be able to scroll

### Screenshot/demo
image

## Checklist

### DIFF
--- a/query_panel/style.css
+++ b/query_panel/style.css
@@ -143,3 +143,7 @@ th {
 .flex-column {
   flex-direction: column;
 }
+
+regular-table::-webkit-scrollbar-thumb {
+  border-width: 2px;
+}


anandgupta42 (approved):
Screenshot 2024-02-19 at 10 35 51 PM

LGTM

svipulrao98 (approved):

PR #919, opened 2024-02-20
fix: wrong formatting (#856)

### LINKED ISSUE #856: One specific BQ-DBT SQL file gets wrongly formatted and becomes unusable
### Summary

I am considering using sqlfmt and found out that when I format it with sqlfmt, it makes it unusable because it accidentally comments out code that was in used.
This is not a critical bug since playing around with the file before running sqlfmt solves the issue.
I also created an issue in sqlfmt repo but I actually can't reproduce it with their command line tool. It's only when I run it through VSCode following your documentation on sqlfmt.
I will send you the file via the Chat because I cannot share it publicly.
I think you are simply using sqlfmt with different parameter as what I'm using, but I can't see which parameter you add and thus I cannot raise the issue to them "properly":
https://github.com/tconbeer/sqlfmt/issues/554

### Expected behavior

Expected: The file is formatted and the code is still valid.

### Actual behavior

The DBT-jinja code is invalid.

### Steps To Reproduce

Put the file I sent in a VSCode repo.
Notice that the comments from line 99 to 104 use the {##} syntax to have DBT-jinja comments that are not rendered in the final SQL code.
Setup "dbt.sqlFmtAdditionalParams": ["-l", "120"]
Right-click: Format Document.
Notice that the comment mentioned above now extended from line 74 to line 104 and thus breaking the code.

### Log output/Screenshots

_No response_

### Operating System

Ubuntu WSL (Linux VM in Windows)

### dbt version

1.5.1

### dbt Adapter

BigQuery

### dbt Power User version

v0.30.0

- [X] Yes I am willing to submit a PR!

tgmof:
## Overview

Resolves #856

As (partly) explained in #856, the issue is that, taking any file:
- that is "unformatted enough" i.e. that it has several "chunks" output by sqlfmt
- that "deletes enough lines" (so that the already nicely-formatted text between 2 "chunks" should be moved too)

The file will be formatted wrongly to the extend that the SQL code is not valid anymore.

Example: See related issue for a file that would get corrupted before this PR.

Code walkthrough:

Option 1: Let sqlfmt do the job entirely instead of parsing its results (faster and simpler). https://github.com/AltimateAI/vscode-dbt-power-user/pull/930

Option 2 (this PR): Parse sqlfmt results and create 'TextEdit' changes for each row to change and each row in-between chunks:
- I added a "replace" function since this delete/insert method was already used once and I needed the same logic again.
- I save the "lastChunk" which is the previous chunk before the current one being processed. 
Example of 2 chunks output by sqlfmt (taking the example described in the issue) that are delimited by @@:
While I am in the 2nd chunk (@@ -32,15 +23,9 @@) I still have access to the previous chunk (@@ -13,17 +13,8 @@).
```diff
$ sqlfmt -l 120 --diff debug_project/models/example/my_first_dbt_model.sql      
1 file failed formatting check.
0 files passed formatting check.
debug_project/models/example/my_first_dbt_model.sql failed formatting check.
--- source_query
+++ formatted_query
@@ -13,17 +13,8 @@
 
         select 1 as id
 
-
-
-
-
-
-
-
-
-
-
-        union all select null as id
+        union all
+        select null as id
 
     )
 
@@ -32,15 +23,9 @@
 
 union all
 select *
-from 
+from
 
-
-
-
-
-
-
-source_data
+    source_data
 
     /*
     Uncomment the line below to remove records with null `id` values

```

- In the diff above, you can see that the lines in-between the 2 chunks don't appear in the diff (because they are already correctly formatted - line 30 to 32 in the original file).
- These lines that don't appear in the diff need to be moved from their original location (line 30 to 32) to their new location (just after the end of chunk 1, line 13+8 (as indicated by the @@ delimiter) = 21 so it should be moved to lines 21 to 23)

## Checklist

=> No it's a bugfix and thus it was an expected working feature, so no documentation change needed.

### DIFF
--- a/src/document_formatting_edit_provider/dbtDocumentFormattingEditProvider.ts
+++ b/src/document_formatting_edit_provider/dbtDocumentFormattingEditProvider.ts
@@ -110,7 +110,26 @@ export class DbtDocumentFormattingEditProvider
     const textEdits: TextEdit[] = [];
     const diffs = parseDiff(diffOutput);
     diffs.forEach((diff) => {
+      let lastChunk: parseDiff.Chunk;
       diff.chunks.forEach((chunk) => {
+        if (lastChunk) {
+          // Move the lines in-between chunks to their new positions
+          // (The lines after the last chunk don't need to be handled)
+          for (
+            let index = lastChunk.oldStart + lastChunk.oldLines, lineNb = 0;
+            index < chunk.oldStart;
+            index++, lineNb++
+          ) {
+            textEdits.push(
+              ...this.replace(
+                document,
+                index - 1,
+                lastChunk.newStart + lastChunk.newLines - 2 + lineNb,
+                document.lineAt(index - 1).text + "\n",
+              ),
+            );
+          }
+        }
         // Ensure lines added are not out of bounds of chunk
         const oldBoundChunk = chunk.oldLines + chunk.oldStart - 1;
         chunk.changes.forEach((change) => {
@@ -124,18 +143,11 @@ export class DbtDocumentFormattingEditProvider
             );
           }
           if (this.isNormalChange(change)) {
-            // Reflect "replace" edits as delete & insert
-            // First, delete line
             textEdits.push(
-              TextEdit.delete(
-                document.lineAt(change.ln1 - 1).rangeIncludingLineBreak,
-              ),
-            );
-            // Add line
-            textEdits.push(
-              TextEdit.insert(
-                document.lineAt(Math.min(change.ln2, oldBoundChunk) - 1).range
-                  .start,
+              ...this.replace(
+                document,
+                change.ln1 - 1,
+                Math.min(change.ln2, oldBoundChunk) - 1,
                 change.content.slice(1) + "\n",
               ),
             );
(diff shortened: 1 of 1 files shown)


anandgupta42:
Thanks for the contribution @tgmof. We will review it soon.

tgmof:
Still thinking about it and wondering if it wouldn't be simpler to delegate the responsibility to sqlfmt entirely (running it against the file path instead of the string content read from the file) since their program already does all the parsing work (so they already probably efficiently replace only what's needed, saving us to maintain this logic).

saravmajestic (commented):
Code changes look good for me, but I would let @mdesmet to approve if the change works as expected

tgmof:
> Code changes look good for me, but I would let @mdesmet to approve if the change works as expected

@saravmajestic could you also review this PR? https://github.com/AltimateAI/vscode-dbt-power-user/pull/919#pullrequestreview-1903177776
It's another solution solving the same problem but I think it's a better solution for the reason explained in the PR description. Any thoughts on it?

mdesmet:
@tgmof : I think I prefer the other approach but we need some way to validate the command has run succesfully or not.

mdesmet (approved):

mdesmet:
I will merge this as this is ready to be released, we can always discuss the other PR at a later stage.

tconbeer:
TIL y'all were parsing the output from `--diff`; I'm pretty sure dbt cloud and the other implementation I know of just replace the contents of the whole file.

PR #922, opened 2024-02-20
feat: lineage fixes

AdiGajbhiye:
## Overview
- propagation of edge type throughout column lineage
- during cll, 1st level not exanding
- added settings sidebar
Screenshot 2024-02-21 at 5 35 16 PM

## Checklist


suryaiyer95 (commented):
Tested

[x] Expanding column lineage on edge nodes and it auto expands the level
[x] Non select edges of children/ parents get propogated

new_lineage_panel/src/App.tsx
```diff
@@ -92,9 +92,13 @@ export const LineageContext = createContext<{
   setMinRange: Dispatch<SetStateAction<[number, number]>>;
   nodeCount: number;
   setNodeCount: Dispatch<SetStateAction<number>>;
+  selectCheck: boolean;
```
saravmajestic, line 95: Food for thought: you can create a slice like this https://github.com/AltimateAI/vscode-dbt-power-user/blob/master/webview_panels/src/modules/app/appSlice.ts#L8 and use it in your context/provider 
https://github.com/AltimateAI/vscode-dbt-power-user/blob/master/webview_panels/src/modules/app/AppProvider.tsx#L17
Then you can dispatch any actions in slice and get state values easily in any component https://github.com/AltimateAI/vscode-dbt-power-user/blob/master/webview_panels/src/modules/dataPilot/components/queryAnalysis/provider/useQueryAnalysisContext.ts#L18. This will help to avoid setting all these getters and setters in context everytime

new_lineage_panel/src/App.tsx
```diff
@@ -173,13 +182,13 @@ function App() {
...
       if (existingNode) {
         setSelectedTable(node.table);
-        const [nodes, edges] = highlightTableConnections(
-          _flow.getNodes(),
-          _flow.getEdges(),
-          node.table
-        );
-        _flow.setNodes(nodes);
-        _flow.setEdges(edges);
+        let nodes = _flow.getNodes();
```
saravmajestic, line 185: Better to avoid mutable variables and extract this code to separate function which could return nodes and edges

new_lineage_panel/src/graph.ts
```diff
@@ -362,6 +364,39 @@ const processColumnLineage = async (
 
   const seeMoreLineage: ColumnLineage[] = [];
 
+  // since many edges can come to same node, one node can have multiple direct/indirect edges
+  // 1st pass to collect all type of edges that column at current level can have
+  const columnEdgeTypeCandidates: Record<string, string[]> = {};
```
saravmajestic, line 369: are there any e2e tests for checking these cases?

saravmajestic (approved):
Just some thoughts on improvement

new_lineage_panel/src/App.tsx
AdiGajbhiye: Rest of the code is not design for this. Will redesign it later in future PR.

new_lineage_panel/src/graph.ts
AdiGajbhiye: currently no.

PR #924, opened 2024-02-21
new interface for adding external executors

anandgupta42:
## Checklist


package.json
```diff
@@ -160,6 +160,15 @@
...
             "description": "Add additional params to `sqlfmt`"
+          },
+          "dbt.dbtIntegration": {
+            "type": "string",
+            "enum": [
+              "core",
+              "cloud"
+            ],
+            "default": "core",
+            "description": "Use dbt core or cloud"
```
anandgupta42, line 171: remove this from public setting for now

anandgupta42: make it a readonly property

src/dbt_client/dbtIntegration.ts
```diff
@@ -305,52 +346,99 @@ export class DBTCommandExecutionInfrastructure {
...
+          },
+          async (_, token) => {
+            await commandExecution(token);
+          },
+        );
+      } else {
+        await commandExecution();
+      }
+      this.queueStates.set(queueName, false);
+      this.pickCommandToRun(queueName);
```
AdiGajbhiye, line 404: not sure about this. This might hit max recursion depth limit. Maybe better to use while loop.

AdiGajbhiye: i guess should be fine if queues are small.

src/webview_provider/queryResultPanel.ts
```diff
@@ -371,14 +371,9 @@ export class QueryResultPanel implements WebviewViewProvider {
         );
         return;
       }
-      window.showErrorMessage(
-        extendErrorWithSupportLinks(
-          "Encountered an unknown issue: " + exc.message + ".",
-        ),
-      );
```
AdiGajbhiye, line 378: any particular reason to remove this??

src/dbt_client/dbtIntegration.ts
mdesmet: This is not really new code. we have been using it in core for ages. being able to support more than one queue is new.

src/webview_provider/queryResultPanel.ts
mdesmet: The thing is that dbt cloud doesn't have PythonExceptions, this code path was never really supposed to hit before but now it is every dbt cloud error.

src/commandProcessExecution.ts
```diff
@@ -49,11 +49,13 @@ export class CommandProcessExecution {
       cwd: this.cwd,
       env: this.envVars,
     });
-    if (this.token !== undefined) {
-      this.disposables.push(
-        this.token.onCancellationRequested(() => {
-          proc.kill("SIGINT");
-        }),
+    if (this.tokens !== undefined) {
```
saravmajestic, line 52:
```suggestion
    if (this.tokens?.length) {
```

saravmajestic: this will check both falsy and is array condition

src/dbt_client/dbtIntegration.ts
```diff
@@ -39,27 +40,34 @@ export class CLIDBTCommandExecutionStrategy
...
+    if (command.logToTerminal) {
+      this.terminal.log(
+        `> Executing task: ${command.getCommandAsString()}\n\r`,
+      );
+    }
     this.telemetry.sendTelemetryEvent("dbtCommand", {
       command: command.getCommandAsString(),
     });
-    if (command.focus) {
+    if (command.logToTerminal && command.focus) {
```
saravmajestic, line 70: shouldn't be this as first line of this block? so the `executing task` log above will show show?

src/dbt_client/dbtIntegration.ts
```diff
@@ -264,12 +301,16 @@ export interface DBTProjectIntegration extends Disposable {
...
 
 @provide(DBTCommandExecutionInfrastructure)
 export class DBTCommandExecutionInfrastructure {
-  private queue: DBTCommandExecution[] = [];
-  private running = false;
+  private queues: Map<string, DBTCommandExecution[]> = new Map<
+    string,
+    DBTCommandExecution[]
+  >();
+  private queueStates: Map<string, boolean> = new Map<string, boolean>();
```
saravmajestic, line 313: it will be useful if the value for each key is json object with field like status? If we want to extend this map for handling other values related to execution, it will be easy to add more fields.

src/dbt_client/dbtIntegration.ts
```diff
@@ -305,52 +346,99 @@ export class DBTCommandExecutionInfrastructure {
...
+    const queue = this.queues.get(queueName)!;
+    const running = this.queueStates.get(queueName);
+    if (!running && queue.length > 0) {
+      this.queueStates.set(queueName, true);
+      const { command, statusMessage, focus, showProgress } = queue.shift()!;
+      const commandExecution = async (token?: CancellationToken) => {
+        try {
+          await command(token);
+        } catch (error) {
+          window.showErrorMessage(
```
saravmajestic, line 376: better to use logging service?

src/dbt_client/dbtIntegration.ts
```diff
@@ -305,52 +346,99 @@ export class DBTCommandExecutionInfrastructure {
...
+          ? ProgressLocation.Notification
+          : ProgressLocation.Window,
+        cancellable: true,
+        title: commandExecution.statusMessage,
+      },
+      async (_, token) => {
+        try {
+          return await commandExecution.command(token);
+        } catch (error) {
+          window.showErrorMessage(
```
saravmajestic, line 428: better to use logging service?

src/dbt_client/index.ts
```diff
@@ -25,11 +27,26 @@ export class DBTClient implements Disposable {
...
   private shownError = false;
+  private dbtDetection: DBTDetection;
 
   constructor(
     private pythonEnvironment: PythonEnvironment,
     private dbtCoreDetection: DBTCoreDetection,
-  ) {}
+    private dbtCloudDetection: DBTCloudDetection,
+  ) {
+    const dbtIntegrationMode = workspace
```
saravmajestic, line 37: should we think about multi project setting as well? If we have to support multi project, then this should go under resource level config

PR #925, opened 2024-02-22
fix: handle errors for query execution appropriately

suryaiyer95:
## Overview

If project is not set up correctly due to misconfigurations in dbt_project.yml or 2 models with same name, etc. Executing query results in constant spinning. 

## Checklist


saravmajestic (approved):

anandgupta42 (approved):
LGTM

PR #927, opened 2024-02-22
Fix terminal integrated env

mdesmet:
## Overview

This fix ensures invalid config are not used by the extension resulting in inconsistent behaviour in the terminal vs the extension.

Can be simulated by putting this in .vscode/settings.json

```
  "terminal.integrated.env.tst": {
    "DBT_PROFILES_DIR": "~/.dbt/profiles.yml"
  } 

```
## Checklist

### DIFF
--- a/src/manifest/pythonEnvironment.ts
+++ b/src/manifest/pythonEnvironment.ts
@@ -129,11 +129,28 @@ export class PythonEnvironment implements Disposable {
           const env = config.terminal.integrated.env;
           // parse vs code environment variables
           for (const prop in env) {
+            // Ignore any settings not supported by the terminal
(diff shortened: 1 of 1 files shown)


src/manifest/pythonEnvironment.ts
```diff
@@ -127,6 +127,11 @@ export class PythonEnvironment implements Disposable {
           const env = config.terminal.integrated.env;
           // parse vs code environment variables
           for (const prop in env) {
+            // Ignore any settings not supported by the terminal
+            // We don't know which os is used in terminal unfortunately, so we just merge all of them.
+            if (!["osx", "windows", "linux"].includes(prop)) {
+              continue;
```
saravmajestic, line 133: adding a log that we have ignored this env variable will be useful

PR #930, opened 2024-02-25
fix: wrong formatting (#856)

### LINKED ISSUE #856: One specific BQ-DBT SQL file gets wrongly formatted and becomes unusable

tgmof:
## Overview

For explanations and context, read: https://github.com/AltimateAI/vscode-dbt-power-user/pull/919 This PR is what is called "Option 1" in that other PR.

Resolves #856 by letting sqlfmt do the job entirely instead of parsing its results:
- It seems faster to me (but to be fair, I didn't time it, it's just an impression).
- It removes the complexity of parsing the diff and moving code "in-between chunks" to their new location (see other PR if you want to know what I mean by "in-between chunks").
- It will make it easier if any issue is ever found in sqlfmt because it's clear that vscode-dbt-power-user doesn't intervene in the formatting process.

## Checklist


src/document_formatting_edit_provider/dbtDocumentFormattingEditProvider.ts
```diff
@@ -69,27 +67,10 @@ export class DbtDocumentFormattingEditProvider
...
-          window.showErrorMessage(
-            extendErrorWithSupportLinks(
-              "Could not process difference output from sqlfmt. Detailed error: " +
-                error +
-                ".",
-            ),
-          );
-        }
-      }
+      } catch (sqlfmtOutput) {}
```
mdesmet, line 73: How do we verify success here? We are ignoring the output?

tgmof: Indeed, the stdout is always something like "1 file changed, 0 file left unchanged". Parsing the "expected successful message" doesn't sound like a great idea because I could imagine that they might feel more free to change it than the diff or some error code. I guess we could simply use the return error code 0 from the shell but I wasn't sure how to do that because the try/catch looks to only capture the stdout/stderr. I can try to have a look this weekend to understand how to get the shell output code and use that for validation and keep the telemetry call.

tgmof: Fixed! Let me know what you think. I don't understand why it looks like sqlfmt systematically write to stderr, but anyway the shell error code looks more reliable to detect errors.

mdesmet:
@tgmof : I think there is an issue with this approach as you can format the document without saving it.

tgmof:
@mdesmet sorry I didn't understand your last comment and forgot about it after that. I think "format without autosave" is actually the normal behavior. Only running the tool directly doesn't ask before saving. But anyway, this also works without this PR, so it's totally fine. It's probably faster and simpler with this PR but not a life-changing thing worth investing too much time. Thanks for having looked at it anyway and thanks for all the amazing work in DBT power user!

PR #931, opened 2024-02-26
fix: limit needs to be an argument

suryaiyer95:
## Overview
The show command needs to have limit set as the argument. Otherwise it always defaults to 5. If undefined or negative, I am not  passing the parameter. So, it will default to 5

## Checklist

### DIFF
--- a/src/dbt_client/dbtCloudIntegration.ts
+++ b/src/dbt_client/dbtCloudIntegration.ts
@@ -144,21 +144,18 @@ export class DBTCloudProjectIntegration
   }
 
   async executeSQL(query: string, limit: number): Promise<QueryExecution> {
-    const commandArgs = [
-      "show",
-      "--inline",
-      query,
-      "--output",
-      "json",
-      "--log-format",
-      "json",
-    ];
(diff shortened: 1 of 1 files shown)


AdiGajbhiye (approved):
lgtm

src/dbt_client/dbtCloudIntegration.ts
```diff
@@ -143,17 +143,22 @@ export class DBTCloudProjectIntegration
...
-      new DBTCommand("Running sql...", [
-        "show",
-        "--inline",
-        query,
-        "--output",
-        "json",
-        "--log-format",
-        "json",
-      ]),
+      new DBTCommand("Running sql...", commandArgs),
```
mdesmet, line 161: dbtProject already has

```
if (limit <= 0) {
      window.showErrorMessage("Please enter a positive number for query limit");
      return;
    }
```

You can just add the limit to the parameter, no need for additional validation.

suryaiyer95: Makes sense. Thanks!

mdesmet (approved):

PR #932, opened 2024-02-26
adding debugging/troubleshooting logs

anandgupta42:
## Overview
- Adding debugging/troubleshooting logs for the extension

## Checklist

- [X ] I have run this code and it appears to resolve the stated issue


mdesmet:
FYI Now with the external executor, the CustomUnknownException  might not be anymore valid as the main/potentially failing logic is not always executed in Python. Before it was usually a right assumption if there is an exception that is not a PythonException it is a bug in the code and not just an environment issue. Please take that into account. Maybe we need tor rename that class to CustomJavascriptException so that there is no confusion about it.

src/dbt_client/index.ts
```diff
@@ -112,12 +113,18 @@ export class DBTClient implements Disposable {
...
+          message = "Please ensure dbt cloud cli is installed.";
+          window.showErrorMessage("Please ensure dbt cloud cli is installed.");
+        } else {
+          const answer = await window.showErrorMessage(
+            "Please ensure dbt is installed.",
+            DbtInstallationPromptAnswer.INSTALL,
+          );
+          if (answer === DbtInstallationPromptAnswer.INSTALL) {
+            commands.executeCommand("dbtPowerUser.installDbt");
+          }
```
mdesmet, line 127: I think we can comment this for now as dbt cannot yet be installed. Add a TODO

src/manifest/dbtProject.ts
```diff
@@ -625,7 +629,7 @@ select * from renamed
       window.showErrorMessage("Please enter a positive number for query limit");
       return;
     }
-    this.telemetry.sendTelemetryEvent("executeSQL", {
+    this.terminal.info("executeSQL", query, true, {
```
mdesmet, line 632: Change to debug

src/webview_provider/altimateWebviewProvider.ts
```diff
@@ -200,7 +204,11 @@ export class AltimateWebviewProvider implements WebviewViewProvider {
           break;
       }
     } catch (err) {
-      this.dbtTerminal.debug("error while handling command", err);
+      this.dbtTerminal.debug(
```
mdesmet, line 207: Shouldn't this be error?

src/dbt_client/index.ts
AdiGajbhiye: currently in `master`, it shows `Install dbt` button in both cases. This shows the button only for core

mdesmet (approved):

PR #936, opened 2024-02-27
feat: gaps for external executor

AdiGajbhiye:
## Overview

- check connectivity with backend before running operation
- add telemetry events for external executor

## Checklist

### DIFF
--- a/src/dbt_client/dbtCloudIntegration.ts
+++ b/src/dbt_client/dbtCloudIntegration.ts
@@ -104,7 +104,10 @@ export class DBTCloudDetection implements DBTDetection {
 export class DBTCloudProjectDetection
   implements DBTProjectDetection, Disposable
 {
+  constructor(private altimate: AltimateRequest) {}
+
   async discoverProjects(projectDirectories: Uri[]): Promise<Uri[]> {
+    this.altimate.handlePreviewFeatures();
     const packagesInstallPaths = projectDirectories.map((projectDirectory) =>
       path.join(projectDirectory.fsPath, "dbt_packages"),
     );
@@ -255,7 +258,6 @@ export class DBTCloudProjectIntegration
       );
     }
     this.dbtPath = getDBTPath(this.pythonEnvironment, this.terminal);
-    this.altimate.handlePreviewFeatures();
   }
 
   getTargetPath(): string | undefined {


src/dbt_client/dbtCloudIntegration.ts
```diff
@@ -137,13 +138,36 @@ export class DBTCloudProjectIntegration
       this.rebuildManifestDiagnostics,
       this.pythonBridgeDiagnostics,
     );
+    this.checkConnectivity();
+  }
+
+  throwIfNotAuthenticated() {
```
mdesmet, line 144: make method private

src/dbt_client/dbtCloudIntegration.ts
```diff
@@ -137,13 +138,36 @@ export class DBTCloudProjectIntegration
...
+    if (!this.isAuthenticated) {
+      throw new Error("Unable to connect with altimate backend");
```
mdesmet, line 146: error message might be misleading.

src/dbt_client/dbtCloudIntegration.ts
```diff
@@ -395,6 +412,8 @@ export class DBTCloudProjectIntegration
   }
 
   async unsafeCompileQuery(query: string): Promise<string | undefined> {
+    this.telemetry.sendTelemetryEvent("dbtCloud:unsafeCompileQuery");
+    this.throwIfNotAuthenticated();
```
mdesmet, line 416: I think this should work differently. we should send an event in dbtProject and add a projectIntegration dimension. that way we can easily compare events for both cloud vs core.

src/validation_provider/index.ts
```diff
@@ -20,28 +21,31 @@ export class ValidationProvider implements Disposable {
...
     );
   }
 
-  private async validateCredentials() {
-    const config = this.altimate.getConfig();
-    if (!config) {
-      return;
-    }
-    const { key, instance } = config;
+  async validateCredentials(silent: boolean) {
```
mdesmet, line 32: Create method alias for silent mode, booleans are difficult to understand in the calling code.

src/validation_provider/index.ts
```diff
@@ -20,28 +21,31 @@ export class ValidationProvider implements Disposable {
           if (!e.affectsConfiguration("dbt")) {
             return;
           }
-          this.validateCredentials();
+          this.validateCredentials(false);
```
mdesmet, line 24: it will work differently from before as it was showing an error message. Is that intentional?

AdiGajbhiye: not sure what you mean. i am still getting error notification if credentials are wrong.

src/validation_provider/index.ts
```diff
@@ -28,20 +29,31 @@ export class ValidationProvider implements Disposable {
...
-    const { key, instance } = config;
+  validateCredentials() {
+    this._validateCredentials(false);
+  }
+
+  validateCredentialsSilently() {
+    this._validateCredentials(true);
+  }
+
+  private async _validateCredentials(silent: boolean) {
```
mdesmet, line 40:
```suggestion
  private async validateCredentials(silent: boolean = false) {
```

src/validation_provider/index.ts
```diff
@@ -9,6 +9,7 @@ export class ValidationProvider implements Disposable {
   private disposables: Disposable[] = [];
   private currInstanceName: string | undefined;
   private currAPIKey: string | undefined;
+  isAuthenticated = false;
```
mdesmet, line 12: maybe it is better to provide a method for this and make this private?

mdesmet (approved):

PR #939, opened 2024-02-28
fix: command process execute to get compete output from stdout and stderr

AdiGajbhiye:
## Checklist


src/commandProcessExecution.ts
```diff
@@ -32,6 +32,11 @@ export class CommandProcessExecutionFactory {
   }
 }
 
+export interface CommandProcessResult {
+  stdout: string;
```
mdesmet, line 36: Please also add a field that contains the full output in the order that it is received.

PR #951, opened 2024-03-01
feat: Custom packages fallback if python fails

svipulrao98:
## Overview
Read custom packages if python fails to do so

## Checklist

### DIFF
--- a/src/dbt_client/dbtCoreIntegration.ts
+++ b/src/dbt_client/dbtCoreIntegration.ts
@@ -139,13 +139,12 @@ export class DBTCoreProjectDetection
         "An error occured while finding package paths: " + error,
       );
       // Fallback to reading yaml files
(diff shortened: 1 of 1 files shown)


src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -118,6 +138,10 @@ export class DBTCoreProjectDetection
         "dbtCoreIntegration:discoverProjects",
         "An error occured while finding package paths: " + error,
       );
+      // Fallback to reading yaml files
+      packagesInstallPaths = projectDirectories.map((projectDirectory) =>
+        this.getPackageInstallPathFallback(projectDirectory),
+      );
```
mdesmet, line 144: Only replace if fallback returns a value

PR #955, opened 2024-03-05
feat: generate dbt tests using datapilot

saravmajestic:
### Requirement
Allow user to use datapilot to generate dbt tests for model/columns using NLP

### Solution
- Added `Custom test` button along with dbt generic tests in add test flow
- Show datapilot panel when the above button is clicked with option to let user enter requirement
- Datapilot will generate sql based custom tests or macros from dbt_utils/expectations if available in dbt project
- If it is yaml suggestion, allow user to insert to schema directly
- if it is sql based, user will manually create sql files 

### Screenshot/Demo
image

### How to test
- Click Custom test button in add test button flow for model/column
- In the datapilot panel, enter your prompt. Ex: `make order_id not to be null`
- Datapilot will suggest `not_null` dbt test for the column
- Click Insert button
- This should add not_null test to the order_id column and show save button
- regression - query analysis
    - check skip followup for query change 
- regression - doc generator

## Checklist


src/services/dbtTestService.ts
```diff
@@ -168,4 +175,121 @@ export class DbtTestService {
...
+          columns: documentation.columns.map((c) => ({
+            column_name: c.name,
+            data_type: c.type,
+          })),
+        },
+      },
+    });
+  }
+
+  public async getTestsForCurrentModel() {
```
saravmajestic, line 226: moved from src/services/docGenService.ts

src/webview_provider/datapilotPanel.ts
```diff
@@ -107,74 +108,79 @@ export class DataPilotPanel extends AltimateWebviewProvider {
...
-            syncRequestId,
-          });
-        } catch (err) {
-          this.sendResponseToWebview({
-            command: "response",
-            syncRequestId,
-            error: (err as Error).message,
-          });
-        }
+      case "queryAnalysis:explain":
```
saravmajestic, line 120: used `handleSyncRequestFromWebview` instead of using try catch implementation for each command

src/webview_provider/newDocsGenPanel.ts
```diff
@@ -130,49 +126,37 @@ export class NewDocsGenPanel
 
     switch (command) {
       case "getTestCode":
-        this.sendResponseToWebview({
-          command: "response",
-          data: this.getDbtTestCode(
-            args.test as TestMetaData,
-            args.model as string,
-          ),
+        this.handleSyncRequestFromWebview(
```
saravmajestic, line 129: used `handleSyncRequestFromWebview` instead of repeated try catch blocks

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -909,6 +909,33 @@ export class DBTCoreProjectIntegration
     }
   }
 
+  findPackageVersion(packageName: string) {
+    if (!this.packagesInstallPath || !packageName) {
+      return;
```
mdesmet, line 914: Is this not an error?

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -909,6 +909,33 @@ export class DBTCoreProjectIntegration
...
+    }
+
+    const dbtProjectYmlFilePath = path.join(
+      this.packagesInstallPath,
+      packageName,
+      "dbt_project.yml",
+    );
+    if (!existsSync(dbtProjectYmlFilePath)) {
+      return;
```
mdesmet, line 923: Is this not an error? this means package is not installed.

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -909,6 +909,33 @@ export class DBTCoreProjectIntegration
...
+    }
+    const fileContents = readFileSync(dbtProjectYmlFilePath, {
+      encoding: "utf-8",
+    });
+    if (!fileContents) {
+      return;
```
mdesmet, line 929: Error?

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -909,6 +909,33 @@ export class DBTCoreProjectIntegration
...
+    }
+    const parsedConfig = parse(fileContents, {
+      strict: false,
+      uniqueKeys: false,
+      maxAliasCount: -1,
+    });
+    return parsedConfig?.version;
```
mdesmet, line 936: if version is not defined is also an error.

src/dbt_client/dbtIntegration.ts
```diff
@@ -346,6 +346,7 @@ export interface DBTProjectIntegration extends Disposable {
   getCatalog(): Promise<Catalog>;
   getDebounceForRebuildManifest(): number;
   getBulkSchema(nodes: DBTNode[]): Promise<Record<string, DBColumn[]>>;
+  findPackageVersion(packageName: string): string | undefined;
```
mdesmet, line 349:
```suggestion
  findPackageVersion(packageName: string): string;
```
What will you do when it doesn't return string?

webview_panels/src/modules/dataPilot/DataPilotProvider.tsx
```diff
@@ -41,46 +32,21 @@ const DataPilotProvider = ({
...
-      requestType: RequestTypes.QUERY_ANALYSIS,
-      state: RequestState.UNINITIALIZED,
-      query: request.query,
-      fileName: request.fileName,
-      //If analysis type is undefined, dont trigger api call
-      analysisType: triggerOnLoad ? analysisType : undefined,
-    } as DatapilotQueryAnalysisChat;
+  const handleIncomingDatapilotMessage = (request: Partial<DataPilotChat>) => {
+    panelLogger.info("datapilot incoming message", request);
+    const id = request.id ?? crypto.randomUUID();
```
mdesmet, line 37: How can request.id be empty? is that not a bug?

mdesmet (commented):
I have slight concerns on how much we go to check the active text editor. I think this will be a headache to fix if we add unsaved changes handlers. Can't we refactor this to only retrieve the project once and pass as reference to the messages?

src/dbt_client/dbtCoreIntegration.ts
saravmajestic: This is not an error. If the file does not exist, then user does not have this package installed and we dont have to use it.

src/altimate.ts
```diff
@@ -113,6 +113,13 @@ export interface QueryAnalysisRequest {
   history?: QueryAnalysisChat[];
 }
 
+export interface CreateDbtTestRequest {
+  session_id: string;
+  model: DocsGenerateModelRequestV2["dbt_model"];
```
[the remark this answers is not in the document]
saravmajestic: fixed

src/dbt_client/dbtCoreIntegration.ts
saravmajestic: fixed

src/dbt_client/dbtCoreIntegration.ts
saravmajestic: fixed

src/dbt_client/dbtCoreIntegration.ts
saravmajestic: fixed

src/dbt_client/dbtIntegration.ts
saravmajestic: Current requirement does not expect the version to be always available. If version is not returned, we can assume that the package is not installed and wont use it. I have updated it to return string or throw error, so the caller will handle this as per the need

webview_panels/src/modules/dataPilot/DataPilotProvider.tsx
saravmajestic: In case of doc generation, id will be passed from doc editor panel to webview provider and then to datapilot panel. In other cases, id is optional and will be created/used only in datapilot. So this is not a required field

saravmajestic:
> I have slight concerns on how much we go to check the active text editor. I think this will be a headache to fix if we add unsaved changes handlers. Can't we refactor this to only retrieve the project once and pass as reference to the messages?

@mdesmet i have fixed the active editor changes for the dbt test creation flow and updated other places where current active file is enough.

src/dbt_client/dbtCloudIntegration.ts
```diff
@@ -832,6 +832,10 @@ export class DBTCloudProjectIntegration
     }
   }
 
+  findPackageVersion(packageName: string) {
+    return "";
```
mdesmet, line 836: I didn't think of dbt cloud, if on cloud we can never get the version and that is not an exception then the return value should indeed be string | undefined. Core integration should still throw an error if it cannot detect the version as this indicates a problem.

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -909,6 +909,40 @@ export class DBTCoreProjectIntegration
...
+    if (!fileContents) {
+      throw new Error(`${packageName} has empty dbt_project.yml`);
+    }
+    const parsedConfig = parse(fileContents, {
+      strict: false,
+      uniqueKeys: false,
+      maxAliasCount: -1,
+    });
+    if (!parsedConfig?.version) {
+      throw new Error(`Missing version in ${packageName}/dbt_project.yml`);
```
mdesmet, line 940: maybe use full path, otherwise we need to guess which project in a multi project setup.

src/webview_provider/newDocsGenPanel.ts
```diff
@@ -130,147 +126,118 @@ export class NewDocsGenPanel
...
-            .getProject()
-            ?.getColumnsOfSource(args.source as string, args.table as string);
-          this.sendResponseToWebview({
-            command: "response",
-            data: {
+        this.handleSyncRequestFromWebview(
+          syncRequestId,
+          async () => {
+            const columnsFromSources = await this.queryManifestService
+              .getProject()
```
mdesmet, line 179: get project from message?

src/webview_provider/newDocsGenPanel.ts
```diff
@@ -130,147 +126,118 @@ export class NewDocsGenPanel
...
-          const columns = await this.queryManifestService
-            .getProject()
-            ?.getColumnsOfModel(args.model as string);
-          this.sendResponseToWebview({
-            command: "response",
-            data: {
+        this.handleSyncRequestFromWebview(
+          syncRequestId,
+          async () => {
+            const columns = await this.queryManifestService
```
mdesmet, line 196: get project from message?

src/webview_provider/newDocsGenPanel.ts
```diff
@@ -130,147 +126,118 @@ export class NewDocsGenPanel
...
         break;
+
       case "getSourcesInProject":
-        const sources = this.queryManifestService.getSourcesInProject(
-          window.activeTextEditor?.document.uri,
-        );
+        this.handleSyncRequestFromWebview(
+          syncRequestId,
+          async () => {
+            const sources = this.queryManifestService.getSourcesInProject(
```
mdesmet, line 212: get project from message?

webview_panels/src/modules/dataPilot/DataPilotProvider.tsx
mdesmet: I think it would be better to have it always there in that case to keep things simple, but not blocking for this PR.

mdesmet (commented):
Please fix signature of dbtIntegration `findPackageVersion`, other stuff is not blocking.

src/webview_provider/newDocsGenPanel.ts
saravmajestic: all these commands are triggered based on the documentation of current model that is open in editor. do we still need to handle that from message?

PR #960, opened 2024-03-06
feat: fetch bulk schemas

suryaiyer95:
## Checklist

### DIFF
--- a/dbt_core_integration.py
+++ b/dbt_core_integration.py
@@ -33,6 +33,7 @@ from typing import (
     Tuple,
     TypeVar,
     Union,
+    Set,
 )
 
 import agate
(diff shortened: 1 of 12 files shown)


dbt_core_integration.py
```diff
@@ -136,6 +137,29 @@ def to_dict(obj):
     return obj
 
 
+def extract_db_columns(data, mapping):
+    tables = {}
+
+    # Iterate over each row in the data
+    for row in data['rows']:
```
suryaiyer95, line 144: TODO: Nicer way for this

mdesmet (commented):
Is this api compatible with all dbt versions?

suryaiyer95:
> Is this api compatible with all dbt versions?

Need to check. Also not important anymore because it uses the information schema and we can't use it

PR #946, opened 2024-02-29
improve: use watch while debugging extension

saravmajestic:
### Problem
During development (either webviews or extension), hot reload code changes is not working. Developer has to wait for ~10-15s to rebuild whole extension for every change

### Solution
- Ideal solution will be hot reloading code changes in debug extension host
- For now, I have created build in watch mode (webpack watch for extension and vite build watch for webviews)
- Modified the launch task to use the `debug-task` in `tasks.json` which will run all builds in watch mode
- Any code changes will trigger the rebuild and we need to reload the extension using `cmd+r` (mac) or use `Reload window` in command paletter in vscode

### How to test
- Go to Run and Debug view
- Click `Launch Extension`
- This will open extension host
- Make any change in extension or webview
- Reload the extension
- should see the changes

## Checklist

- [x] I have run this code and it appears to resolve the stated issue
- [ ] I have added the necessary tests
- [ ] `README.md` updated and added information about my change
- [ ] Documentation website is updated with information related to this change

PR #973, opened 2024-03-08
fix: debug launch config

saravmajestic:
### Problem
- recently added [debug](https://github.com/AltimateAI/vscode-dbt-power-user/issues/946) config for development is not working reliable. It stops watch command abruptly
- also, the debug source code is not reflecting the updated code
- icons from codicons are not showing up

### Solution
- update `launch` config to use `webpack` as before
- added new `debug extension` config to use this new flow.

### Screenshot/Demo
NA

### How to test
- Click `Launch extension` / `debug extension` 

## Checklist


mdesmet (approved):
I'm ok with the change but ideally we can fix the debug experience instead.

PR #974, opened 2024-03-08
feat: Query sharing s3

gaurpulkit:
- Modifying the existing Query Result Sharing PR by using a pre-signed s3 url to send the data directly to the user's s3 server instead of storing the data in Altimate's database.

### DIFF
--- a/src/webview_provider/queryResultPanel.ts
+++ b/src/webview_provider/queryResultPanel.ts
@@ -237,20 +237,10 @@ export class QueryResultPanel implements WebviewViewProvider {
                   window.showInformationMessage(
                     "Generated signed url. Uploading data...",
                   );
-                  const uploadResponse =
-                    await this.altimate.uploadDataToSignedUrl(
-                      result!.signed_url,
-                      message,
-                    );
-                  if (uploadResponse.status !== 200) {
-                    window.showErrorMessage(
-                      "Error verifying upload. Please try again.",
-                    );
-                    this.telemetry.sendTelemetryError(
-                      "ShareQueryResult",
-                      "Upload response status not 200",
-                    );
-                  }
+                  await this.altimate.uploadDataToSignedUrl(
+                    result!.signed_url,
+                    message,
+                  );
                   // verifying upload
                   const verifyResponse =
                     await this.altimate.verifyShareQueryUpload({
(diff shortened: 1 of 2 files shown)


mdesmet:
I think the point is to have both ways. We should offer the choice to the user to either use s3 bucket or use our saas.

@anandgupta42: Can you confirm?

src/webview_provider/queryResultPanel.ts
```diff
@@ -219,25 +219,53 @@ export class QueryResultPanel implements WebviewViewProvider {
             window.withProgress(
               {
                 location: ProgressLocation.Notification,
-                title: "Uploading data to Altimate backend",
+                title: "Generating Shareable URL",
```
saravmajestic, line 222: add telemetry

src/webview_provider/queryResultPanel.ts
```diff
@@ -219,25 +219,53 @@ export class QueryResultPanel implements WebviewViewProvider {
...
                     window.showInformationMessage(
-                      `URL is copied to clipboard: [${result.share_url}](${result.share_url})`,
+                      "Generated signed url. Uploading data...",
                     );
+                    const uploadResponse =
+                      await this.altimate.uploadDataToSignedUrl(
+                        result.signed_url,
+                        message,
+                      );
+                    if (uploadResponse.status === 200) {
```
saravmajestic, line 237: early returns

src/webview_provider/queryResultPanel.ts
```diff
@@ -219,25 +219,53 @@ export class QueryResultPanel implements WebviewViewProvider {
...
+                      const verifyResponse =
+                        await this.altimate.verifyShareQueryUpload({
+                          sharing_table_id: result.sharing_table_id,
+                        });
+
+                      const verifyData = verifyResponse as {
+                        ok: boolean;
+                        share_url: string;
+                      };
+                      if (verifyData.ok) {
```
saravmajestic, line 248: early returns

saravmajestic (changes requested):
some cleanup needed

src/webview_provider/queryResultPanel.ts
mdesmet: Also http related stuff shouldn't leak into implementations.

saravmajestic:
Not needed anymore

PR #980, opened 2024-03-11
fix: send defer event to backend only for remote

saravmajestic:
### Problem

For local mode in defer to prod feature, dont send event to backend to track the run

### Solution

Removed tracking event and added telemetry event

### Screenshot/Demo

NA

### How to test

- Enable defer to prod with local integration
- Run a dbt command like run,test,build
- Should not send event to backend but to telemetry

## Checklist

### DIFF
--- a/src/dbt_client/dbtCoreIntegration.ts
+++ b/src/dbt_client/dbtCoreIntegration.ts
@@ -620,7 +620,12 @@ export class DBTCoreProjectIntegration
         "localManifest",
         `local defer params: ${args.join(" ")}`,
       );
-      this.altimateRequest.sendDeferToProdEvent(ManifestPathType.LOCAL);
+      this.dbtTerminal.info(
+        "deferToProd",
+        "executing dbt command with defer params local mode",
+        true,
+        args,
+      );
       return args;
     }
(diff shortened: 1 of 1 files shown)


src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -651,6 +656,12 @@ export class DBTCoreProjectIntegration
           args.push("--favor-state");
         }
         this.altimateRequest.sendDeferToProdEvent(ManifestPathType.REMOTE);
+        this.dbtTerminal.info(
```
mdesmet, line 659: Should be debug or you pollute the terminal for normal commands.

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -620,7 +620,12 @@ export class DBTCoreProjectIntegration
         "localManifest",
         `local defer params: ${args.join(" ")}`,
       );
-      this.altimateRequest.sendDeferToProdEvent(ManifestPathType.LOCAL);
+      this.dbtTerminal.info(
```
mdesmet, line 623: See other comment

mdesmet (approved):

PR #998, opened 2024-03-15
Trigger manifest refresh on seed changes

mdesmet:
### Problem

Trigger manifest refresh on seed changes

Fix for #999 

### Solution

Include csv and seed directories in source file changed event

### Screenshot/Demo

N/A

### How to test

Add/Change/Remove seed files

## Checklist

### DIFF
--- a/src/dbt_client/dbtCoreIntegration.ts
+++ b/src/dbt_client/dbtCoreIntegration.ts
@@ -206,6 +206,7 @@ export class DBTCoreProjectIntegration
   private version?: number[];
   private packagesInstallPath?: string;
   private modelPaths?: string[];
+  private seedPaths?: string[];
   private macroPaths?: string[];
   private python: PythonBridge;
   private disposables: Disposable[] = [];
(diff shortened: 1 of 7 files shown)


src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -825,6 +831,19 @@ export class DBTCoreProjectIntegration
     return modelPaths;
   }
 
+  private async findSeedPaths(): Promise<string[]> {
+    let seedPaths = await this.python.lock(
+      (python) => python`to_dict(project.config.seed_paths)`,
+    );
+    seedPaths = seedPaths.map((seedPath: string) => {
```
saravmajestic, line 838: some logs could be useful here. also you can directly return result of map instead of mutating `seedPaths`

saravmajestic (commented):
better to have some logs

src/dbt_client/dbtCoreIntegration.ts
mdesmet: Agreed, I was thinking of that but it seems we already applied same code above. will change both cases.

saravmajestic (approved):

Issue #992, opened 2024-03-15
Enable setting key and instance name using an environment variable


anandgupta42:
### Describe the feature

One should be able to set altimate instance and altimate key using an environment variable in the setting .json 

Example: "dbt.altimateAiKey": "${env:ALTIMATE_TOKEN}"

### Who will benefit?

All the users who want to set the settings using an env variable

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

anandgupta42:
Fix released in pre-release version 0.37.9

PR #1002, opened 2024-03-15
fix: settings from env

AdiGajbhiye:
### Problem

Fix for issue #992 

### Solution

Added support for env var from shell and .env file

### Screenshot/Demo

### How to test

## Checklist


mdesmet (commented):
LGTM but seems not to support our env setup guidelines, I think it will ignore env vars setup in integrated.terminal.env settings or python .env file.

AdiGajbhiye:
integrated.terminal.env will be tricky. There can be multiple terminal and each terminal have different value for same env var. Also not sure about whether the value be retained after reload?

src/manifest/pythonEnvironment.ts
```diff
@@ -33,12 +41,20 @@ export class PythonEnvironment implements Disposable {
...
+        "dbtPythonPathOverride",
+        this.environmentVariables,
+      ) || this.executionDetails!.getPythonPath()
     );
   }
 
-  public get environmentVariables() {
-    return this.executionDetails!.getEnvVars();
+  public get environmentVariables(): EnvironmentVariables {
+    const envVarsDebug = this.executionDetails!.getEnvVars();
```
mdesmet, line 52: Why Debug? Seems weird name for something we use to get the env vars for production purposes.

src/manifest/pythonEnvironment.ts
```diff
@@ -118,50 +114,66 @@ export class PythonEnvironment implements Disposable {
...
+              "pythonEnvironment:envVars",
+              `workspacePath:${workspacePath.uri.fsPath}`,
+            );
+            const workspaceEnv =
+              api.environments.getEnvironmentVariables(workspacePath);
+            this.dbtTerminal.debug(
+              "pythonEnvironment:envVars",
+              `workspaceEnv:${Object.keys(workspaceEnv)}`,
+            );
+            setEnvVars(workspaceEnv, "dotenv");
```
mdesmet, line 176: Can we move this code out of the python extension piece. it is unrelated to the python extension interface.

src/altimate.ts
```diff
@@ -207,22 +207,22 @@ export class AltimateRequest {
...
     private dbtTerminal: DBTTerminal,
+    private pythonEnvironment: PythonEnvironment,
   ) {}
 
-  getConfig(): AltimateConfig | undefined {
-    const key = workspace.getConfiguration("dbt").get<string>("altimateAiKey");
-    const instance = workspace
-      .getConfiguration("dbt")
-      .get<string>("altimateInstanceName");
+  getInstanceName() {
```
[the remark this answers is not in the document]
AdiGajbhiye: I think since getting API key is not simple anymore as we are doing env var substitutions, we should have only one way getting them. And in some places we need them separately. WDYT

package.json
```diff
@@ -521,6 +521,12 @@
           "light": "./media/images/build-model_light.svg",
           "dark": "./media/images/build-model_dark.svg"
         }
+      },
+      {
+        "command": "dbtPowerUser.printEnvVars",
+        "title": "Print env vars",
```
mdesmet, line 527: Don't abbreviate.

```suggestion
        "title": "Print environment variables",
```

src/altimate.ts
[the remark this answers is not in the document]
mdesmet: I would think they only make sense if they are together and if you have not one of them you don't have a config.

src/manifest/pythonEnvironment.ts
```diff
@@ -15,7 +16,7 @@ interface PythonExecutionDetails {
 export class PythonEnvironment implements Disposable {
   private executionDetails?: PythonExecutionDetails;
   private disposables: Disposable[] = [];
-
+  private envFrom: Record<string, EnvFrom> = {};
```
mdesmet, line 19:
```suggestion
  private environmentVariableSource: Record<string, EnvFrom> = {};
```

src/manifest/pythonEnvironment.ts
```diff
@@ -53,26 +66,63 @@ export class PythonEnvironment implements Disposable {
...
+    for (const key in currEnvVars) {
+      newEnvVars[key] = this.substituteSettingsVariables(
+        currEnvVars[key],
+        parentEnvVars,
+      );
+    }
+    return newEnvVars;
   };
 
+  substituteSettingsVariables(
```
mdesmet, line 89: This method should be private

src/manifest/pythonEnvironment.ts
```diff
@@ -53,26 +66,63 @@ export class PythonEnvironment implements Disposable {
...
+    value: any,
+    vsCodeEnv: EnvironmentVariables,
```
mdesmet, line 91: This should not be a parameter.

src/manifest/pythonEnvironment.ts
```diff
@@ -53,26 +66,63 @@ export class PythonEnvironment implements Disposable {
...
+    const value = workspace.getConfiguration("dbt").get<string>(key, "");
+    return this.substituteSettingsVariables(value, this.environmentVariables);
   }
 
-  private parseEnvVarsFromUserSettings = (vsCodeEnv: {
-    [k: string]: string;
-  }) => {
+  private parseEnvVarsFromUserSettings = (
+    currEnvVars: EnvironmentVariables,
+    parentEnvVars: EnvironmentVariables,
```
mdesmet, line 76: This seems very complicated for what we need to do. Why not simply only use `substituteSettingsVariables` on any new env var using the variable that are known at that time?

src/manifest/pythonEnvironment.ts
```diff
@@ -118,50 +153,61 @@ export class PythonEnvironment implements Disposable {
...
+            );
+            const workspaceEnv =
+              api.environments.getEnvironmentVariables(workspacePath);
+            this.dbtTerminal.debug(
+              "pythonEnvironment:envVars",
+              `workspaceEnv:${Object.keys(workspaceEnv)}`,
+            );
+            for (const key in workspaceEnv) {
+              envVars[key] = workspaceEnv[key];
+              this.environmentVariableSource[key] = "dotenv";
```
mdesmet, line 209: A lot of env variables from process are actually reported as coming from dotenv.

mdesmet (commented):
LGTM except source.

PR #1009, opened 2024-03-18
fix: show test macro code in tests editor

saravmajestic:
### Problem
1. In the new tests section in documentation editor, if a model/column has manually created macro test with below sample code in `macros` directory, the sql code is not display in right side panel, when clicking the macro name.
```
{% test positive_value(model, column_name) %}
SELECT
    *
FROM
    {{ model }}
WHERE
    {{ column_name}} < 1
{% endtest %}
```
2. If a test for column/model has extra config like [this example ](https://docs.getdbt.com/reference/resource-configs/where#examples), show these extra config in right side panel

### Solution
1. Modified the code to add the test macros also from `macros` directory into `macroMetamap` and use the information to find the file path for generating tests data
2. Modified `getTestCode` command to get sql source and also extra configs from schema.yml

### Screenshot/Demo:
Before             |  After
:-------------------------:|:-------------------------:
image  | image

image image image

### How to test
- Add a test macro with above sample code in dbt project/macros directory
- Apply the test to a column in schema file like below
```
    - name: minimum_nights
        tests:
          - positive_value
```
- Open documentation editor and select tests
- Click `positive_value` test from the column
- Right side panel should show the sql code from macros directory
- Add extra configs for generic tests ex: 
```
- unique:
              where: "minimum_nights < 10"
```
- Add tests from external packages like dbt utils or expectations. Ex:
```
- dbt_expectations.expect_column_quantile_values_to_be_between:
              quantile: .99
              min_value: 50
              max_value: 500
```
- Right side panel for each test should show the extra config as in screenshots

## Checklist

### DIFF
--- a/src/manifest/parsers/macroParser.ts
+++ b/src/manifest/parsers/macroParser.ts
@@ -55,7 +55,12 @@ export class MacroParser {
 
           for (let index = 0; index < macroFileLines.length; index++) {
             const currentLine = macroFileLines[index];
-            if (currentLine.match(new RegExp(`macro\\s${name}\\(`))) {
+            if (
+              currentLine.match(new RegExp(`macro\\s${name}\\(`)) ||
+              currentLine.match(
+                new RegExp(`test\\s${name.replace("test_", "")}\\(`),
+              )
+            ) {
               macroMetaMap.set(macroName, {
                 path: fullPath,
                 line: index,
--- a/src/services/dbtTestsService.ts
+++ b/src/services/dbtTestsService.ts
@@ -1,13 +1,9 @@
-import { AltimateRequest } from "../altimate";
 import { provideSingleton } from "../utils";
-import { DocGenService } from "./docGenService";
-import { StreamingService } from "./streamingService";
-import { QueryManifestService } from "./queryManifestService";
-import { DBTTerminal } from "../dbt_client/dbtTerminal";
 import { MacroMetaMap } from "../domain";
 
 @provideSingleton(DbtTestService)
 export class DbtTestService {
+  // Find the file path of test macro
   public getMacroFilePath = (
     macros: [string],
     projectName: string,
@@ -17,10 +13,14 @@ export class DbtTestService {
     if (!testName) {
       return;
     }
+
+    // Find if current test depends on test macro in current project
     const macro = macros.find(
       (m) => m === `macro.${projectName}.test_${testName}`,
     );
+
     if (macro) {
+      // return the file path if it ends with sql
       const macroData = macroMetaMap.get(`test_${testName}`);
       return macroData?.path.endsWith(".sql") ? macroData?.path : undefined;
     }
--- a/src/services/dbtTestsService.ts
+++ b/src/services/dbtTestsService.ts
@@ -13,7 +13,7 @@ export class DbtTestService {
     private queryManifestService: QueryManifestService,
   ) {}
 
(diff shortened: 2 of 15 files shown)


src/manifest/parsers/macroParser.ts
```diff
@@ -55,7 +55,12 @@ export class MacroParser {
 
           for (let index = 0; index < macroFileLines.length; index++) {
             const currentLine = macroFileLines[index];
-            if (currentLine.match(new RegExp(`macro\\s${name}\\(`))) {
+            if (
+              currentLine.match(new RegExp(`macro\\s${name}\\(`)) ||
+              currentLine.match(
+                new RegExp(`test\\s${name.replace("test_", "")}\\(`),
```
saravmajestic, line 61: Currently, if test macro is created with below code, this is not added to `macroMetaMap`. Added this matching code to make sure these macros are also added to `macroMetaMap`
```
{% test positive_value(model, column_name) %}
SELECT
    *
FROM
    {{ model }}
WHERE
    {{ column_name}} < 1
{% endtest %}
```

src/webview_provider/newDocsGenPanel.ts
```diff
@@ -91,11 +91,12 @@ export class NewDocsGenPanel
       return;
     }
 
+    const project = this.queryManifestService.getProject()?.getProjectName();
```
mdesmet, line 94: rename to projectName

src/services/dbtTestsService.ts
```diff
@@ -0,0 +1,28 @@
...
+      (m) => m === `macro.${projectName}.test_${testName}`,
+    );
+
+    if (macro) {
+      // return the file path if it ends with sql
+      const macroData = macroMetaMap.get(`test_${testName}`);
+      return macroData?.path.endsWith(".sql") ? macroData?.path : undefined;
+    }
+  };
+}
```
mdesmet, line 28: Can we not add this code in the macroParser? This way people will still need to remember to inject this class every time we us `macro.path`

saravmajestic: Actually, macros in macroMetamap has the right path values, but not the tests in testsMetamap. We could move this to testParser, but we dont have macroMetamap available there. Is there a way to get macro data in testParser? @mdesmet

mdesmet (approved):

src/webview_provider/newDocsGenPanel.ts
```diff
@@ -91,100 +92,37 @@ export class NewDocsGenPanel
...
     const tests = await this.docGenService.getTestsForCurrentModel();
     this.sendResponseToWebview({
       command: "renderTests",
       tests,
-      project: this.queryManifestService.getProject()?.getProjectName(),
+      project: projectName,
     });
   }
 
-  private stringifyTest = (
```
saravmajestic, line 102: moved this to dbtTestService

webview_panels/src/modules/documentationEditor/components/tests/DisplayTestDetails.tsx
```diff
@@ -197,76 +195,6 @@ const DisplayTestDetails = ({ onClose, test, column }: Props): JSX.Element => {
     }
   };
 
-  const getDisplayContent = () => {
```
saravmajestic, line 200: moved to `TestDetails.tsx`

mdesmet (approved):

Issue #746, opened 2023-12-04
 when there is an error executing a query it doesn't show the compiled SQL


anandgupta42:
### Expected behavior

The compiled SQL should be made available

### Actual behavior

Compiled SQL is not displayed for the failed queries

### Steps To Reproduce

run a query that fails 

### Operating System

Mac

### dbt version

1.6

### dbt Adapter

snowflake

### dbt Power User version

0.27

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #1011, opened 2024-03-18
fix: show compiled sql when execute error

AdiGajbhiye:
### Problem

[Describe the problem you are solving. Mention the ticket/issue if applicable.](https://github.com/AltimateAI/vscode-dbt-power-user/issues/746)

### Solution

### Screenshot/Demo

### How to test

## Checklist


anandgupta42:
@AdiGajbhiye Will this work for the external executor also?

src/webview_provider/queryResultPanel.ts
```diff
@@ -336,6 +336,10 @@ export class QueryResultPanel implements WebviewViewProvider {
       const output = await queryExecution.executeQuery();
       await this.transmitDataWrapper(output, query);
     } catch (exc: any) {
+      const activeFile = window.activeTextEditor?.document.uri;
+      if (activeFile) {
+        this.dbtProjectContainer.showCompiledSQL(activeFile);
```
mdesmet, line 341: This method has nothing to with executed sql, it will also only work if the user has run `dbt compile/build/run -m <model>` before. it will refer to the compiled sql of that time.

mdesmet:
> @AdiGajbhiye Will this work for the external executor also?

The only way to do that is to decouple compilation from executing the query, but it will have noticeable performance impact.

anandgupta42:
We can skip the external executor part then for now

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -358,11 +359,15 @@ export class DBTCoreProjectIntegration
...
           }
+          return { ...result, compiled_stmt };
         } catch (err) {
-          console.error("error while executing querytemplate conversion", err);
+          const message = `Error while executing sql: ${compiledQuery}`;
+          this.dbtTerminal.error("dbtCore:executeSQL", message, err);
+          if (err instanceof PythonException) {
+            throw new ExecuteSQLError(err.exception.message, compiledQuery!);
+          }
+          throw new ExecuteSQLError((err as Error).message, compiledQuery!);
```
mdesmet, line 369: The implementation has changed. Before exception on query template was ignored. 

I think the exception handling should be split in two, one for actual exectuion and one for the query template replacement.

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -340,11 +341,11 @@ export class DBTCoreProjectIntegration
...
         const compiledQuery = await this.unsafeCompileQuery(limitQuery);
-        const result = await queryThread!.lock<ExecuteSQLResult>(
-          (python) => python`to_dict(project.execute_sql(${compiledQuery}))`,
-        );
-        let compiled_stmt = result.compiled_sql;
         try {
+          const result = await queryThread!.lock<ExecuteSQLResult>(
+            (python) => python`to_dict(project.execute_sql(${compiledQuery}))`,
+          );
+          let compiled_stmt = result.compiled_sql;
```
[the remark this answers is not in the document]
anandgupta42: @AdiGajbhiye Can you please address this comment?

src/dbt_client/dbtCoreIntegration.ts
anandgupta42: @AdiGajbhiye Can you please address this comment?

mdesmet: `const matches = queryRegex.exec(result.compiled_sql);`

If above line fails it will throw an exception, but we are not really interested in it as we can just use the `compiledStatement`. That's what I meant that the code changes you made are not the same as it used to be. Also the code that is tried can never throw a PythonExxception.

mdesmet (approved):

PR #1013, opened 2024-03-19
feat: query translation

saravmajestic:
### Problem
Allow user to translate models/sql from one sql dialect to others supported by dbt.

### Solution
- added new option `translate` in datapilot and also context menu -> datapilot
- Target dialect (same as project adapter type) will be preselected
- User will need to select source dialect and translate the sql
- Explanation will be provided based on the translated code

### Screenshot/Demo

https://www.loom.com/share/54f02bd488ed4e2badd880cf5f77b5b6?sid=7c20fd85-7a0b-4a38-86d2-79349457b5de

### How to test

- Open a new file/existing file with sql query
- right click the file -> datapilot-> select translate option
- select source sql dialect
- Click translate
- Should see translated sql and replace button should replace the contents of original file
- Also explanation for the translation should show up

## Checklist


src/webview_provider/altimateWebviewProvider.ts
```diff
@@ -188,6 +188,15 @@ export class AltimateWebviewProvider implements WebviewViewProvider {
 
     try {
       switch (command) {
+        case "getProjectAdapterType":
+          this.handleSyncRequestFromWebview(
+            syncRequestId,
+            () => {
+              return this.queryManifestService.getProject()?.getAdapterType();
```
mdesmet, line 195: Can you guarantee this is the same project as on load. I assume the getProject will check the selected text editor again, what if we want to implement a onSelectedTextEditor handler that deals with unsaved data. You would have to fix all these occurrences.

src/webview_provider/datapilotPanel.ts
```diff
@@ -166,6 +207,31 @@ export class DataPilotPanel extends AltimateWebviewProvider {
         );
         break;
 
+      case "file:replace-contents":
+        if (!window.activeTextEditor || !params.sql || !params.filePath) {
+          return;
+        }
+        this.dbtTerminal.debug(
+          "file:replace-contents",
```
mdesmet, line 215: Very cryptic name of debug message

src/services/queryAnalysisService.ts
```diff
@@ -46,6 +62,100 @@ export class QueryAnalysisService {
...
+    if (!this.altimateRequest.handlePreviewFeatures()) {
+      return;
+    }
+
+    const editor = await this.fileService.openFileByPath(params.filePath);
+
+    const sql = editor.document.getText();
+
+    if (!params.source) {
+      throw new Error("Invalid source dialect");
```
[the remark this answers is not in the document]
saravmajestic: Created a new error class `UserInputError` and used that to log as debug instead of error. Same for below

src/services/queryAnalysisService.ts
```diff
@@ -46,6 +62,100 @@ export class QueryAnalysisService {
...
+    }
+
+    if (!params.target) {
+      throw new Error("Invalid target dialect");
```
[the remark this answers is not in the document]
saravmajestic: fixed

src/services/queryAnalysisService.ts
```diff
@@ -46,6 +62,100 @@ export class QueryAnalysisService {
...
+    }
+
+    const dbtProject = this.queryManifestService.getProject();
+
+    if (!dbtProject) {
+      const error = new Error("Invalid dbt project");
```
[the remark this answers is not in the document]
saravmajestic: Query translation in datapilot will not show if no file is selected. So ideally this code will be executed only if the file is selected. Even after having the file open, if we cannot still find dbt project, we are throwing error here

src/webview_provider/altimateWebviewProvider.ts
saravmajestic: This command is executed when user selects query translate and land in datapilot panel. So user will have the file open in editor and we find the adapter type based on this data. So, I think this should be fine

src/webview_provider/datapilotPanel.ts
saravmajestic: renamed :)

src/services/queryAnalysisService.ts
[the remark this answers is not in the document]
mdesmet: But this will become an unhandled error? or do we handle the error?

src/webview_provider/altimateWebviewProvider.ts
mdesmet: Why not assemble all this information at the start? I think this is more complicated than necessary.

src/webview_provider/datapilotPanel.ts
mdesmet: It's still very generic. I really suggest to always use class name+methodname so we know where this error comes from.

src/services/queryAnalysisService.ts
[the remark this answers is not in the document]
saravmajestic: This is handled in https://github.com/AltimateAI/vscode-dbt-power-user/blob/master/src/webview_provider/altimateWebviewProvider.ts#L114 which is a wrapper around the command handler

src/webview_provider/altimateWebviewProvider.ts
saravmajestic: This will need changes in multiple places, i will address this in separate PR. Created a ticket here: https://altimateai.atlassian.net/browse/AI-1571

mdesmet (approved):

PR #1020, opened 2024-03-21
Support metrics

mdesmet:
### Problem

Support metrics in lineage panel and tree view

### Solution

Only semantic_model is supported

### Screenshot/Demo

image

Currently incorrectly shown as view @AdiGajbhiye 

### How to test

Add metrics to your project for example:

```
semantic_models:
  #The name of the semantic model.
  - name: orders
    defaults:
      agg_time_dimension: ordered_at
    description: |
      Order fact table. This table is at the order grain with one row per order. 
    #The name of the dbt model and schema
    model: ref('orders')
```

## Checklist

### DIFF
--- a/src/manifest/parsers/graphParser.ts
+++ b/src/manifest/parsers/graphParser.ts
@@ -3,6 +3,8 @@ import {
   Analysis,
   Exposure,
   GraphMetaMap,
+  Metric,
+  MetricMetaMap,
   Model,
   Node,
   NodeGraphMap,
@@ -33,6 +35,7 @@ export class GraphParser {
     nodeMetaMap: NodeMetaMap,
     sourceMetaMap: SourceMetaMap,
     testMetaMap: TestMetaMap,
+    metricMetaMap: MetricMetaMap,
   ): GraphMetaMap {
     this.terminal.debug(
       "GraphParser",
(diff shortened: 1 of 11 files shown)


mdesmet (commented):
A metric is not a view
The icon should be updated.
Details should not be displayed for metrics.

src/manifest/parsers/graphParser.ts
```diff
@@ -68,7 +85,33 @@ export class GraphParser {
...
+        const currentNodes = unique(nodes)
+          .map(
+            this.mapToNode(
+              sourceMetaMap,
+              nodeMetaMap,
+              testMetaMap,
+              metricMetaMap,
+            ),
+          )
           .filter((n) => n instanceof Test)
```
AdiGajbhiye, line 115: Shouldn't this be `Metric`

src/webview_provider/newLineagePanel.ts
```diff
@@ -710,11 +706,24 @@ export class NewLineagePanel implements LineagePanelView {
...
+    if (nodeType === DBTProject.RESOURCE_TYPE_METRIC) {
+      return {
+        table: key,
+        label: splits[2],
+        url: tableUrl,
+        upstreamCount,
+        downstreamCount,
+        nodeType,
+        materialization: undefined,
+        tests: (graphMetaMap["tests"].get(key)?.nodes || []).map((n) => {
```
saravmajestic, line 718: semantic model could have test?

saravmajestic (commented):
LGTM, except Aditya's comment

src/webview_provider/newLineagePanel.ts
mdesmet: They don't, also exposures don't have tests. Will remove.

PR #1043, opened 2024-04-04
fix: save documentation in different yml s

saravmajestic:
### Problem
Issues:
1. 
```
when I am working on a branch and have edited files in multiple folders, each with their own models.yml file I get the following.

1. First file i try to document with the documentation editor, let's say it is in the orders folder, so i try to save its documentation into existing file, select the orders_models.yml file, and that works great
2. i try to document another file, in the customers folder lets say. if i pick existing file, it doesn't ask me to pick a file, it writes it to the order_models.yml file, which is not expected nor something I easily see how to fix without ending up copying and pasting the generated yml to the correct file
```
2. Deleting a custom test in model is not saved

### Solution
- Reason for issue#1 is patch path is not overwritten when user switches models and the model does not have any patch path. So it carried over the previous patch path. Updated code to reset patch path when switching models
- Removed delete option for deleting test in model. Will add a proper fix to delete the test in future PR

### Screenshot/Demo
https://www.awesomescreenshot.com/video/26412512?key=bb5ba9242b57f665e12d471b8076b31b

### How to test
Described in video

## Checklist

### DIFF
--- a/webview_panels/src/modules/documentationEditor/components/saveDocumentation/SaveDocumentation.tsx
+++ b/webview_panels/src/modules/documentationEditor/components/saveDocumentation/SaveDocumentation.tsx
@@ -37,11 +37,7 @@ const SaveDocumentation = (): JSX.Element | null => {
   };
 
   useEffect(() => {
-    if (!currentDocsData?.patchPath) {
-      return;
-    }
-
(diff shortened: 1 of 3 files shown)


AdiGajbhiye (approved):
lgtm

mdesmet (approved):

PR #1045, opened 2024-04-04
feat: execute sql for defer to prod

AdiGajbhiye:
### Problem

Implementing execute sql where defer to prod is enabled.

### Solution

- using `merge_from_artifact` to merge prod manifest with project manifest
- as this mutates `self.dbt` object, added functions to apply and clear prod manifest

### Screenshot/Demo

### How to test

## Checklist


src/dbt_client/dbtIntegration.ts
```diff
@@ -367,6 +367,8 @@ export interface DBTProjectIntegration extends Disposable {
   performDatapilotHealthcheck(
     args: HealthcheckArgs,
   ): Promise<ProjectHealthcheck>;
+  enableDefer(manifestPath: string): Promise<void>;
+  disableDefer(): Promise<void>;
```
mdesmet, line 371: I think we can contain the code in dbtCoreIntegration as this is not necessary in dbt cloud.

src/manifest/dbtProject.ts
```diff
@@ -409,6 +415,15 @@ export class DBTProject implements Disposable {
...
     await this.dbtProjectIntegration.rebuildManifest();
+    const currentConfig: Record<string, DeferConfig> =
+      this.deferToProdService.getDeferConfigByWorkspace();
+    const root = getProjectRelativePath(this.projectRoot);
+    if (currentConfig[root].deferToProduction) {
+      this.applyDeferToProject(
+        true,
+        currentConfig[root].manifestPathForDeferral,
+      );
+    }
```
mdesmet, line 426: This seems not the appropriate place.

src/webview_provider/insightsPanel.ts
```diff
@@ -193,6 +193,18 @@ export class InsightsPanel extends AltimateWebviewProvider {
...
+        currentConfig[root].deferToProduction !==
+        newConfig[root].deferToProduction
+      ) {
+        await this.dbtProjectContainer
+          .findDBTProject(workspaceFolder?.uri!)
+          ?.applyDeferToProject(
+            newConfig[root].deferToProduction,
+            newConfig[root].manifestPathForDeferral,
+          );
+      }
```
mdesmet, line 207: I think the merge should happen after any change in the current manifest.

src/webview_provider/insightsPanel.ts
```diff
@@ -193,6 +193,9 @@ export class InsightsPanel extends AltimateWebviewProvider {
           },
         });
       }
+      await this.dbtProjectContainer
+        .findDBTProject(Uri.parse(params.projectRoot))
```
mdesmet, line 197: Does projectRoot contains schema? Would it work with `Uri.parse(params.projectRoot, true)`?

src/manifest/dbtProject.ts
```diff
@@ -397,7 +397,7 @@ export class DBTProject implements Disposable {
     );
   }
 
-  private async rebuildManifest() {
```
mdesmet, line 400: This will also impact dbt cloud. I think it shouldn't

src/webview_provider/insightsPanel.ts
AdiGajbhiye: its absolute path. and not working with `Uri.parse(params.projectRoot, true)`

dbt_core_integration.py
```diff
@@ -361,6 +361,17 @@ def parse_project(self) -> None:
         self._sql_compiler = None
         self._sql_runner = None
 
+    def apply_defer_config(self, manifest_path: str, favor_state: bool) -> None:
+        with open(manifest_path) as f:
+            manifest = WritableManifest.from_dict(json.load(f))
+            selected = set()
+            self.dbt.merge_from_artifact(
+                self.adapter, other=manifest, selected=selected, favor_state=favor_state
+            )
```
mdesmet, line 370: I thought we discussed that everytime we need to parse the manifest we need to do the following:

1. parse the manifest as is
2. Write the artifacts
3. merge with defer manifest if defer is on

suryaiyer95: - isn't above the same? the manifest path getting loaded is the deffered manifest and that  is getting merged with current manifest

mdesmet: You are right, i missed the call in `rebuildManifest`

mdesmet (approved):

src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -1007,4 +1012,33 @@ export class DBTCoreProjectIntegration
...
+      const manifestPath = path.join(manifestFolder, DBTProject.MANIFEST_FILE);
+      await this.python?.lock<void>(
+        (python) =>
+          python!`project.apply_defer_config(${manifestPath}, ${favorState})`,
+      );
+    } else {
+      await this.python?.lock<void>(
+        (python) => python!`project.clear_defer_config()`,
+      );
+    }
```
mdesmet, line 1042: Can we not simply call `project.rebuildManifest() here?`

AdiGajbhiye: replacing `clear_defer_config` with `safe_parse_project`. Can't directly used `rebuildManifest()` cause this function itself is getting called in `rebuildManifest()` so that defer config applied on extension start.

mdesmet: That can be solved right. Just add the logic to `rebuildManifest` instead of splitting it into multiple functions.

mdesmet (approved):

PR #1051, opened 2024-04-08
fix: query explain unnecessary errors

saravmajestic:
### Problem
- DataPilot explain functionality is not giving text box to ask additional questions.
(Also for a few seconds) - valid SQL model needs to be open in editor message flashes. Also, the text shows for this SQL query instead of dbt model when explanation is printed. 

### Solution
- get follow up questions api was expecting a valid file path which was not passed from webview properly. Added the file path from where the conversation is initiated

### Screenshot/Demo
NA

### How to test
- Open a valid model
- right click -> datapilot -> explain
- Should see valid explanation and follow up questions if available

## Checklist


AdiGajbhiye (approved):
lgtm

mdesmet (approved):

PR #1061, opened 2024-04-12
fix: have to print env vars twice to display anything

AdiGajbhiye:
### Problem

have to print env vars twice to display 

### Solution

adding delay between showing and writing to terminal

### Screenshot/Demo

### How to test

## Checklist

### DIFF
--- a/src/manifest/pythonEnvironment.ts
+++ b/src/manifest/pythonEnvironment.ts
@@ -32,10 +32,11 @@ export class PythonEnvironment implements Disposable {
     }
   }
 
-  printEnvVars() {
+  async printEnvVars() {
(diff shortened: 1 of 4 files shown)


src/manifest/pythonEnvironment.ts
```diff
@@ -32,10 +32,11 @@ export class PythonEnvironment implements Disposable {
     }
   }
 
-  printEnvVars() {
+  async printEnvVars() {
+    this.dbtTerminal.show(true);
+    await new Promise((resolve) => setTimeout(resolve, 100));
```
mdesmet, line 37: Is it possible to fix it for all occurrences instead of just this one?

PR #1090, opened 2024-04-29
Fix/docs bugs

saravmajestic:
### Problem
- UI issue in dbt docs share functionality in doc editor: https://altimateai.atlassian.net/browse/AI-1687
- Too many api calls for dbt docs on extension load
- When @ mentioning users, github user names are showing

### Solution
- fixed ui issues in doc editor panel
- added extra text `Altimate` near user names when @ mentioning to indicate altimate users
- Backend api is changed to send conversations with shared docs. Modified extension to handle the new response
- Instead of getting shares after every manifest cache changed event, make it a single call after all projects are initialized

### Screenshot/Demo

### How to test
- Load sample_dbt_projects repo in vscode with dbt power user extension
- Single api call `dbt/dbt_docs_share/all?projects=` should be made after all projects are initialized
- While creating new comment, and using @ mention, `ALtimate` text should be visible after user name

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
           "dbt.enableCollaboration": {
             "type": "boolean",
             "description": "Enable dbt docs collaboration.",
-            "default": false
+            "default": true
           },
           "dbt.enableNewDocsPanel": {
             "type": "boolean",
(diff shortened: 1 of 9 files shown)


AdiGajbhiye (approved):
lgtm

mdesmet (approved):

PR #1091, opened 2024-04-29
fix: run all scan in project healthcheck

AdiGajbhiye:
### Problem

 in project healthcheck, run all scan errors out at the start. 

### Solution

file required was not properly handled for run all scan. now in case of run all checks, assuming manifest and catalog both are required.

### Screenshot/Demo

### How to test

## Checklist


suryaiyer95 (commented):
Smoke test looks good

saravmajestic (approved):
Looks fine, it will be better to use an enum instead of string comparison ex: `if (args.configType === "Saas") {`

PR #1093, opened 2024-04-30
query panel migration revamp

saravmajestic:
### Problem
- Migrate query results panel to webview react

### Solution
- migrated required components and functionalities to webview components
- fixed query cancel issue (applies for existing query panel)
- fixed switching to query results panel if a query is triggered before panel render is completed (applies for existing query panel)
- alternative colors for the rows in perspective table
- Added `QueryPanelV2Activated` telemetry event if activated and `QueryPanelV2Deactivated` if deactivated

### Screenshot/Demo
image

### How to test

- Load the extension
- enable `dbt.enableQueryPanelV2` in vscode settings
- switch to a model and exectute the model. Should be able to see latest panel

## Checklist


src/dbt_client/dbtCoreIntegration.ts
```diff
@@ -367,13 +367,13 @@ export class DBTCoreProjectIntegration
...
     );
-    await this.createPythonDbtProject(queryThread);
-    await queryThread.ex`project.init_project()`;
     return new QueryExecution(
       async () => {
         queryThread.kill(2);
       },
       async () => {
+        await this.createPythonDbtProject(queryThread);
+        await queryThread.ex`project.init_project()`;
```
AdiGajbhiye, line 376: why this change?

AdiGajbhiye (approved):
lgtm

src/dbt_client/dbtCoreIntegration.ts
mdesmet: Doesn't matter as this is not affecting main thread.

mdesmet (approved):

mdesmet (approved):

PR #1105, opened 2024-05-06
fix: pricing error handling

saravmajestic:
### Problem
- If user has exhausted execution limit in backend, 402 payment required error is thrown. Extension is not handling the error.

### Solution
- Update the fetch and fetchAsStream API methods to show right error message

### Screenshot/Demo

image

### How to test
- exhaust your allowed limit per month
- try to run an execution, for ex: generate a doc
- should see the error with saas ui link

## Checklist


mdesmet (approved):
LGTM

PR #1110, opened 2024-05-07
fix: column names in lowercase setting

saravmajestic:
### Problem

### Solution

### Screenshot/Demo

### How to test

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,11 @@
       {
         "title": "dbt Power User",
         "properties": {
+          "dbt.showColumnNamesInLowercase": {
+            "type": "boolean",
+            "description": "If enabled, documentation editor will show the column names in lower case regardless of database.",
+            "default": false
+          },
           "dbt.dbtPythonPathOverride": {
             "type": "string",
             "description": "Path to a python executable or entrypoint. If undefined, we will use the Python extension's configured environment. Most users will not need to change this setting without specific reasoning such as custom wrappers and should instead modify the Python extensions selected interpreter."
--- a/webview_panels/src/modules/documentationEditor/state/documentationSlice.ts
+++ b/webview_panels/src/modules/documentationEditor/state/documentationSlice.ts
@@ -26,6 +26,7 @@ export const initialState = {
   conversations: {},
   showConversationsRightPanel: false,
   collaborationEnabled: false,
+  showColumnNamesInLowercase: false,
 } as DocumentationStateProps;
 
 const documentationSlice = createSlice({
@@ -78,6 +79,14 @@ const documentationSlice = createSlice({
       action: PayloadAction<DocumentationStateProps["project"]>,
     ) => {
       state.project = action.payload;
+    },
+    setShowColumnNamesInLowercase: (
+      state,
+      action: PayloadAction<
+        DocumentationStateProps["showColumnNamesInLowercase"]
+      >,
+    ) => {
+      state.showColumnNamesInLowercase = action.payload;
     },
     updateCurrentDocsTests: (
       state,
@@ -232,5 +241,6 @@ export const {
   updateConversationsRightPanelState,
   updateSelectedConversationGroup,
   updateCollaborationEnabled,
+  setShowColumnNamesInLowercase,
 } = documentationSlice.actions;
 export default documentationSlice;
--- a/src/webview_provider/docsEditPanel.ts
+++ b/src/webview_provider/docsEditPanel.ts
@@ -151,6 +151,9 @@ export class DocsEditViewPanel implements WebviewViewProvider {
         collaborationEnabled: workspace
           .getConfiguration("dbt")
           .get<boolean>("enableCollaboration", false),
+        showColumnNamesInLowercase: workspace
+          .getConfiguration("dbt")
+          .get<boolean>("showColumnNamesInLowercase", false),
       });
     }
   }
--- a/src/webview_provider/newDocsGenPanel.ts
+++ b/src/webview_provider/newDocsGenPanel.ts
@@ -238,6 +238,9 @@ export class NewDocsGenPanel
           collaborationEnabled: workspace
             .getConfiguration("dbt")
             .get<boolean>("enableCollaboration", false),
+          showColumnNamesInLowercase: workspace
+            .getConfiguration("dbt")
+            .get<boolean>("showColumnNamesInLowercase", false),
         });
         break;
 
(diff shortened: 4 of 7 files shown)


package.json
```diff
@@ -101,6 +101,11 @@
       {
         "title": "dbt Power User",
         "properties": {
+          "dbt.showColumnNamesInLowercase": {
+            "type": "boolean",
+            "description": "If enabled, documentation editor will show the column names in lower case regardless of database.",
```
saravmajestic, line 106: @mdesmet pls check this sentence

AdiGajbhiye (approved):
lgtm

src/commands/tests/staleModelColumnTest.ts
```diff
@@ -84,11 +84,8 @@ export class StaleModelColumnTest implements AltimateScanStep {
         // do model-level checks here.
         const modelDict =
           altimateCatalog[projectName + projectRootUri][modelKey];
-        const existingColumnsLowered = Object.keys(value.columns).map((key) =>
```
saravmajestic, line 87: removed unused variable

webview_panels/src/modules/documentationEditor/state/documentationSlice.ts
```diff
@@ -142,7 +142,7 @@ const documentationSlice = createSlice({
       }
       state.currentDocsData.columns = columns.map((column) => {
         const existingColumn = state.currentDocsData?.columns.find(
-          (c) => column.name?.toLowerCase() === c.name.toLowerCase(),
+          (c) => column.name === c.name,
```
AdiGajbhiye, line 145: `isColumnNameEqual`??

src/manifest/dbtProject.ts
```diff
@@ -1005,18 +1006,20 @@ select * from renamed
     }
     const columns: Record<string, ColumnMetaData> = {};
     Object.entries(node.columns).forEach(([k, v]) => {
-      columns[k.toLowerCase()] = v;
+      columns[getColumnNameByCase(k)] = v;
```
saravmajestic, line 1009: @AdiGajbhiye @mdesmet @suryaiyer95 here we are merging columns from local manifest and from db. This particular line is creating map of column name and column from manifest nodes. Since this actually takes column names from schema.yml, i am thinking we will not need this conversion `getColumnNameByCase` here. If user has lower case column name in schema.yml and from db if it is upper case, and if i dont do this conversion, both columns will not be matched and we might see duplicate columns. What do you guys think if I leave this change as it is? Will this have any side effects?

src/utils.ts
```diff
@@ -173,3 +174,33 @@ export const deepEqual = (obj1: any, obj2: any): boolean => {
...
+export const getColumnNameByCase = (columnName: string) => {
+  const showColumnNamesInLowercase = workspace
+    .getConfiguration("dbt")
+    .get<boolean>("showColumnNamesInLowercase", true);
+  return showColumnNamesInLowercase ? columnName.toLowerCase() : columnName;
+};
+
+export const isColumnNameEqual = (
+  columnNameA: string | undefined,
+  columnNameB: string | undefined,
```
mdesmet, line 187: Maybe it is better to call them by what is the source of the column name. if A is coming from yaml and B from database.

that way it will be easier later on if we need to process the one other than the other.

src/manifest/dbtProject.ts
suryaiyer95: So duplicates will be created right in this case?

suryaiyer95: We should Ideally match making them more lower case treating existing name in schema.yml as priority? @mdesmet what do you think?

src/manifest/dbtProject.ts
```diff
@@ -1005,18 +1006,20 @@ select * from renamed
...
     });
 
     for (const c of columnsFromDB) {
-      const existing_column = columns[c.column.toLowerCase()];
+      const existing_column = columns[getColumnNameByCase(c.column)];
       if (existing_column) {
-        existing_column.data_type = existing_column.data_type || c.dtype;
+        existing_column.data_type = (
+          existing_column.data_type || c.dtype
+        )?.toLowerCase();
```
suryaiyer95, line 1017: Should the lower case config be used here as well?

src/services/dbtTestService.ts
```diff
@@ -145,7 +145,9 @@ export class DbtTestService {
 
     const column =
       model.columns &&
-      model.columns.find((yamlColumn: any) => yamlColumn.name === column_name);
+      model.columns.find((yamlColumn: any) =>
+        isColumnNameEqual(yamlColumn.name, column_name),
+      );
```
suryaiyer95, line 150: We won't need casing here?

suryaiyer95 (commented):
@AdiGajbhiye Does this have any impact on column lineage? We have sync with db there as well, do we need to do it?

src/manifest/dbtProject.ts
saravmajestic: for data_type it will always be in lowercase from now

src/services/dbtTestService.ts
saravmajestic: `column_name` is not from db, it is from metadata of test node, so it will match the case of schema.yml

src/utils.ts
```diff
@@ -173,3 +174,39 @@ export const deepEqual = (obj1: any, obj2: any): boolean => {
...
+
+  if (showColumnNamesInLowercase) {
+    return columnNameFromYml.toLowerCase() === columnNameFromDb.toLowerCase();
+  }
+
+  return false;
+};
+
+export const isQuotedIdentifier = (name: string) =>
+  /^\d|(?=.*[a-z])(?=.*[A-Z])|[- "']/.test(name);
```
mdesmet, line 212: I think this is the wrong approach. You should test if is an unquoted identifier using the `UNQUOTED_CASE_INSENSITIVE = re.compile("^([_A-Za-z]+[_A-Za-z0-9$]*)$")`

src/manifest/dbtProject.ts
mdesmet: That's what we discussed indeed. If is not a quoted identifier, the matching should occur in lowercase and we should prefer the yaml identifier if it is already there.

src/services/dbtTestService.ts
mdesmet: Variable name can be improved, please make it clear what is the source of the column_name. for example columnNameFromManifestTestData. Longer but will help if we need to look at this again in a few months.

src/manifest/dbtProject.ts
saravmajestic: ok, this is already handled in that way. I have modified the variable names to be more clear.

src/services/dbtTestService.ts
saravmajestic: updated

src/utils.ts
saravmajestic: modified using this regex

mdesmet (approved):

PR #1111, opened 2024-05-07
fix: show alert when switching models

saravmajestic:
### Problem
After making some changes in doc editor for a model, if user switches to other files in editor, unsaved changes in the model is lost

### Solution
Show an alert to user while switching editors with unsaved changes in doc editor
Options will be:
- discard the changes
- save and show doc editor for opened file
- stay with unsaved changes

### Screenshot/Demo
https://www.loom.com/share/34ee530df193427a93175d0622d7ae31

### How to test
- Open a model
- open doc editor
- add/change some description for column
- open another model/yml file
- doc editor should show the alert and allow user to select an option

## Checklist


src/webview_provider/newDocsGenPanel.ts
```diff
@@ -62,32 +62,14 @@ export class NewDocsGenPanel
       queryManifestService,
       userService,
     );
-
```
saravmajestic, line 65: instead of sending separate message to webview, sending the tests data along with renderdocumentation message

AdiGajbhiye (approved):
lgtm

webview_panels/src/uiCore/uiToolkitStories/Components.stories.tsx
```diff
@@ -136,6 +141,20 @@ export const UIKit = {
...
+          title="Save changes?"
+          actionsFooter={
+            <>
+              <Button>Stay, do not proceed</Button>
+              <Button>Cancel and proceed</Button>
+            </>
+          }
+        >
+          Your changes will be lost!
+        </AlertModal>
```
mdesmet, line 157: Shouldn't we use vscode api?

mdesmet: Using vscode api is better as it would be displayed even if the panel is not selected.

saravmajestic: that is good idea. it actually works even if switch panels. thank you

webview_panels/src/modules/documentationEditor/useIncomingDocsDataHandler.tsx
```diff
@@ -0,0 +1,84 @@
...
+    }
+  };
+
+  const showAlert = incomingDocsData
+    ? Object.keys(incomingDocsData).length > 0
+    : false;
+
+  useEffect(() => {
+    if (showAlert) {
+      executeRequestInSync("showInformationMessage", {
```
mdesmet, line 64: Doesn't this time out? Maybe better use showWarning or error?

mdesmet (approved):

PR #1117, opened 2024-05-08
fix: quoted identifier regex and  generate documentation for missing columns not working

saravmajestic:
### Problem
- When using snowflake db (or any db with column name as uppercase), bulk generate docs for missing columns is generating docs for all columns
- Quoted identifiers are not parsed properly

Resolves: https://altimateai.atlassian.net/jira/software/c/projects/AI/issues/AI-1720
Partially resolves: https://github.com/AltimateAI/vscode-dbt-power-user/issues/1119

### Solution
- While doing bulk generate for missing columns, we do sync the columns from db. The synced columns were not merged with local column updates. Fix applied to merge with local data and then trigger bulk update for missing columns
- updated the quoted identifier regex specific to adapter and added a config for user to specify own regex as well

### Screenshot/Demo
https://www.loom.com/share/54f076745ab246f4a89accca77ab556c

### How to test
- use snowflake profile for a dbt project
- check if your model shows column name in lowercase
- add documentation for few columns and save 
- do bulk generate for missing columns
- should generate docs only for columns without desc

## Checklist


webview_panels/src/modules/documentationEditor/state/documentationSlice.ts
```diff
@@ -130,28 +130,18 @@ const documentationSlice = createSlice({
...
       }>,
     ) => {
       if (!state.currentDocsData) {
         return;
       }
-      state.currentDocsData.columns = columns.map((column) => {
-        const existingColumn = state.currentDocsData?.columns.find(
-          (c) => column.name?.toLowerCase() === c.name.toLowerCase(),
+
+      state.currentDocsData.columns =
```
saravmajestic, line 140: extracted this to utility function

AdiGajbhiye (approved):
lgtm

package.json
```diff
@@ -101,6 +101,10 @@
       {
         "title": "dbt Power User",
         "properties": {
+          "dbt.unquotedCaseInsensitiveIdentifierRegex": {
```
saravmajestic, line 104: new config for users if regex [here](https://github.com/AltimateAI/vscode-dbt-power-user/pull/1117/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R211) does not work for a project

src/webview_provider/docsEditPanel.ts
```diff
@@ -417,7 +417,10 @@ export class DocsEditViewPanel implements WebviewViewProvider {
       }
 
       // new column, save the name by checking the config
-      return { ...c, name: getColumnNameByCase(c.name) };
+      return {
```
saravmajestic, line 420: This is while syncing with db and we should send name as the case received from db to webview, so while saving we will be able to identify right case and quote attribute

mdesmet (approved):

PR #1136, opened 2024-05-17
fix: save alert when switching in multi project

saravmajestic:
### Problem
If vscode workspace contains multiple projects, and if user modifies documentation of a model in a project and switched to another model in another project, the save alert warning message shows up. but the save is not working

### Solution
- This is mainly because of getting project by current active open file.
- Modified code to send filePath along with documentation data and use that to find project and save the data

### Screenshot/Demo

### How to test
- Open a workspace with multi project setup
- modify documentation of a model in a project
- without saving, open another file from another project
- in the warning message, click save changes
- modified documentation should be saved to right model in right project

## Checklist

### DIFF
--- a/src/services/docGenService.ts
+++ b/src/services/docGenService.ts
@@ -199,6 +199,7 @@ export class DocGenService {
       generated: false,
       resource_type: currentNode.resource_type,
       uniqueId: currentNode.uniqueId,
+      filePath,
       columns: Object.values(docColumns).map((column) => {
         return {
           name: column.name,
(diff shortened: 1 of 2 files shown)


src/services/docGenService.ts
```diff
@@ -199,6 +199,7 @@ export class DocGenService {
       generated: false,
       resource_type: currentNode.resource_type,
       uniqueId: currentNode.uniqueId,
+      filePath,
```
saravmajestic, line 202: passing the current file path to webview to send it back while saving

PR #1143, opened 2024-05-19
Make path optional

mdesmet:
### Problem

Packages folder is not downloaded on dbt cloud, so we can't resolve the path to any dbt entities located there.

### Solution

Make path optional

### Screenshot/Demo

### How to test

## Checklist

### DIFF
--- a/src/services/dbtTestService.ts
+++ b/src/services/dbtTestService.ts
@@ -178,7 +178,7 @@ export class DbtTestService {
     if (macro) {
       // return the file path if it ends with sql
       const macroData = macroMetaMap.get(`test_${testName}`);
-      return macroData?.path.endsWith(".sql") ? macroData?.path : undefined;
+      return macroData?.path?.endsWith(".sql") ? macroData?.path : undefined;
     }
   };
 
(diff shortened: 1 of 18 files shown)


mdesmet:
@AdiGajbhiye , @saravmajestic: Can we assess the impact of making this change. What will be the effect on documentation editor, project health check and lineage.

saravmajestic:
> @AdiGajbhiye , @saravmajestic: Can we assess the impact of making this change. What will be the effect on documentation editor, project health check and lineage.

just by looking at the changes, tests added as macros will be affected. I will test the doc editor and will let you know

src/services/dbtTestService.ts
```diff
@@ -103,10 +103,18 @@ export class DbtTestService {
       return;
     }
 
-    const patchPath = node?.patch_path.includes("://")
+    const patchPath = node?.patch_path?.includes("://")
       ? path.join(project.projectRoot.fsPath, node.patch_path.split("://")[1])
       : node.patch_path;
 
+    if (!patchPath) {
```
mdesmet, line 110: This seems unrelated to this change. The patch_path will always be there, if there is no file exists check it should always exist.

saravmajestic: and this is the file exist check

PR #1148, opened 2024-05-21
fix: add guards for doc editor

saravmajestic:
### Problem
- We cannot display documentation editor if there is any error in project or active open file is not a model. So it is confusing for the user to know what is the reason for not showing documentation of model

### Solution
- Show warning message if the open file is not valid model
- show error message if there is any error in project and a button to view problems tab

### Screenshot/Demo
https://www.loom.com/share/b8f5ca6ac24a460690af485830c50723?sid=03fb89e1-9b7a-4601-8f3e-e075fb19c746

### How to test
- open a schema.yml file
- should see warning message
- delete the profile for the project, delete the target directory
- on refresh should see the error message with button

## Checklist


mdesmet (approved):
LGTM except the text. @anandgupta42: any views?

PR #1151, opened 2024-05-21
fix: change session id for column lineage

AdiGajbhiye:
## Overview

- Using `env.sessionId` which will have values like `2fcc86de-6267-4471-8e04-07904223158b1716315453561`. It changes on vscode restart.
- Remove previous `sessionId`

### Problem

Describe the problem you are solving. Mention the ticket/issue if applicable.

### Solution

Describe the implemented solution. Add external references if needed.

### Screenshot/Demo

A picture is worth a thousand words. Please highlight the changes if applicable.

### How to test

- Steps to be followed to verify the solution or code changes
- Mention if there is any settings configuration added/changed/deleted

## Checklist

- [ ] I have run this code and it appears to resolve the stated issue
- [ ] `README.md` updated and added information about my change

PR #1153, opened 2024-05-22
feat: add counts while getting doc gen

svipulrao98:
### Problem

Send the column Counts while getting the Doc Gen.

### Solution

### Screenshot/Demo

### How to test

## Checklist

### DIFF
--- a/src/altimate.ts
+++ b/src/altimate.ts
@@ -172,6 +172,7 @@ interface DocsGenerateModelRequestV2 {
   gen_model_description: boolean;
   column_index_count: number | undefined;
   session_id: string | undefined;
+  is_bulk_gen: boolean;
 }
 
 export interface DocsGenerateResponse {
--- a/src/altimate.ts
+++ b/src/altimate.ts
@@ -170,7 +170,8 @@ interface DocsGenerateModelRequestV2 {
   };
   prompt_hint?: string;
   gen_model_description: boolean;
-  columns_count: number;
+  column_index_count: number | undefined;
+  session_id: string | undefined;
 }
 
 export interface DocsGenerateResponse {
--- a/src/altimate.ts
+++ b/src/altimate.ts
@@ -170,6 +170,7 @@ interface DocsGenerateModelRequestV2 {
   };
   prompt_hint?: string;
   gen_model_description: boolean;
+  columns_count: number;
 }
 
 export interface DocsGenerateResponse {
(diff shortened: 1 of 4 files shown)


src/services/docGenService.ts
```diff
@@ -107,6 +110,7 @@ export class DocGenService {
                 adapter,
                 message,
                 columns,
+                1,
```
saravmajestic, line 113: this should be columnCount?

src/altimate.ts
```diff
@@ -170,6 +170,7 @@ interface DocsGenerateModelRequestV2 {
   };
   prompt_hint?: string;
   gen_model_description: boolean;
+  columns_count: number;
```
saravmajestic, line 173: variable naming could be improved. `current_column_index` or `completed_columns_count`?

suryaiyer95: Agree

src/services/docGenService.ts
suryaiyer95: ColumnIndex right?

suryaiyer95: actually no. It should be completedColumns count and is applicable only for doc gen

src/altimate.ts
suryaiyer95: I think along with this, we send an unique id. session_id.table_id.bulk_gen or somthing like that. 

https://github.com/AltimateAI/vscode-dbt-power-user/pull/1151/files#diff-f5d2312039a56f9d5d5c2eff4188b2ab44446ae157202ee46a51da2a5ef242c7R16

You can find it here

suryaiyer95: You can store this in the backend and say (2000 executions) charged for (300) or smthg like that

suryaiyer95: vscode_session_id-{model-unique-id}-{bulk.gen}

src/services/docGenService.ts
```diff
@@ -402,6 +416,8 @@ export class DocGenService {
                   message.user_instructions.prompt_hint || "generate",
               },
               follow_up_instructions: message.follow_up_instructions,
+              column_index_count: columnIndexCount,
```
suryaiyer95, line 419: @saravmajestic When retry happens due to rate limiting, the columnIndexCount does not change right?

saravmajestic: it wont change. it will pickup same index

saravmajestic (approved):

AdiGajbhiye (approved):
lgtm

PR #1155, opened 2024-05-23
fix: guards for lineage panel

saravmajestic:
### Problem
User could not understand why lineage is not showing in some cases

### Solution
- Added a warning message to user to open a valid node to see lineage 
- Also if there is any error in project, show the error in lineage panel

### Screenshot/Demo
https://www.loom.com/share/13522a7b28a34d7e89755410eab42495?sid=775dc7de-2b38-4293-95e8-8013f23a972a

### How to test
- open a schema.yml file and open lineage panel
- should see warning message
- If project has any error, error will also be shown

## Checklist


mdesmet (approved):
LGTM except message @anandgupta42, @ppradnesh

PR #1183, opened 2024-05-31
feat: show external projects in lineage

saravmajestic:
### Problem
in lineage panel for dbt core project, if a model depends on another entity from another dbt project, it is not easily identifiable

### Solution
- added a new label `ext` in the lineage node
- on hover, show `External project: <table name>` to indicate from which project this node is created
- send a telemetry event when we detect connected projects. Event will be sent only once per session

### Screenshot/Demo
https://www.loom.com/share/d0c72f1b69ea4353a7ffdae7702b25ea?sid=de5e029c-fcd0-4936-a4bb-e8741d3f4813

### How to test
- Setup multi project: https://github.com/Bl3f/dbt-loom-example
- Open stg_revenue in finance project
- Open lineage panel
- should see orders node with `ext` label
- on hover should show `External Project: model.core.orders`

## Checklist


src/webview_provider/newLineagePanel.ts
```diff
@@ -814,6 +848,13 @@ export class NewLineagePanel implements LineagePanelView {
     };
   }
 
+  private isExternalProject(
+    node: NodeMetaData | SourceMetaData,
+    projectName: string,
+  ) {
+    return node.package_name !== projectName;
```
saravmajestic, line 855: Only package_name has information about the project. I am assuming we dont show nodes from installed dependency packages. @mdesmet pls check if this ok

src/webview_provider/newLineagePanel.ts
```diff
@@ -759,6 +789,7 @@ export class NewLineagePanel implements LineagePanelView {
         upstreamCount,
         downstreamCount,
         nodeType,
+        isExternalProject: this.isExternalProject(_node, projectName),
```
saravmajestic, line 792: This is for sources. I beleive we could refer sources from other projects as well, if not needed @mdesmet pls lmk

src/webview_provider/newLineagePanel.ts
```diff
@@ -286,6 +291,20 @@ export class NewLineagePanel implements LineagePanelView {
     );
   }
 
+  private sendExternalProjectTelemetryEventSent(body: {
```
saravmajestic, line 294: Sends an event only once per session

AdiGajbhiye (approved):
lgtm

src/webview_provider/newLineagePanel.ts
mdesmet: We do show nodes from installed packages. This is not correct.

mdesmet: Also this business logic should be returned by the parser instead of doing it in a consumer of the metadata

mdesmet (changes requested):

src/webview_provider/newLineagePanel.ts
mdesmet: I personally prefer to send it as an attribute of our events and base it on if dbt-loom is installed or not (that means the intention is that the user would like to use it).

@anandgupta42 : WDYT?

src/webview_provider/newLineagePanel.ts
mdesmet: In dbt cloud this is not allowed but I think in dbt loom it is supported as those sources become part of the manifest.

src/webview_provider/newLineagePanel.ts
anandgupta42: Yes. We should only send it if there project has dbt loom configured.

src/telemetry/index.ts
```diff
@@ -1,13 +1,44 @@
...
+import { PythonEnvironment } from "../manifest/pythonEnvironment";
 
 @provideSingleton(TelemetryService)
 export class TelemetryService implements vscode.Disposable {
+  private dbtLoomInstalled = "false";
+  constructor(
+    private pythonEnvironment: PythonEnvironment,
+    private commandProcessExecutionFactory: CommandProcessExecutionFactory,
+  ) {
+    this.isDbtLoomInstalled().then((isInstalled) => {
```
saravmajestic, line 14: since we are dealing with less no of users, this is not re calculated when manifest changes. If needed we can add that as well

src/telemetry/index.ts
```diff
@@ -1,13 +1,44 @@
 import TelemetryReporter from "@vscode/extension-telemetry";
 import * as vscode from "vscode";
-import { provideSingleton } from "../utils";
+import { getFirstWorkspacePath, provideSingleton } from "../utils";
+import { CommandProcessExecutionFactory } from "../commandProcessExecution";
+import { PythonEnvironment } from "../manifest/pythonEnvironment";
 
 @provideSingleton(TelemetryService)
 export class TelemetryService implements vscode.Disposable {
+  private dbtLoomInstalled = "false";
```
saravmajestic, line 9: setting this to string, because telemetry accepts string attributes

mdesmet (approved):

PR #1189, opened 2024-06-03
fix: query results copy from cell issue

saravmajestic:
### Problem
New query results view does not allow user to select contents of cells in table

### Solution
Enabled text select for cells

### Screenshot/Demo

### How to test
- Run a query and open query results panel
- try to select the text in the gird
- user should be able to select the text

## Checklist

### DIFF
--- a/webview_panels/src/modules/queryPanel/components/perspective/PerspectiveViewer.tsx
+++ b/webview_panels/src/modules/queryPanel/components/perspective/PerspectiveViewer.tsx
@@ -128,13 +128,14 @@ const PerspectiveViewer = ({
     const id = "altimate-styles";
     shadowRoot.getElementById(id)?.remove();
 
-    if (currentTheme === "Vintage") {
-      const style = document.createElement("style");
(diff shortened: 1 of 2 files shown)


webview_panels/src/modules/queryPanel/components/perspective/PerspectiveViewer.tsx
```diff
@@ -128,13 +128,14 @@ const PerspectiveViewer = ({
     const id = "altimate-styles";
     shadowRoot.getElementById(id)?.remove();
 
-    if (currentTheme === "Vintage") {
```
saravmajestic, line 131: removed this check and added the styles always

webview_panels/src/modules/queryPanel/components/perspective/perspective.scss
```diff
@@ -1,4 +1,11 @@
 regular-table {
+  td, th{
```
saravmajestic, line 2: this is the actual fix

mdesmet (approved):

PR #1195, opened 2024-06-04
feat: add views for lineage column nodes

saravmajestic:
### Problem
While viewing column level lineage, user wants to understand the transformation of the column

### Changes
- Added border around column and a label to show type of view
- Added a legend near bottom left end
- Model edges will be hidden when column lineage is viewed

### Screenshot/Demo
image

https://www.loom.com/share/94b76383ead744b1a8a685ecf859edb3?sid=c8b55b30-8fb9-4bcf-911e-7a2fc29adbf7

### How to test
- Open tuva project -> dbt_packages\the_tuva_project\models\cms_hcc\final\cms_hcc__patient_risk_factors.sql
- Open lineage panel
- Open column lineage for `coefficient` column
- should see different colored edges

## Checklist


new_lineage_panel/src/App.tsx
```diff
@@ -339,6 +340,7 @@ function App() {
               >
                 <Background />
                 <Controls />
+                <LineageLegend />
```
AdiGajbhiye, line 343: i think we should move `LineageLegend` out of `ReactFlow` tag. cant say how well this will behave with `ReactFlow` canvas especially with for z-index.

new_lineage_panel/src/TableDetails.tsx
```diff
@@ -248,6 +249,11 @@ const TableDetails = () => {
         flow.getNodes(),
         flow.getEdges()
       );
+      // Model edges will be hidden when column lineage is selected, so unhide them
+      // _edges.forEach(edge => edge.hidden = false)
```
AdiGajbhiye, line 253: can remove commented code.

new_lineage_panel/src/graph.ts
```diff
@@ -340,12 +349,12 @@ const processColumnLineage = async (
     right ? contains(curr, e.source) : contains(curr, e.target)
   );
   const newCurr = columnLineage.map((e) => (right ? e.target : e.source));
-  const collectColumns: Record<string, string[]> = {};
+  const collectColumns: Record<string, {column: string, lensType?: LensTypes}[]> = {};
```
AdiGajbhiye, line 352: i think we should create separate type for collectColumns. This is getting in too many places and will probably change in future.

new_lineage_panel/src/App.tsx
saravmajestic: fixed

new_lineage_panel/src/TableDetails.tsx
saravmajestic: removed wherever i have added commented code

new_lineage_panel/src/graph.ts
saravmajestic: done

new_lineage_panel/src/graph.ts
```diff
@@ -833,3 +856,105 @@ export const calculateNodeCount = async (
...
+        }
+      });
+      memo.edges = Array.from(new Set([...memo.edges, ..._edgeIds]));
+    }
+
+    return memo;
+  }, {nodes: [] as Node[], edges: [] as string[]});
+};
+
+export const highlightColumnConnections = (
```
AdiGajbhiye, line 920: this whole logic could be simplified. but for now its ok.

new_lineage_panel/src/graph.ts
```diff
@@ -472,19 +488,26 @@ export const removeColumnNodes = (
...
       if (!(t in collectColumns)) {
         collectColumns[t] = _columns;
         continue;
       }
+      // merge _columns with collectColumns[t]
+      const updatedColumns = _columns.map((c) => {
+        const existing = collectColumns[t].findIndex((x) => x.column === c.column);
+        if (existing === -1) return c;
+        if (c.lensType) collectColumns[t][existing].lensType = c.lensType;
+        return null;
```
AdiGajbhiye, line 507: in this case we are basically ignoring column when `lensType` is not there. maybe it would be better to show it with unknown lensType

AdiGajbhiye (approved):
lgtm

new_lineage_panel/src/graph.ts
saravmajestic: here we are not ignoring the column. We update the lenstype if the column is already existing.

PR #1209, opened 2024-06-13
fix: render tests for quoted column names

saravmajestic:
### Problem
Doc editor does not display tests for quoted columns

### Solution
In manifest, quoted column name are stored with extra quotes for tests. So, modified testParser to remove extra quotes
image

### Screenshot/Demo
image

### How to test
- create quoted column
- add tests in schema.yml
- doc editor should show the tests

## Checklist

### DIFF
--- a/src/manifest/parsers/testParser.ts
+++ b/src/manifest/parsers/testParser.ts
@@ -47,7 +47,10 @@ export class TestParser {
               database,
               schema,
               alias,
-              column_name,
+              // for quoted column names, remove the quotes
+              // ex: in manifest, it will be stored as "column_name": "\"Customer ID\"" for tests
+              // here we remove the extra quotes
+              column_name: column_name?.replace(/^"(.*)"$/, "$1"),
               test_metadata,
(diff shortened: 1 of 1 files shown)


src/manifest/parsers/testParser.ts
```diff
@@ -47,7 +47,10 @@ export class TestParser {
               database,
               schema,
               alias,
-              column_name,
+              // for quoted column names, remove the quotes
+              // ex: in manifest, it will be stored as "column_name": "\"Customer ID\"" for tests
+              // here we remove the extra quotes
+              column_name: column_name?.replace(/^"(.*)"$/, "$1"),
```
mdesmet, line 53: Instead of using regexes can we not just use replace operations and replace any \" with " and remove the enclosing quotes?

saravmajestic: modified

mdesmet (approved):
There can also be quotes within the column name, that is obviously rather rare, should be easy to fix though.

Issue #1219, opened 2024-06-17
Documentation editor mess up tests when same test exist with different configs


saravmajestic:
### Expected behavior

- should preserve the tests as it is

### Actual behavior

- replaces config of tests

### Steps To Reproduce

I have two accepted values tests with two different configs but the Documentation Editor changed them to give them the same config.

#### before

       
```
tests:
          - accepted_values:
              values: ['deep_dish', 'detroit', 'thin_crust']
              config:
                severity: 'warn'
                where: "menu_item = 'pizza'"
          - accepted_values:
              values: ['gyro', 'sub', 'footlong']
              config:
                severity: 'warn'
                where: "menu_item = 'sandwich'"

```

#### after

```
- accepted_values:
              values:
                - deep_dish
                - detroit
                - thin_crust
              config: &a1
                severity: warn
                where: menu_item = 'pizza'
          - accepted_values:
              values:
                - gyro
                - sub
                - footlong
              config: *a1
```

### Operating System

MacOs

### dbt version

1.7

### dbt Adapter

any

### dbt Power User version

any

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

saravmajestic:
Sweep: In src\webview_provider\docsEditPanel.ts, function: getTestDataByColumn, fix the tests which has same test name with different config to be saved properly in yaml file

PR #1223, opened 2024-06-17
fix: find right config if same dbt test is added with diff configs in yml

saravmajestic:
### Problem
If a column has same type of generic test with diff config, doc editor does not save properly

### Solution
- While finding the config of the test from existing schema, equate the values to get right config

### Screenshot/Demo

### How to test
- add test as in issue #1219 for any column in dbt project
- change some description of model from doc editor
- save
- should save with right config

## Checklist

### DIFF
--- a/src/services/dbtTestService.ts
+++ b/src/services/dbtTestService.ts
@@ -4,7 +4,11 @@ import {
   CreateDbtTestRequest,
   UserInputError,
 } from "../altimate";
-import { isColumnNameEqual, provideSingleton } from "../utils";
+import {
+  getColumnTestConfigFromYml,
+  isColumnNameEqual,
+  provideSingleton,
+} from "../utils";
 import { DocGenService } from "./docGenService";
 import { StreamingService } from "./streamingService";
 import { QueryManifestService } from "./queryManifestService";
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -256,3 +260,51 @@ export const getExternalProjectNamesFromDbtLoomConfig = (
   }
   return null;
 };
+
+export const isRelationship = (
```
saravmajestic, line 264: extracted these utilities to separate file from docEditPanel.ts

src/utils.ts
```diff
@@ -256,3 +260,51 @@ export const getExternalProjectNamesFromDbtLoomConfig = (
...
+
+    if (isAcceptedValues(kwargs)) {
+      return (
+        kwargs.values?.sort().toString() ===
+        t.accepted_values.values.sort().toString()
+      );
+    }
+
+    return true;
+  });
```
saravmajestic, line 309: this is the actual fix, which finds the config by values/field and to for generic tests

src/services/dbtTestService.ts
```diff
@@ -25,10 +29,10 @@ export class DbtTestService {
   ) {}
 
   private filterAndStringifyTest = (
-    tests: Record<string, Record<string, unknown>>[],
+    testsPerColumnOrModelFromYml: Record<string, Record<string, unknown>>[],
```
saravmajestic, line 32: renamed variables

src/services/dbtTestService.ts
```diff
@@ -41,25 +45,25 @@ export class DbtTestService {
     const { name, namespace } = test.test_metadata;
     const fullName = namespace ? `${namespace}.${name}` : name;
 
-    const selectedTest = tests.find(
-      (t: Record<string, Record<string, unknown>>) => {
-        return Boolean(t[fullName]);
-      },
+    const existingConfig = getColumnTestConfigFromYml(
```
saravmajestic, line 48: use the common utility method to get the right config

mdesmet (approved):

Issue #683, opened 2023-10-30
Multiple Query Result Tabs Feature


Duncannn:
### Describe the feature

A feature that allows users to have multiple query result tabs open simultaneously. When working on analyses, users often execute different queries and need to switch between the results. This feature would retain the query results in separate tabs, eliminating the need to re-execute the query or update the data visualization pane each time. It would be similar to the BigQuery UI, where a query result for any executed BigQuery query is stored and can be revisited without re-execution.

### Describe alternatives you've considered

Currently, when switching between different query results, users must re-execute the query and update the data visualization pane. Another alternative could be having a history or cache of the recent query results which can be revisited, but this still doesn't offer the same level of convenience and efficiency as multiple tabs.

### Who will benefit?

This feature will benefit engineers, data analysts, and any other users who need to run and compare multiple queries concurrently. It will streamline their workflow and save time due to the need for repeated executions.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

anandgupta42:
@Duncannn @rightx2 This feature was shipped in version 0.41.0.

PR #1231, opened 2024-06-19
feat: add query panel tab

gaurpulkit:
### Problem

https://github.com/AltimateAI/vscode-dbt-power-user/issues/683#event-13194579565

### Solution

Showing query panel in tabs

### Screenshot/Demo

Screenshot 2024-06-19 at 6 18 31 PM

### How to test

## Checklist


mdesmet (commented):
What happens if vscode is restarted?
I don't think we should allow to clear the results of result document.

src/webview_provider/QueryResultTab.ts
```diff
@@ -0,0 +1,128 @@
...
+      dbtProjectContainer,
+      altimateRequest,
+      telemetry,
+      emitterService,
+      dbtTerminal,
+      queryManifestService,
+      usersService,
+    );
+
+    const t = this;
```
mdesmet, line 42: What's this for?

webview_panels/src/modules/queryPanel/QueryPanel.tsx
```diff
@@ -30,8 +31,9 @@ const QueryPanel = (): JSX.Element => {
           />
         </Stack>
         <Stack>
-          <ShowOldUxButton />
+          {!isTab && <ShowOldUxButton />}
```
mdesmet, line 34:
```suggestion
          {isPanel && <ShowOldUxButton />}
```

mdesmet (commented):
LGTM I would not name it tab but `VirtualDocument` as in https://github.com/microsoft/vscode-extension-samples/blob/main/virtual-document-sample/README.md

src/webview_provider/newDocsGenPanel.ts
```diff
@@ -29,6 +29,7 @@ import { UsersService } from "../services/usersService";
 import { ConversationProvider } from "../comment_provider/conversationProvider";
 import { DbtDocsView } from "./DbtDocsView";
 import { ConversationService } from "../services/conversationService";
+import { QueryResultTab } from "./QueryResultTab";
```
saravmajestic, line 32: this can be part of queryResultPanel.ts. I have added DbtDocsView, because it was related to doc editor

webview_panels/src/modules/queryPanel/QueryPanel.tsx
```diff
@@ -30,8 +31,9 @@ const QueryPanel = (): JSX.Element => {
...
           <ClearResultsButton />
```
saravmajestic, line 35: clear results is also not needed in tab

webview_panels/src/modules/queryPanel/useQueryPanelListeners.ts
```diff
@@ -154,7 +158,24 @@ const useQueryPanelListeners = (): { loading: boolean } => {
     };
   }, [onMesssage]);
 
-  return { loading };
+  useEffect(() => {
+    if (!tabData) {
+      void executeRequestInSync("getQueryTabData", {}).then((data) => {
+        if (data) {
+          const typedData = data as QueryPanelStateProps;
+          setTabData(data as Record<string, unknown>);
```
saravmajestic, line 166: why is this typecasted to unknown? can you use `QueryPanelStateProps`?

webview_panels/src/modules/queryPanel/useQueryPanelListeners.ts
```diff
@@ -154,7 +158,24 @@ const useQueryPanelListeners = (): { loading: boolean } => {
     };
   }, [onMesssage]);
 
-  return { loading };
+  useEffect(() => {
+    if (!tabData) {
```
saravmajestic, line 162: Looks like you can avoid storing all the data in tabData. This is only needed to store whether this is a tab or not. You can just store this state value as boolean. This will avoid storing duplicate data in memory

src/webview_provider/QueryResultTab.ts
```diff
@@ -0,0 +1,128 @@
...
+import { getFormattedDateTime, provideSingleton } from "../utils";
+import {
+  AltimateWebviewProvider,
+  HandleCommandProps,
+  SharedStateEventEmitterProps,
+} from "./altimateWebviewProvider";
+import { UsersService } from "../services/usersService";
+
+@provideSingleton(QueryResultTab)
+export class QueryResultTab extends AltimateWebviewProvider {
```
saravmajestic, line 17: this new class can be avoided and can be combined with existing queryResultsPanel.

saravmajestic (changes requested):

anandgupta42:
@gaurpulkit make sure we add some tests for this as well. Also, please make sure to test it out manually in both native desktop environment and also in codespaces

gaurpulkit:
> What happens if vscode is restarted? I don't think we should allow to clear the results of result document.

It clears the tabs for now

src/webview_provider/queryResultPanel.ts
```diff
@@ -107,8 +109,10 @@ export class QueryResultPanel extends AltimateWebviewProvider {
   public static readonly viewType = "dbtPowerUser.PreviewResults";
   protected viewPath = "/query-panel";
   protected panelDescription = "Query results panel";
+  private _queryTabData: any;
 
   protected _panel: WebviewView | undefined;
+  protected _tabPanel: WebviewPanel | undefined = undefined;
```
saravmajestic, line 115: this can be private

src/webview_provider/queryResultPanel.ts
```diff
@@ -172,12 +176,49 @@ export class QueryResultPanel extends AltimateWebviewProvider {
...
+    this._webview = webview.webview;
+    this.renderTabWebviewView();
+  }
+
+  protected renderTabWebviewView() {
+    if (!this._webview) {
+      return;
+    }
+
+    this._webview.onDidReceiveMessage(this.handleCommand, this, []);
```
saravmajestic, line 214: use `setupWebviewHooks`

webview_panels/src/uiCore/components/gradientButton/GradientButton.tsx
```diff
@@ -0,0 +1,12 @@
+import { Button as ReactStrapButton, ButtonProps } from "reactstrap";
+import Tooltip from "../tooltip/Tooltip";
+import classes from "./styles.module.scss";
+
+export const GradientButton = (props: ButtonProps): JSX.Element => (
```
saravmajestic, line 5: add a sample in webview_panels\src\uiCore\uiToolkitStories\Components.stories.tsx

webview_panels/src/modules/queryPanel/QueryPanel.tsx
```diff
@@ -30,8 +31,9 @@ const QueryPanel = (): JSX.Element => {
           />
         </Stack>
         <Stack>
-          <ShowOldUxButton />
-          <ClearResultsButton />
+          {isPanel && <ShowOldUxButton />}
+          {isPanel && <ClearResultsButton />}
+          {isPanel && <ShowInTabButton />}
           <HelpButton />
```
saravmajestic, line 37:
```suggestion
          {isPanel && (
            <>
              <ShowOldUxButton />
              <ClearResultsButton />
              <ShowInTabButton />
            </>
          )}
          <HelpButton />
```

webview_panels/src/modules/queryPanel/useQueryPanelListeners.ts
```diff
@@ -15,13 +15,17 @@ import {
...
+  executeRequestInSync,
+} from "@modules/app/requestExecutor";
 import { HINTS, HINT_VISIBILITY_DELAY } from "./constants";
 import { QueryPanelStateProps } from "./context/types";
 
-const useQueryPanelListeners = (): { loading: boolean } => {
+const useQueryPanelListeners = (): { loading: boolean; isPanel: boolean } => {
   const dispatch = useQueryPanelDispatch();
   const { loading, lastHintTimestamp, hintIndex } = useQueryPanelState();
+  const [isPanel, setIsPanel] = useState(true);
```
saravmajestic, line 28: this state var name and useQueryPanelListeners argument are conflicting. Do you know in run time, which variable will be used? and if you want useQueryPanelListeners argument to override the state variable, update the state variable using useEffect

saravmajestic (changes requested):

src/webview_provider/queryResultPanel.ts
```diff
@@ -312,6 +354,32 @@ export class QueryResultPanel extends AltimateWebviewProvider {
     this._panel!.onDidChangeVisibility(sendQueryPanelViewEvent);
   }
 
+  private setupTabWebviewHooks() {
+    this._tabPanel!.webview.onDidReceiveMessage(
```
saravmajestic, line 358: it will be better to reuse the existing onDidReceiveMessage handler, otherwise this tab will lose the other command handling. You should be using existing `setupWebviewHooks` so that all the commands will go through same flow. Otherwise it will be difficult to maintain multiple versions of the query results panel

src/webview_provider/queryResultPanel.ts
```diff
@@ -172,12 +177,49 @@ export class QueryResultPanel extends AltimateWebviewProvider {
...
+  }
+
+  protected renderTabWebviewView() {
+    if (!this._webview) {
+      return;
+    }
+
+    this.setupTabWebviewHooks();
+
+    this._webview.html = this.getHtml(
```
saravmajestic, line 217: It is better to use `renderWebviewView`, which has few conditions to render different type of views based on configuration.

src/webview_provider/queryResultPanel.ts
```diff
@@ -312,6 +354,32 @@ export class QueryResultPanel extends AltimateWebviewProvider {
...
+              syncRequestId: message.syncRequestId,
+            });
+          default:
+            super.handleCommand(message);
+        }
+      },
+      this,
+      this._disposables,
+    );
+    const sendQueryTabViewEvent = () => {
```
saravmajestic, line 374: extract this as separate private function to avoid multiple bindings

saravmajestic (changes requested):

src/webview_provider/queryResultPanel.ts
gaurpulkit: I already tried to do that but the webview of tab isn't available when only panel opens. So, we need to set it when it opens.

src/webview_provider/queryResultPanel.ts
gaurpulkit: Again, the setup time of tab and panel will be different

saravmajestic (approved):

Issue #810, opened 2024-01-12
Run arbitrary SQL


bramrodenburg:
### Describe the feature

It would be nice if it were possible to directly run (dbt) SQL from a draft (an unsaved file) in VSCode. I often find myself wanting to quickly execute a piece of SQL to check a few things in the data. To do this, I either have to save the file first (which I don't want to) and then run it with dbt power user, or I have to use a different plugin (e.g. sqltools) but this gives me a completely different user experience.

### Describe alternatives you've considered

In VSCode the sqltools plugin, though this gives a different experience than the VSCode dbt Power User plugin.

### Who will benefit?

Everyone that wants to run SQL without saving it to a file.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

OneCyrus:
we would welcome this feature. there are quite some queries we would like to execute which don't result in a pure table.

thins like running explain/analyze queries or just some materializing queries. currently they all break because the extension adds the "limit" syntax which only works for pure queries.

it would be great if the result could be shown in a pure console style output when not returning table results (e.g. query plan)

Issue #578, opened 2023-09-12
Keep history of queries and their result sets


tdeenes:
### Describe the feature

Being able to run a SQL query and check the result set in a webview is very helpful during development or debugging. However, one usually wants to run multiple queries - it would be great to get a browsable history of queries and their result sets.

### Who will benefit?

Anyone who is already using the "Query Results" view.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

tdeenes:
resolved by https://github.com/AltimateAI/vscode-dbt-power-user/pull/1234

PR #1234, opened 2024-06-20
feat: query bookmarks

saravmajestic:
### Problem
- need ability to run arbitrary sql queries
- need to store reused queries

Resolves: https://github.com/AltimateAI/vscode-dbt-power-user/issues/810 https://github.com/AltimateAI/vscode-dbt-power-user/issues/578

### Solution
- added option to open new sql file using 1. `File -> New File` option and in query preview, added a button `+ new query` to run arbitrary queries
- introduced query history and bookmarks tab
- query history will show executed queries in current session
- queries can be bookmarked from history tab and also can be tagged
- bookmarked queries can be shared within organization
- queries can be re executed from history/bookmarks tab
- queries can be filtered by tags and searched by text

### Other changes
- removed legacy query results view

### Screenshot/Demo
https://www.loom.com/share/fa4801e3c632454887bf7bdee9f18f05
https://www.loom.com/share/9e180399c2284bca9d2a66142c0b6eeb

### How to test
- set config `dbt.enableQueryBookmarks` to true in vscode settings
- Execute a query
- should see history, bookmarks tab and + new query button

## Checklist


package.json
```diff
@@ -76,10 +76,10 @@
             "description": "Enable dbt docs collaboration.",
             "default": true
           },
-          "dbt.enableNewQueryPanel": {
```
saravmajestic, line 79: this flag is removed for legacy panel support

mdesmet (approved):

webview_panels/src/modules/queryPanel/useQueryPanelCommonActions.ts
```diff
@@ -0,0 +1,57 @@
...
+      privacy,
+      { tags: [], searchQuery: "" },
+    );
+
+    if (response) {
+      dispatch(setQueryBookmarks(response));
+    }
+  };
+
+  const refetchBookmarks = () => {
```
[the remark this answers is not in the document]
saravmajestic: currently no, but when they switch to bookmarks tab, it will be refreshed

src/webview_provider/queryResultPanel.ts
```diff
@@ -135,36 +162,65 @@ export class QueryResultPanel extends AltimateWebviewProvider {
...
+            }
           }
         },
         this,
         this._disposables,
       ),
     );
-    window.onDidChangeActiveColorTheme(
-      (e) => {
-        if (this._panel) {
```
AdiGajbhiye, line 151: does light/dark mode work without this?

src/commands/index.ts
```diff
@@ -534,6 +539,77 @@ export class VSCodeCommands implements Disposable {
...
+          const uri = Uri.parse(
+            `${project.projectRoot}/poweruser-${getFormattedDateTime()}.sql`,
+          ).with({ scheme: "untitled" });
+          const annotationDecoration: TextEditorDecorationType =
+            window.createTextEditorDecorationType({
+              rangeBehavior: DecorationRangeBehavior.OpenOpen,
+            });
+
+          const contentText =
+            "Enter your query here and execute it just like any dbt model file. This file is unsaved, you can either save it to your project or save it as a bookmark for later usage or share it with your team members.";
```
AdiGajbhiye, line 560: we probably should make this a sql comment.

src/commands/index.ts
```diff
@@ -534,6 +539,77 @@ export class VSCodeCommands implements Disposable {
...
+                setTimeout(() => {
+                  commands.executeCommand("cursorMove", {
+                    to: "up",
+                    by: "line",
+                    value: 1,
+                  });
+                }, 0);
+                const disposable = workspace.onDidChangeTextDocument((e) => {
+                  const editor = window.activeTextEditor;
+                  if (editor && e.document === editor.document) {
```
AdiGajbhiye, line 597: wont `e` always be activeTextEditor

src/webview_provider/altimateWebviewProvider.ts
```diff
@@ -104,7 +104,7 @@ export class AltimateWebviewProvider implements WebviewViewProvider {
     syncRequestId,
     ...rest
   }: SendMessageProps) {
-    this._webview?.postMessage({
+    this._panel?.webview?.postMessage({
```
AdiGajbhiye, line 107: i am assuming that `this._webview` is not getting used anymore

AdiGajbhiye (approved):
lgtm

src/webview_provider/queryResultPanel.ts
saravmajestic: yes, vscode adds vscode-dark/light class name to body tag when loading webview. we are using that in webview_panels for all pages. So this will not be needed anymore

src/commands/index.ts
saravmajestic: Initially we had it like that, but it adds to the saved query history. so we want to show as just one time text for every new editor

src/commands/index.ts
saravmajestic: Good catch. This is variable override and fixed it now

src/webview_provider/altimateWebviewProvider.ts
saravmajestic: it is used in some other places

PR #1244, opened 2024-06-26
fix: find error when modifying column names

saravmajestic:
### Problem
If schema.yml has entry only for model and no columns, then getting below error 
```
TypeError: Cannot read properties of undefined (reading 'find')
```

### Solution
use default array if columns are not added in schema.yml

### Screenshot/Demo

### How to test
- in schema.yml, create a model without columns
- sync with db from doc editor
- should not see any error

## Checklist

### DIFF
--- a/src/webview_provider/docsEditPanel.ts
+++ b/src/webview_provider/docsEditPanel.ts
@@ -444,9 +444,8 @@ export class DocsEditViewPanel implements WebviewViewProvider {
       return columns;
     }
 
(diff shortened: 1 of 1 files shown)


anandgupta42:
@saravmajestic Can you add a test for this?

saravmajestic:
> @saravmajestic Can you add a test for this?

yes, will add it

mdesmet (approved):

PR #1289, opened 2024-07-09
fix: lineage schema caching

AdiGajbhiye:
## Overview

- added caching for schema

what to test:
- when doing column lineage, doing it second time should be faster than first
- building and run model either through extension or terminal, should invalidate cache

### Problem

### Solution

### Screenshot/Demo

### How to test

## Checklist


src/webview_provider/newLineagePanel.ts
```diff
@@ -84,6 +86,19 @@ export class NewLineagePanel implements LineagePanelView {
     private terminal: DBTTerminal,
   ) {}
 
+  private getValueFromCache(key: string): ModelNode | undefined {
+    if (this.currSession !== env.sessionId) {
```
saravmajestic, line 90: since the `cache` is private, the data in cache will be lost once we reload/change workspace. do we need to handle it based on sessionid?

mdesmet (changes requested):
Following changes are required:

- Ensure cache entry is invalidated based on run_results.json. Create an event as there might be more use cases that are dependent on running results. The cache key should only be invalidated if the run was succesfull for that model
- Create a cache per project directory and use model id as cache key

src/manifest/dbtProject.ts
```diff
@@ -170,6 +173,40 @@ export class DBTProject implements Disposable {
...
   }
 
+  private createLastRunResultsWatcher() {
+    const watcher = workspace.createFileSystemWatcher(
+      new RelativePattern(this.getTargetPath()!, `run_results.json`),
+    );
+
+    watcher.onDidChange((e) => this.invalidateCacheUsingLastRun(e));
+    watcher.onDidCreate((e) => this.invalidateCacheUsingLastRun(e));
+    watcher.onDidDelete((e) => this.invalidateCacheUsingLastRun(e));
```
mdesmet, line 183: Do we need this one?

src/manifest/dbtProject.ts
```diff
@@ -170,6 +173,40 @@ export class DBTProject implements Disposable {
     );
   }
 
+  private createLastRunResultsWatcher() {
+    const watcher = workspace.createFileSystemWatcher(
+      new RelativePattern(this.getTargetPath()!, `run_results.json`),
+    );
+
+    watcher.onDidChange((e) => this.invalidateCacheUsingLastRun(e));
```
mdesmet, line 181: Is this event ever invoked? or is it always created?

src/manifest/dbtProject.ts
```diff
@@ -170,6 +173,40 @@ export class DBTProject implements Disposable {
...
+    try {
+      const runResults = JSON.parse(fileContent);
+      for (const n of runResults["results"]) {
+        if (n["unique_id"] in this.dbSchemaCache) {
+          delete this.dbSchemaCache[n["unique_id"]];
+        }
+      }
+    } catch (e) {
+      this.terminal.error(
+        "invalidCache",
```
mdesmet, line 202: I thought we were going to use events?

AdiGajbhiye: this seemed easier.

src/manifest/dbtProject.ts
AdiGajbhiye: yeah. i am assuming if the file already exists, it  considered as file change

src/manifest/dbtProject.ts
```diff
@@ -170,6 +173,39 @@ export class DBTProject implements Disposable {
...
+      new RelativePattern(this.getTargetPath()!, `run_results.json`),
+    );
+
+    watcher.onDidChange((e) => this.invalidateCacheUsingLastRun(e));
+    watcher.onDidCreate((e) => this.invalidateCacheUsingLastRun(e));
+    return watcher;
+  }
+
+  private async invalidateCacheUsingLastRun(file: Uri) {
+    const fileContent = (await workspace.fs.readFile(file)).toString();
```
mdesmet, line 187: It seems that almost everywhere in our code base we use `readFileSync(file, "utf8")`. Why not use the same?

mdesmet (commented):
LGTM, except file api usage. What is the difference between both methods?

anandgupta42:
@saravmajestic @AdiGajbhiye Please check if any tests can be added for this.

anandgupta42:
@AdiGajbhiye add a telemetry event for schema hit

mdesmet (approved):

PR #1341, opened 2024-07-29
feat: uiCore button text and icon hover effect changes

svipulrao98:
### Problem
new buttons UI

### Screenshot/Demo
https://www.loom.com/share/de5d984dc64d4d91bfee005b292e2952?sid=53ed6f38-e86c-4aa3-ba03-c2bc72afb5dd

### How to test

## Checklist


saravmajestic (changes requested):
https://www.loom.com/share/9279518c4acc47d29ddcec8534af5217?sid=72a0e4f7-c0f0-4883-8eea-0f5118fc918d

Few buttons are not showing properly. Make sure you verify all the places where the button component is used and shows as expected

svipulrao98:
> https://www.loom.com/share/9279518c4acc47d29ddcec8534af5217?sid=72a0e4f7-c0f0-4883-8eea-0f5118fc918d
> 
> Few buttons are not showing properly. Make sure you verify all the places where the button component is used and shows as expected

Done, made the changes for all the buttons

saravmajestic (changes requested):
https://www.loom.com/share/8da0f90f318a40fa9f4ef37c3c528ad5?sid=c658d7e0-6b5f-4ec1-a82e-504a5db01cb2
Still i can see lots of buttons broken.
When adding a new prop to a component, make sure you preserve the existing behaviour and your prop should be set only in places where it is needed. For ex: `showButtonText` should be true by default and if it not required, should be set false in places where it is not needed. 
Make sure you find all references where the button component is used and verify if all buttons are showing up properly

webview_panels/src/uiCore/components/button/Button.tsx
```diff
@@ -1,8 +1,31 @@
...
 import Tooltip from "../tooltip/Tooltip";
 
-export const Button = (props: ButtonProps): JSX.Element => (
-  <Tooltip title={props.title}>
-    <ReactStrapButton {...props} />
-  </Tooltip>
-);
\ No newline at end of file
+export const Button = (props: ButtonProps): JSX.Element => {
+  const defaultShowText = props?.showtext ? true : false;
```
saravmajestic, line 6: make sure the new props are defined explicitly. Other developers will not be able to find the props otherwise. They have to check the code to understand what is needed

webview_panels/src/uiCore/components/button/Button.tsx
```diff
@@ -1,8 +1,31 @@
...
+  const isIconPresent = props?.icon ? true : false;
```
saravmajestic, line 7: btw: `isIconPresent` and `defaultShowText` are boolean by default why do you need another line to set that again?

svipulrao98:
@saravmajestic https://www.loom.com/share/a3fd0c96d0e74022bcf79dcbf2d6032b?sid=20a57a59-3858-4f80-83ea-08df56562c6b

new icons ⬆️

webview_panels/src/modules/queryPanel/components/runAdhocQueryButton/RunAdhocQueryButton.tsx
```diff
@@ -8,9 +8,12 @@ const RunAdhocQueryButton = (): JSX.Element => {
   };
   return (
     <NewFeatureIndicator featureKey="run-adhoc-query-button-clicked">
-      <Button outline onClick={handleClick}>
-        + New query
-      </Button>
+      <Button
+        outline
+        onClick={handleClick}
+        icon={<>+</>}
```
saravmajestic, line 14: this should be icon and should be referenced from assets/icons/index.tsx

webview_panels/src/uiCore/components/button/Button.tsx
```diff
@@ -1,8 +1,42 @@
...
 
-export const Button = (props: ButtonProps): JSX.Element => (
-  <Tooltip title={props.title}>
-    <ReactStrapButton {...props} />
-  </Tooltip>
-);
\ No newline at end of file
+interface CustomButtonProps extends ButtonProps {
+  icon?: React.ReactNode;
+  showtext?: boolean;
```
saravmajestic, line 7: Change this component in this way:
- instead of buttonText props, it should be children. This will take care of all existing buttons
- if icon props is set, then it should show in new format, if not, old format
- no need for showtext props
- title is already part of buttonprops

saravmajestic (changes requested):

webview_panels/src/uiCore/components/button/Button.tsx
svipulrao98: noted

svipulrao98:
Closing this PR: re-opened in https://github.com/AltimateAI/vscode-dbt-power-user/pull/1367

PR #1377, opened 2024-08-16
feat: button with children as IDs

svipulrao98:
### Problem
Playwright is not able to detect as the button text is not visible(unless hovered)

### Solution
Add IDs to the buttons

### How to test

## Checklist


saravmajestic:
NOt needed anymore

PR #1384, opened 2024-08-28
Remove secrets from stacktrace and make less calls in background for cloud

mdesmet:
### Problem

### Solution

### Screenshot/Demo

### How to test

## Checklist

### DIFF
--- a/src/dbt_client/dbtCloudIntegration.ts
+++ b/src/dbt_client/dbtCloudIntegration.ts
@@ -236,11 +236,11 @@ export class DBTCloudProjectIntegration
     } else {
       this.initializePaths();
     }
-    this.findAdapterType();
+    if (!this.adapterType) {
+      this.findAdapterType();
+    }
     if (!this.version) {
       await this.findVersion();
-    } else {
-      this.findVersion();
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/telemetry/index.ts
```diff
@@ -96,16 +96,29 @@ export class TelemetryService implements vscode.Disposable {
...
             ? error.stack
             : JSON.stringify(error),
+        ),
         ...this.customAttributes,
       },
       measurements,
     );
   }
 
+  private removeGenericSecretsFromStackTrace(
```
anandgupta42, line 110: instead of removing, should we not redact and replace with "****" or some other marker so that we know secret was redacted

src/dbt_client/dbtCloudIntegration.ts
```diff
@@ -236,11 +236,11 @@ export class DBTCloudProjectIntegration
     } else {
       this.initializePaths();
     }
-    this.findAdapterType();
+    if (!this.adapterType) {
+      this.findAdapterType();
+    }
```
anandgupta42, line 241: nit: add a comment so that we know why this optimization check was put in place

PR #1390, opened 2024-08-30
feat: new feature message

svipulrao98:
### Problem
New Notebook feature

### Solution
Add a note in query panel

## Checklist

- [ ] Get the text verified from @ppradnesh / @mdesmet / @anandgupta42


webview_panels/src/modules/queryPanel/QueryPanelDefaultView.tsx
```diff
@@ -11,7 +11,11 @@ const QueryPanelDefaultView = (): JSX.Element => {
       <div>
         <div>
           <h2>New Features: Query History, Bookmarks, and More</h2>
```
saravmajestic, line 13: in this title also we need to add notebooks

PR #1475, opened 2024-10-21
feat: Enhance dbt-power-user Side Panel to Show Model Info for Non-SQL Files

toukoudo:
### Problem

* Our team creates dbt docs for each model, named after the model names.
   * e.g. We have a dbt docs file `a_model.md` for a dbt model `a_model`.
* The Side Panel of dbt-power-user only shows the model information when a SQL file is active.
* We want to see the model information even when a dbt docs file is active.

### Solution

* In `modelTreeviewProvider.ts`, `path.basename(fileName, ".sql")` is used to get a basename (=model name).
   * If an active file is not a SQL file, the extension is not removed, and the model cannot be fetched.
* I changed the code to use `path.parse(fileName).name` to get a basename. In this way, we can get a basename of any extension.

### Screenshot/Demo

Single model in a YAML file:
スクリーンショット 2024-11-10 18 19 00

Multiple mode in a YAML file. The model at the cursor is shown in a side panel:
スクリーンショット 2024-11-10 18 18 49

### How to test

1. Open dbt-power-user extension in a development host.
2. Open a dbt project folder.
3. Open a dbt docs file, and the Side Panel of dbt-power-user should show the model information.

## Checklist

- [x] I have run this code and it appears to resolve the stated issue.
- [ ] `README.md` updated and added information about my change.

----

> [!IMPORTANT]
> Enhance dbt-power-user extension to display model info for non-SQL files by updating model name extraction logic and adding YAML support.
> 
>   - **Behavior**:
>     - Updated `modelTreeviewProvider.ts` to use `lookupModelByEditorContent()` for model name extraction, supporting non-SQL files.
>     - Added event listener for text editor selection changes to refresh tree data in `ModelTreeviewProvider` and `DocumentationTreeviewProvider`.
>   - **Functions**:
>     - Added `lookupModelByEditorContent()` in `modelTreeviewProvider.ts` to determine model names from YAML content or file names.
>     - Added `getCurrentlySelectedModelNameInYamlConfig()` in `utils.ts` to extract model names from YAML files based on cursor position.
>   - **Misc**:
>     - Improved error handling in `getCurrentlySelectedModelNameInYamlConfig()` to log parsing errors.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=AltimateAI%2Fvscode-dbt-power-user&utm_source=github&utm_medium=referral)<sup> for 57db4e5afcc3e691706fdcba64b107220c2a3d45. It will automatically update as commits are pushed.</sup>

----

## Summary by CodeRabbit

- **New Features**
	- Enhanced model retrieval logic in the tree view, improving accuracy based on the active editor's content.
	- Introduced a new function to extract model names from the active YAML document.
	- Added responsiveness to tree data refresh when text editor selection changes.

- **Bug Fixes**
	- Improved error handling in the new model name retrieval function to ensure graceful failures.

- **Documentation**
	- Updated documentation to reflect changes in method signatures and new functionalities.


anandgupta42:
@toukoudo Thanks for your contribution. How will this work if there are multiple models defined within the same markdown file?

toukoudo:
Thanks for the comment! 

> How will this work if there are multiple models defined within the same markdown file?

Sadly, this PR doesn't handle multiple models in the same markdown file. 
I'll  be happy if the Side Panel works in the situation, but for now, it's set up to work with one model per file.

(If necessary, I would like to create another pull request to address that functionality in the future! )

mdesmet:
We should always aim at bringing something to production that works reliably. So we need to know what's the behaviour in case of multiple models. We could use the texteditor position in case of multiple models.

mdesmet:
Also when you say markdown, you mean yaml file right? Again it is better to support only yaml files if that is the ones that you tested. In case of markdown I'm not totally sure if this currently is associated with a model based on file basename. So let's only support yaml files for now.

Also the yaml file name is completely free and not linked to a model. You will need to get the model name from the yaml instead.

src/treeview_provider/modelTreeviewProvider.ts
```diff
@@ -104,10 +104,9 @@ abstract class ModelTreeviewProvider
     if (element) {
       return Promise.resolve(this.getTreeItems(element.key, event));
     }
-    const fileName = path.basename(
+    const fileName = path.parse(
       window.activeTextEditor!.document.fileName,
-      ".sql",
-    );
+    ).name;
```
mdesmet, line 109: We should adapt this code to follow this logic for sql files, but different logic for yaml files. to use the texteditor position in case of multiple models within the yaml file, or select the single model in case of a single model in the yaml file

mdesmet (changes requested):

toukoudo:
Thank you for the feedback!

 I'll work on implementing the code to extract the model name from the YAML file itself, rather than relying on the file name.

toukoudo:
I have updated the code to support multiple models in dbt docs.
(Demo screenshots are on the updated PR description.)

Please review again 🙇 @anandgupta42 @mdesmet

toukoudo:
just a friendly reminder > @anandgupta42 @mdesmet

Let me know if something should be changed!

anandgupta42:
@mdesmet  @saravmajestic  Can you please review?

src/treeview_provider/modelTreeviewProvider.ts
```diff
@@ -104,15 +108,28 @@ abstract class ModelTreeviewProvider
...
+
+    const fileName = path.parse(
+      window.activeTextEditor!.document.fileName,
+    ).name;
+    const model_by_file_name = event.nodeMetaMap.lookupByBaseName(fileName);
+    if (model_by_file_name) {
+      return Promise.resolve(
+        this.getTreeItems(model_by_file_name.uniqueId, event),
+      );
+    }
```
mdesmet, line 130: I think this code should have different handling for yaml files and sql files not in case the first part fails than go on to the second part. Only one lookup on the nodeMetaMap should be done.

src/utils.ts
```diff
@@ -359,3 +360,56 @@ export const getStringSizeInMb = (str: string): number => {
   const sizeInMB = sizeInBytes / (1024 * 1024);
   return sizeInMB;
 };
+
+export function getModelNameInActiveEditor(): string {
```
mdesmet, line 364: Rename method indicating that it only applies to YamlFiles. for example `getCurrentlySelectedModelNameInYamlConfig`

mdesmet (commented):
Looks getting there. I think if we can handle these remarks and succesfully test it we can release it soonish.

src/treeview_provider/modelTreeviewProvider.ts
toukoudo: I changed the code to use file content only if a file is YAML, and lookupByBaseName is called once.

Call me if my change is not your intention!

src/utils.ts
toukoudo: fixed

toukoudo:
Sorry for the delay! I believe I have addressed your comments, so please check!  > @mdesmet

anandgupta42:
@toukoudo we are testing it and will get back to you soon

src/treeview_provider/modelTreeviewProvider.ts
```diff
@@ -500,3 +515,16 @@ export class DocumentationTreeview extends DocumentationTreeviewProvider {
...
+// Find appropriate a model from file content (if YAML) or from a file name (otherwise)
+export function lookupModelByEditorContent(
+  nodeMetaMap: NodeMetaMap,
+  document: TextDocument,
+): NodeMetaData | undefined {
+  const modelCandidateName =
+    document.languageId === "yaml" &&
+    getCurrentlySelectedModelNameInYamlConfig()
+      ? getCurrentlySelectedModelNameInYamlConfig()
+      : path.parse(document.fileName).name;
```
mdesmet, line 528: This is not necessarily correct as the name might not be a valid model name or even just wrong while  parsing yaml should be correct, but I guess it is OK to have as fallback.

mdesmet (approved):

mdesmet:
Thanks a lot for your great contribution, @toukoudo !

PR #1573, opened 2025-02-06
fix: resolve to DBT_PROFILES_DIR if specified without joining with project_dir

### LINKED ISSUE #1554: Extension not picking up env variables from .env file
### Expected behavior

To use our dbt core we simply execute a source command that gets all the variables in our .env file of the workspace and it runs perfectly, the extension should use them too as described [here](https://docs.myaltimate.com/setup/optConfig/#Environment%20variables%20through%C2%A0python.envFile) in the docs of the extension.

### Actual behavior

The extension doesn't pick up the variables in .env file and keeps returning the same error (image bellow).
Those env variables are necessary to find the profiles yml of our dbt project.

Image

### Steps To Reproduce

- Create a .env file in your project.
- Create an env variable called: `DBT_PROFILES_DIR`
- Open the project in VSCode
- Extension doesn't work.

Bellow is my .env file as a .txt

[env-example.txt](https://github.com/user-attachments/files/18481274/env-example.txt)

 for example.

### Log output/Screenshots

_No response_

### Operating System

WSL for Windows

### dbt version

1.7.18

### dbt Adapter

BigQuery

### dbt Power User version

0.51.2

diegoquintanav:
### Problem

see https://github.com/AltimateAI/vscode-dbt-power-user/issues/1554#issuecomment-2638276756

Closes #1554

### Solution

Removing the unnecessary path join

### Screenshot/Demo

### How to test

## Checklist

### DIFF
--- a/dbt_core_integration.py
+++ b/dbt_core_integration.py
@@ -217,7 +217,7 @@ def default_profiles_dir(project_dir):
         profiles_dir = os.path.expanduser(os.environ["DBT_PROFILES_DIR"])
         if os.path.isabs(profiles_dir):
             return os.path.normpath(profiles_dir)
-        return os.path.normpath(os.path.join(project_dir, profiles_dir))
+        return os.path.normpath(os.path.join(profiles_dir))
     project_profiles_file = os.path.normpath(os.path.join(project_dir, "profiles.yml"))
     return (
         project_dir


mdesmet:
This change will break things for other people.For the extension the CWD is not obvously set, compared to CLI where there is clear concept of CWD. Becasue of that we went with choosing the project directory as the CWD.

anandgupta42:
@diegoquintanav I am not sure how this resolves the original bug #1554 . Can you help explain?

diegoquintanav:
> This change will break things for other people.For the extension the CWD is not obvously set, compared to CLI where there is clear concept of CWD. Becasue of that we went with choosing the project directory as the CWD.


Hey thanks for checking this PR!

This change affects only the scenario where DBT_PROFILES_DIR is set and is not an absolute path. I've pasted my original reply to the issue in the PR description if that helps.

In what cases would you say it breaks behaviour for other people?

diegoquintanav:
> In the code that is removed in this PR the path is relative to the project root. In your case it is to be defined, by the calling code (probably the first workspace root directory). All people that have setup a relative path from the project root would have an issue if we merge this.

thanks for the follow-up @mdesmet, I'm not sure if I understand the scenario you present. 

My assumptions are that
* vscode is run from `/home/me/myproject`
* my dbt project is in `/home/me/myproject/mydbtproject` with a profiles file in `/home/me/myproject/mydbtproject/profiles/profiles.yml`
* environment variable `DBT_PROJECT_DIR` is set to `./mydbtproject`
* environment variable `DBT_PROFILES_DIR` is set to `./mydbtproject/profiles`

In this scenario, what would you say is the value of `project_dir` in <https://github.com/AltimateAI/vscode-dbt-power-user/blob/5735fa6e8f1ce3cbc2bb2b2d6f4a062baa2bf254/dbt_core_integration.py#L203C26-L203C37>?

mdesmet:
> * vscode is run from `/home/me/myproject`
> * my dbt project is in `/home/me/myproject/mydbtproject` with a profiles file in `/home/me/myproject/mydbtproject/profiles/profiles.yml`

`/home/me/myproject/mydbtproject` will be project_dir


> * environment variable `DBT_PROJECT_DIR` is set to `./mydbtproject`

We don't use this environment variable as it is not compatible with multi dbt project projects, but we can rethink that one.

> * environment variable `DBT_PROFILES_DIR` is set to `./mydbtproject/profiles`
> 
> In this scenario, what would you say is the value of `project_dir` in [`5735fa6`/dbt_core_integration.py#L203C26-L203C37](https://github.com/AltimateAI/vscode-dbt-power-user/blob/5735fa6e8f1ce3cbc2bb2b2d6f4a062baa2bf254/dbt_core_integration.py#L203C26-L203C37)?

`/home/me/myproject/mydbtproject/mydbtproject/profiles`

diegoquintanav:
If I set `DBT_PROFILES_DIR` to `./profiles/` then that breaks the native `dbt` behavior. In other words, commands like

```bash
DBT_PROFILES_DIR=./profiles/ DBT_PROJECT_DIR=./mydbtproject/ dbt ls -m foo
``` 

Would break (there's no file at `./profiles/profiles.yml`). If the extension is using `dbt` reserved environment variables, then it should replicate the `dbt` behavior. 

> We don't use this environment variable as it is not compatible with multi dbt project projects, but we can rethink that one.

I think this is key. What is a multi-dbt project? does `dbt` support one natively through environment variables? 

a hacky solution can be done, if the extension decides to use native `dbt`environment variables _and_ support multi project environments, by doing in <https://github.com/AltimateAI/vscode-dbt-power-user/blob/dc24d1e269a70bf44349346bf657a86e1af5bf5d/dbt_core_integration.py#L242C29-L242C40>

```python
    if "DBT_PROFILES_DIR" in os.environ:
        # case when DBT_PROFILES_DIR=~/something
        profiles_dir = os.path.expanduser(os.environ["DBT_PROFILES_DIR"])
        if os.path.isabs(profiles_dir):
            # if path is absolute, then ok, go ahead
            return os.path.normpath(profiles_dir)
        if "DBT_PROJECT_DIR" in os.environ:  # new
            # assume DBT_PROFILES_DIR is relative to the same directory that DBT_PROJECT_DIR was set to
            # return the profiles_dir path as it is 
            return os.path.normpath(profiles_dir)) # new
        return os.path.normpath(os.path.join(project_dir, profiles_dir))
``` 

a different approach perhaps would be that if `DBT_PROJECT_DIR` is set, assume we are working only with one project and in 

https://github.com/AltimateAI/vscode-dbt-power-user/blob/5735fa6e8f1ce3cbc2bb2b2d6f4a062baa2bf254/src/manifest/dbtWorkspaceFolder.ts#L90

resolve the argument passed to `findFiles` as `DBT_PROJECT_DIR` if present.

anandgupta42:
@mdesmet @saravmajestic Can you please review the code?

mdesmet:
> Thanks @mdesmet for your input. If I set `DBT_PROFILES_DIR` to `./profiles/` then that breaks the native `dbt` behavior. In other words, commands like
> 
> ```shell
> DBT_PROFILES_DIR=./profiles/ DBT_PROJECT_DIR=./mydbtproject/ dbt ls -m foo
> ```
> 
> Would break (there's no file at `./profiles/profiles.yml`). If the extension is using `dbt` reserved environment variables, then it should replicate the `dbt` behavior.
> 
> > We don't use this environment variable as it is not compatible with multi dbt project projects, but we can rethink that one.
> 
> I think this is key. What is a multi-dbt project? does `dbt` support one natively through environment variables?
> 
> a hacky solution can be done, if the extension decides to use native `dbt`environment variables _and_ support multi project environments, by doing in [`dc24d1e`/dbt_core_integration.py#L242C29-L242C40](https://github.com/AltimateAI/vscode-dbt-power-user/blob/dc24d1e269a70bf44349346bf657a86e1af5bf5d/dbt_core_integration.py#L242C29-L242C40)
> 
> ```python
>     if "DBT_PROFILES_DIR" in os.environ:
>         # case when DBT_PROFILES_DIR=~/something
>         profiles_dir = os.path.expanduser(os.environ["DBT_PROFILES_DIR"])
>         if os.path.isabs(profiles_dir):
>             # if path is absolute, then ok, go ahead
>             return os.path.normpath(profiles_dir)
>         if "DBT_PROJECT_DIR" in os.environ:  # new
>             # assume DBT_PROFILES_DIR is relative to the same directory that DBT_PROJECT_DIR was set to
>             # return the profiles_dir path as it is 
>             return os.path.normpath(profiles_dir)) # new
>         return os.path.normpath(os.path.join(project_dir, profiles_dir))
> ```
> 

This can work but I would add a check if the path exists and then only go in this path, this will act as an additional safeguard. 

> a different approach perhaps would be that if `DBT_PROJECT_DIR` is set, assume we are working only with one project and in
> 
> https://github.com/AltimateAI/vscode-dbt-power-user/blob/5735fa6e8f1ce3cbc2bb2b2d6f4a062baa2bf254/src/manifest/dbtWorkspaceFolder.ts#L87-L90
> 
> resolve the argument passed to `findFiles` as `DBT_PROJECT_DIR` if present.

This won't resolve the issue, as still the profiles dir will be evaluated against the project dir.

PR #1593, opened 2025-03-04
feat: allow to specify dbt projects directly and skip discovery

OneCyrus:
## Overview

this allows to skip the project discovery as it often fails to discover projects

### Problem

the extension often didn't discover our projects and returned an empty list.

### Solution

allow to specify the dbt projects manually through a setting `dbt.specifiedProjects`

### How to test

- Add dbt project paths to  `dbt.specifiedProjects` in settings.json
- Restart extension and it should pickup the project without using discovery

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -283,6 +283,14 @@
                 }
               }
             }
+          },
+          "dbt.specifiedProjects": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "default": [],
+            "description": "Specify dbt project paths to override the discovered projects."
           }
         }
       }
--- a/src/dbt_client/dbtCloudIntegration.ts
+++ b/src/dbt_client/dbtCloudIntegration.ts
@@ -150,6 +150,14 @@ export class DBTCloudProjectDetection
 
   async discoverProjects(projectDirectories: Uri[]): Promise<Uri[]> {
     this.altimate.handlePreviewFeatures();
+    const specifiedProjects = workspace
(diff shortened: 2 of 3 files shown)


anandgupta42:
@OneCyrus The discovery system is designed to automatically identify projects. I was curious if you investigated why some projects aren't being discovered. I believe this could help us address the root cause of the problem without needing to pass any direct paths.

mdesmet:
Thanks @OneCyrus, have you tried to run the extension in debug mode and check why the projects are not detected in the first place?

OneCyrus:
the issue is that the workspace.findFiles isn't somehow reliable. about 1 in 5 times it finds the files and the other 4 times it returns nothing. also our workspace is quite large with multiple dbt projects. maybe it just runs into a timeout randomly. unfortunately there is no error in the debug logs.

https://github.com/OneCyrus/vscode-dbt-power-user/blob/8133da53b895fc107e41ec1c4bbc4de8e5e93ff8/src/manifest/dbtWorkspaceFolder.ts#L87-L93

mdesmet:
@OneCyrus : Can you make the changes to retry the project detection? Also can you link this issue to the vscode issue that you found where findInFiles is not reliable.

OneCyrus:
@mdesmet i made a new PR with the retry logic instead of the custom project config

https://github.com/AltimateAI/vscode-dbt-power-user/pull/1615

PR #1598, opened 2025-03-09
Allow toolbar buttons visibility to be configured via settings

matthieucan:
### Problem

Hi! Thanks for the great extension.

I would like the toolbar buttons to be configurable so they can be hidden. I find the long list (7 buttons) prone to misclicks, and there are some that I don't use as they are not relevant for my projects.

### Solution

Make every button conditionally appear, via an associated setting. By default, they are all present, so this does not alter the current behavior.

### Screenshot/Demo

A simplified toolbar with only the "Run" and "Test" buttons enabled:
Screenshot 2025-03-09 at 21 56 14

### How to test

- Run this branch
- Go to Settings, and uncheck the options "Enable the 'xxx' button in SQL files"
- Open an SQL file, look at the toolbar

## Checklist

----

> [!IMPORTANT]
> Toolbar buttons in SQL files are now configurable via new boolean settings in `package.json`, allowing users to hide unused buttons.
> 
>   - **Behavior**:
>     - Toolbar buttons in SQL files can now be conditionally hidden via new boolean settings in `package.json`.
>     - Default behavior remains unchanged; all buttons are visible unless settings are modified.
>   - **Settings**:
>     - Added settings: `dbt.enableBuildModel`, `dbt.enableExecuteSql`, `dbt.enableSqlQuickPick`, `dbt.enableRunModel`, `dbt.enableTestModel`, `dbt.enableSqlPreview`, `dbt.enableConvertToModel`.
>   - **Commands**:
>     - Updated `editor/title` and `editor/title/run` commands to check new settings before displaying buttons.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=AltimateAI%2Fvscode-dbt-power-user&utm_source=github&utm_medium=referral)<sup> for 253ae83acc353bf294dc0204a6c4dec978510c3c. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

- **New Features**
  - Added seven user-configurable toggles to control visibility of dbt-related action buttons in SQL files (build, execute, run, test, preview, quick-pick, convert).
  - Menu actions now appear only for SQL/Jinja-SQL files and when the corresponding toggle is enabled.

- **Documentation**
  - Updated configuration docs with a new UI/Display Settings section describing the seven toggles and their effects.


mdesmet:
This is OOTB supported by vscode. I don't have big objection but these things are sometimes easy to overlook when maintaining the extension. So I prefer to not support any features that are already supported by vscode.

image

matthieucan:
> This is OOTB supported by vscode. I don't have big objection but these things are sometimes easy to overlook when maintaining the extension. So I prefer to not support any features that are already supported by vscode.

You're right! However, I think this can't be configured through `.vscode/settings.json` in a given project?

package.json
```diff
@@ -922,32 +957,32 @@
...
         {
           "command": "dbtPowerUser.buildChildrenParentModels",
-          "when": "resourceLangId =~ /^sql$|^jinja-sql$/",
+          "when": "resourceLangId =~ /^sql$|^jinja-sql$/ && config.dbt.enableBuildModel",
           "group": "inline"
         },
         {
           "command": "dbtPowerUser.runCurrentModel",
-          "when": "resourceLangId =~ /^sql$|^jinja-sql$/",
+          "when": "resourceLangId =~ /^sql$|^jinja-sql$/ && config.dbt.enableBuildModel",
```
saravmajestic, line 985: this should be enableRunModel? @matthieucan

matthieucan: That one is trickier than the others: all the actions within the single button "Build project" are behind a single flag.
This allows to display/hide the entire group (the button with the down arrow), rather than sub-items in this group. Only hiding a subset of the sub-items within this button would not hide the button itself, therefore it would be less useful IMO.

Does that make sense?

anandgupta42:
@mdesmet @saravmajestic We will have to update the docs also - https://docs.myaltimate.com/setup/configuration/

matthieucan:
> @mdesmet @saravmajestic We will have to update the docs also - https://docs.myaltimate.com/setup/configuration/

Apologies for the (long) delay in getting back to you. I have now updated the documentation, thanks for the pointer!

dev-punia-altimate:
## ✅ Tests — All Passed



cc @matthieucan

PR #1612, opened 2025-03-18
feat: add command to run apply defer config

grindheim:
### Problem

To simplify the user experience, we have set up a codespace configuration where the `dbt.deferConfigPerProject` is preconfigured, and we have a task that automatically downloads the manifest from project.

However, the user still needs manually go to Actions and toggle the "Enable defer to production" off and on again in order to activate the deferral with the downloaded manifest.

We'd like to be able set up the environment including deferral end to end without any user actions, and in order to do this, the dbt Power User extension needs to provide a command for applying defer config, much like what happens when toggling the button, so that this task can be triggered by us as soon as the updated manifest has been downloaded.

E.g.
```
        "dbt.deferConfigPerProject": {
          "": {
              "deferToProduction": true,
              "manifestPathForDeferral": "/workspaces/<rep>/state",
              "manifestPathType": "local",
              "favorState": false
          }
```

### Solution

The suggested solution adds a new command to `package.json`, and also registers the command in `src\commands\index.ts`.

### Current issues

As there were multiple `applyDeferConfig` functions configured in the project, I was unsure which one to use. I settled on one defined in DBTProject. However, the issue is here how to trigger this function.
The current implementation seems to not be able to fetch the current project, and can thus not run `applyDeferConfig`.
This seems to be the error message behind the scenes: `NameError: name 'project' is not defined`.

I'd appreciate some input on which changes I should implement to make it work as it should

### Screenshot/Demo

### How to test

## Checklist

----

> [!IMPORTANT]
> Add `dbtPowerUser.applyDeferConfig` command to apply defer configuration automatically, but it currently fails due to project fetching issues.
> 
>   - **Behavior**:
>     - Adds `dbtPowerUser.applyDeferConfig` command in `package.json` to apply defer configuration automatically.
>     - Registers command in `src/commands/index.ts` to apply defer configuration to all projects using `project.applyDeferConfig()`.
>   - **Issues**:
>     - Current implementation cannot fetch the current project, causing `applyDeferConfig` to fail with `NameError: name 'project' is not defined`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=AltimateAI%2Fvscode-dbt-power-user&utm_source=github&utm_medium=referral)<sup> for cdb82e111dc1feb57cca1513b5e7572dcf1a06f6. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

- **New Features**
  - Introduced a new command, "Apply defer configuration," which enables you to apply deferred configuration settings across multiple projects directly within the extension.
  - Enhanced error handling and success messaging for the new command to improve user experience.


src/commands/index.ts
```diff
@@ -797,6 +797,17 @@ export class VSCodeCommands implements Disposable {
           );
         },
       ),
+      commands.registerCommand("dbtPowerUser.applyDeferConfig", () => {
+        const pickedProject: ProjectQuickPickItem | undefined =
+          this.dbtProjectContainer.getFromWorkspaceState(
+            "dbtPowerUser.projectSelected",
+          );
```
mdesmet, line 804: This would only work if people have gone through the walkthrough.

Either it works based on the currently selected file, or if there is only a single project, pick that one, or if there are multiple projects pick one from a list of projects. You could use some of that code for inspiration.

mdesmet:
You could try using a tool like Cursor or aider to try develop this feature.

mdesmet:
Can you elaborate on this:

- However, the user still needs manually go to Actions and toggle the "Enable defer to production" off and on again in order to activate the deferral with the downloaded manifest.

why is the config not enough?

src/commands/index.ts
grindheim: Updated logic to (hopefully) work based on either having gone through the walkthrough, having a selected file or by looping projects:
```
commands.registerCommand("dbtPowerUser.applyDeferConfig", () => {
        // If user has completed walkthrough:
        const pickedProject: ProjectQuickPickItem | undefined =
          this.dbtProjectContainer.getFromWorkspaceState(
            "dbtPowerUser.projectSelected",
          );
        if (pickedProject && pickedProject.uri) {
          this.dbtProjectContainer
            .findDBTProject(pickedProject.uri)
            ?.applyDeferConfig();
            return;
        }
        // If active text editor
        if (window.activeTextEditor && window.activeTextEditor.document.uri) {
          const activeFileUri = window.activeTextEditor.document.uri;
          const dbtProject = this.dbtProjectContainer.findDBTProject(activeFileUri);
          if (dbtProject) {
            dbtProject.applyDeferConfig()
            return;
          }
        }
        // Loop projects
        const projects = this.dbtProjectContainer.getProjects();
        if (projects.length >= 0) {
          for (const project of projects) {
            project.applyDeferConfig()
          }
        }
      }
```

mdesmet:
Can you not add this logic to the run command of your docker container before invocation of vscode?

grindheim:
The download of the production manifest needs to be done following user authentication to GCP (using oauth), as it uses those credentials. And secondly, the user might have a long running codespace and would need to update the manifest from time to time.

But in any case, I don't see the downside of having this action available as a command.

grindheim:
Of course, the best thing would be if the extension simply always re-read the manifest at the given path, rather than storing it in cache somewhere, as it seems to be doing now. If we only had to update the file and then be sure that the updated file would be used, then all would be splendid.

src/commands/index.ts
```diff
@@ -797,6 +797,35 @@ export class VSCodeCommands implements Disposable {
...
+            return;
+          }
+        }
+        // Loop projects
+        const projects = this.dbtProjectContainer.getProjects();
+        if (projects.length >= 0) {
+          for (const project of projects) {
+            project.applyDeferConfig()
+          }
+        }
```
mdesmet, line 827:
```suggestion
        const projects = this.dbtProjectContainer.getProjects();
        await Promise.all(projects.map((project) => project.applyDeferConfig()));
```

mdesmet: Maybe just do this?

grindheim: Great suggestion, simplicity is king.

mdesmet (approved):

PR #1655, opened 2025-05-17
Handle zero-index hits in utils

anandgupta42:
## Summary
- fix `get_start_and_end_position` so zero index values are not treated as falsy
- add regression test for detecting invalid tokens at the beginning of SQL

## Testing
- `python -m unittest altimate_packages/tests/test_utils.py`
- `npm ci` *(fails: Exit handler never called)*

### DIFF
--- a/altimate_packages/tests/test_utils.py
+++ b/altimate_packages/tests/test_utils.py
@@ -1,0 +1,15 @@
+import sys
+import unittest
+sys.path.insert(0, 'altimate_packages')
+from altimate.utils import get_start_and_end_position
+
+class TestGetStartEndPosition(unittest.TestCase):
+    def test_invalid_token_at_beginning(self):
+        sql = "invalid_token SELECT * FROM table"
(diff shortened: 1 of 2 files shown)


mdesmet (commented):
I don't think we want to handle 0-index hits.

altimate_packages/tests/test_utils.py
```diff
@@ -0,0 +1,15 @@
...
+sys.path.insert(0, 'altimate_packages')
+from altimate.utils import get_start_and_end_position
+
+class TestGetStartEndPosition(unittest.TestCase):
+    def test_invalid_token_at_beginning(self):
+        sql = "invalid_token SELECT * FROM table"
+        start, end, count = get_start_and_end_position(sql, "invalid_token")
+        self.assertEqual(start, [0, 1])
+        self.assertEqual(end, [0, len("invalid_token") + 1])
+        self.assertEqual(count, 1)
```
mdesmet, line 12: Test case, doesn't test new behavior.

PR #1665, opened 2025-05-19
Fix profile change not updating target

anandgupta42:
## Summary
- refresh project configuration when profiles.yml changes

## Testing
- `npm test`
- `npm run lint`


mdesmet:
As profiles can be located outside of the workspace folders, the vscode file watcher won't work. We should use a node js file watcher instead.

Issue #1708, opened 2025-07-01
Not recognizing when model capitalization has changed


MLSteinw:
### Expected behavior

I have a model that after correcting the capitalization in the model name I can no longer use dbt-power-user to run any commands. it continues to use the model name prior to capitalization change. 

example:
original name: gold__checklistTaskFact
renamed to: gold__ChecklistTaskFact 

when I try to run or compile it will run in the terminal using gold__checklistTaskFact 

which results in dbt not finding a matching node. 

tried cleaning dbt, reloading extension, restarting vscode, renaming to completely different model name then renaming it back (renaming it to something with a new character worked but when renamed back it stopped working). 

### Actual behavior

when file capitalization in file name is changed this should be recognized and built in commands should run using the new capitalization 

### Steps To Reproduce

Create file. 
run/compile file using extension
rename file but only change capitalization, nothing else.
attempt to run/compile with extension.

### Log output/Screenshots

![Image](https://github.com/user-attachments/assets/04aaa41a-e0a3-4f0a-999e-6cff6786672a)

> Executing task: dbt compile --select gold__checklistTaskFact --project-dir c:\Users\example\Documents\AzureRepos\PH_dbt\ph_dbt --profiles-dir c:\Users\example\Documents\AzureRepos\PH_dbt\ph_dbt --target development

### Operating System

windows 

### dbt version

1.8.7

### dbt Adapter

mssql

### dbt Power User version

0.58.3

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

MLSteinw:
I was able to get it to recognize the file rename but I had to uninstall and reinstall dbt power user.

PR #1771, opened 2025-12-03
Handle file rename with casing only properly

stegus64:
## Overview

After renaming a file by only changing casing, right-clicking in an editor window and selecting run dbt model does not work.

### Problem

As described in #1708 

The problem is only visible on case insensitive file systems, like windows and macos.
Linux filesystems are normally case sensitive so they are not affected by this problem.

### Solution

Added a call to fs.realpathSync.native() to make sure that the model name passed to dbt on the command line has exactly the same casing as the filename on disk.

### How to test

- Make sure you are running on a case insensitive filesystem (windows or macos)
- Make sure you have a working dbt project open
- Rename a model file by only changing case (abc.sql -> ABC.sql)
- Right-click in the editor and select run dbt model
- Before this change this will cause an error because the modelname passed to dbt will be the old name (dbt run --select abc.sql)
- After this change the command should work properly (dbt run --select ABC.sql)

## Checklist

----

> [!IMPORTANT]
> Fixes issue with dbt model execution after file renaming with casing changes on case-insensitive filesystems by using `fs.realpathSync.native()` for correct model name casing.
> 
>   - **Behavior**:
>     - Fixes issue with running dbt models after renaming files with casing changes on case-insensitive filesystems (Windows, macOS).
>     - Uses `fs.realpathSync.native()` in `createModelParams()` in `dbtProjectContainer.ts` to ensure correct casing of model names.
>   - **Testing**:
>     - Tested on case-insensitive filesystems by renaming a model file and running the dbt model command.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=AltimateAI%2Fvscode-dbt-power-user&utm_source=github&utm_medium=referral)<sup> for 14300750e7baa96aa8c9217a9bd90e6be32fb99a. You can [customize](https://app.ellipsis.dev/AltimateAI/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model path resolution to correctly identify models in complex directory structures, including symlinked paths.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -357,7 +357,6 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.26.0.tgz",
       "integrity": "sha512-i1SLeK+DzNnQ3LL/CswPCa/E5u4lh1k6IAEphON8F+cXt0t9euTshDru0q7/IqMa1PMPz5RnHuHscF8/ZJsStg==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.26.0",
@@ -4180,7 +4179,8 @@
       "optional": true,
       "os": [
         "android"
-      ]
+      ],
+      "peer": true
     },
     "node_modules/@rollup/rollup-android-arm64": {
       "version": "4.24.4",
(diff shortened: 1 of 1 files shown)


mdesmet (commented):
LGTM but have you searched if there any other occurences in the code that may benefit from this change?

PR #1845, opened 2026-03-23
fix: compile preview ENOENT and race condition for untitled query files

shreyastelkar:
### Before (not project specific. It selected the first one):
image

### After: 
image

Fixes "Compiled dbt Preview" (Cmd+') failing for untitled query files in **multi-project workspaces**. Follow-up to #1842.

### Broken flow

Multi-project workspace → untitled file with project selected → Cmd+' → **ENOENT error** or "Still loading dbt project..."

`SqlPreviewContentProvider.requestCompilation()` assumes `file:` scheme when looking up the source document. For untitled files the scheme is `untitled:`, so the lookup fails, falls back to `readFileSync("Untitled-1")` → ENOENT. Project resolution also fails because `findDBTProject(Uri.file("Untitled-1"))` can't match in multi-project.

Single-project works on `main` because the CWD-relative path resolution happens to land inside the only workspace folder.

### Changes

- `requestCompilation()` finds the source document by matching URI path (scheme-agnostic) and passes `document.uri` to `findDBTProject` which handles untitled resolution internally (#1842)
- Live-update handler uses the same path-based matching so edits refresh the preview
- `compileQueryOnActiveWindow` and `executeQueryOnActiveWindow` capture editor state after async project picker to prevent stale references (defensive)

### DIFF
--- a/src/commands/runModel.ts
+++ b/src/commands/runModel.ts
@@ -50,13 +50,20 @@ export class RunModel {
     if (!window.activeTextEditor) {
       return;
     }
-    const fullPath = window.activeTextEditor.document.uri;
-    if (fullPath.scheme === "untitled") {
+    // For untitled files, ensure a project is selected before capturing
+    // editor state — the await may show a picker, during which the user
+    // could switch editors.
+    if (window.activeTextEditor.document.uri.scheme === "untitled") {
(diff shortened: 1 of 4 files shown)


dev-punia-altimate:
## ✅ Tests — All Passed



cc @shreyastelkar

PR #1877, opened 2026-04-10
fix: harden `resolveSettingsVariables` against Sentry findings on #1871

ralphstodomingo:
## Summary

Addresses two MEDIUM bugs Sentry flagged on #1871 three minutes after that PR merged, in `src/utils.ts::resolveSettingsVariables`. Both are silent data-corruption bugs — they return wrong strings instead of throwing, so users experience them as "my dbt command didn't get the vars I configured".

### Bug 1 — multi-placeholder iteration failure (`src/utils.ts`, L448–461)

The prior implementation used `regexEnv.exec()` in a `while` loop while simultaneously mutating `value` with `value.replace()` inside the loop body. The regex's `lastIndex` is not rewound after the string shrinks, so any string containing more than one `${env:VAR}` placeholder had every placeholder after the first **silently skipped**.

**Repro in Docker E2E container** (`TEST_VAR1=alpha`, `TEST_VAR2=beta`):
```
Input:  "a=${env:TEST_VAR1},b=${env:TEST_VAR2}"
OLD:    "a=alpha,b=${env:TEST_VAR2}"   ← second placeholder lost
NEW:    "a=alpha,b=beta"                ← fixed
```

### Bug 2 — `$` interpretation in replacement string (`src/utils.ts`, L456–459)

Passing the raw env value as the second argument to `String.replace()` causes sequences like `$&`, `$'`, `$\``, `$$` to be interpreted as [backreferences](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement). If an env value contained any of those, the substituted string was silently corrupted.

**Repro in Docker E2E container** (`WEIRD=/tmp/$&/backup`):
```
Input:  "path=${env:WEIRD}"
OLD:    "path=/tmp/${env:WEIRD}/backup"   ← $& expanded to the whole match
NEW:    "path=/tmp/$&/backup"              ← fixed
```

## Fix

Single callback-based `.replace()` pass. The callback:
1. Receives the original (non-mutating) string in one shot → no stateful-regex / mutating-string interaction
2. Returns the env value as a literal string → no backreference interpretation

Unresolved placeholders (env var not set) are left in place — matches prior behavior.

The same callback form is applied to `${workspaceFolder}` for the same `$`-interpretation reason. Windows paths can legitimately contain `$` characters (hidden admin shares), so the prior raw-string substitution had the same latent bug there.

## Tests

Adds `src/test/suite/resolveSettingsVariables.test.ts` with 9 regression cases:

- ✅ Bug 1 guard: multiple `${env:VAR}` placeholders in one string
- ✅ Bug 2 guard: env value contains `$` characters
- ✅ Single placeholder resolution
- ✅ Repeated occurrences of same var
- ✅ Unresolved placeholders preserved
- ✅ Workspace folder with `$` characters
- ✅ Mixed `${env:VAR}` + `${workspaceFolder}`
- ✅ Empty string handling
- ✅ Basic workspace folder resolution

All 9 pass locally.

## Docker E2E

Parity with #1871's Docker E2E testing. Code-server container recreated with test env vars (`TEST_VAR1=alpha`, `TEST_VAR2=beta`, `WEIRD=/tmp/$&/backup`, `REPEATED=x`). `.vscode/settings.json` in the jaffle-shop-duckdb test project configured with `dbt.runModelCommandAdditionalParams` exercising all three scenarios (multi-placeholder, `$`-char value, repeated same var). Standalone verification script run inside the container confirmed:

- **OLD code**: 1/6 params corrupted (multi-placeholder case), Bug 2 reproduced separately via `$&` env value
- **NEW code**: 6/6 params resolved correctly, both bug scenarios preserved as literal

## Why follow-up instead of amending #1871

#1871 was already merged and in `v0.60.3` / `v0.60.4` releases on master. Sentry's review landed 3 minutes after the merge. A new PR is the clean path.

## Sentry review references

- Bug 1: https://github.com/AltimateAI/vscode-dbt-power-user/pull/1871#discussion_r_ref_12970438_0 (stateful regex iteration)
- Bug 2: https://github.com/AltimateAI/vscode-dbt-power-user/pull/1871#discussion_r_ref_12970438_1 (`$` interpretation)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed placeholder resolution to correctly handle `$` characters in environment variable values and workspace folder paths as literal content.

* **Tests**
  * Added comprehensive test suite for placeholder resolution functionality.

### DIFF
--- a/src/test/suite/resolveSettingsVariables.test.ts
+++ b/src/test/suite/resolveSettingsVariables.test.ts
@@ -1,0 +1,86 @@
+import { afterEach, beforeEach, describe, expect, it } from "@jest/globals";
+import type { Uri } from "vscode";
+import { resolveSettingsVariables } from "../../utils";
+
+// The vscode mock's `Uri` re-imports from "vscode" recursively, so construct
+// a minimal shape instead. `resolveSettingsVariables` only reads `.fsPath`.
+const mockUri = (fsPath: string): Uri => ({ fsPath }) as unknown as Uri;
+
+describe("resolveSettingsVariables", () => {
+  const ORIGINAL_ENV = process.env;
+
+  beforeEach(() => {
+    process.env = { ...ORIGINAL_ENV };
+  });
+
+  afterEach(() => {
+    process.env = ORIGINAL_ENV;
+  });
+
+  it("resolves a single ${env:VAR} placeholder", () => {
+    process.env.FOO = "hello";
+    expect(resolveSettingsVariables("--vars=${env:FOO}")).toBe("--vars=hello");
+  });
+
+  it("resolves MULTIPLE ${env:VAR} placeholders in one string", () => {
+    // Regression: the previous implementation used a stateful global regex
+    // with a mutating string, so the second placeholder was skipped.
+    process.env.FOO = "alpha";
+    process.env.BAR = "beta";
+    expect(
+      resolveSettingsVariables("--vars='a=${env:FOO},b=${env:BAR}'"),
+    ).toBe("--vars='a=alpha,b=beta'");
+  });
+
+  it("resolves repeated occurrences of the same ${env:VAR}", () => {
+    process.env.FOO = "x";
+    expect(resolveSettingsVariables("${env:FOO}/${env:FOO}")).toBe("x/x");
+  });
+
+  it("leaves unresolved ${env:VAR} placeholders untouched", () => {
+    delete process.env.NOT_SET;
+    expect(resolveSettingsVariables("before ${env:NOT_SET} after")).toBe(
+      "before ${env:NOT_SET} after",
+    );
+  });
+
+  it("treats env values containing $ characters as literal", () => {
+    // Regression: passing a raw value to String.replace() causes patterns
+    // like `$1`, `$&`, `$$` to be interpreted as backreferences, silently
+    // corrupting paths like `/home/$USER/project`.
+    process.env.WEIRD = "/home/$USER/$1/$&";
+    expect(resolveSettingsVariables("path=${env:WEIRD}")).toBe(
+      "path=/home/$USER/$1/$&",
+    );
+  });
+
+  it("resolves ${workspaceFolder} using the provided Uri", () => {
+    const folder = mockUri("/workspace/project");
+    expect(
+      resolveSettingsVariables("${workspaceFolder}/models", folder),
+    ).toBe("/workspace/project/models");
+  });
+
+  it("treats workspace folder paths containing $ characters as literal", () => {
+    // Windows paths can legitimately contain `$` (e.g. hidden admin shares).
+    const folder = mockUri("/weird/$1/path");
+    expect(
+      resolveSettingsVariables("${workspaceFolder}/models", folder),
+    ).toBe("/weird/$1/path/models");
+  });
+
+  it("handles mixed ${env:VAR} and ${workspaceFolder} substitutions", () => {
+    process.env.PROFILE = "dev";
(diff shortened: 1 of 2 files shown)


ralphstodomingo:
## Docker E2E verification results

Reproduced both bugs against the OLD implementation and confirmed the fix, all inside the `docker-setup-code-server-1` container running the actual compiled extension bundle.

### Container setup

Env vars set via a throwaway `docker-compose.override.yml` (not committed):

```yaml
environment:
  - TEST_VAR1=alpha
  - TEST_VAR2=beta
  - WEIRD_VAR=/tmp/$$1/path/$$USER    # literal "/tmp/$1/path/$USER" after compose escape
  - REPEATED=x
```

`.vscode/settings.json` written into `/home/coder/jaffle-shop-duckdb/`:

```json
{
  "dbt.runModelCommandAdditionalParams": [
    "--vars", "a=${env:TEST_VAR1},b=${env:TEST_VAR2}",
    "--profiles-dir", "${env:WEIRD_VAR}",
    "--target", "${env:REPEATED}/${env:REPEATED}"
  ]
}
```

### Bug 1 — multi-placeholder iteration failure

Ran a verification script inside the container using the OLD implementation (verbatim from pre-fix `master`) against the real env vars:

```
=== OLD implementation output ===
  [0] --vars
  [1] a=alpha,b=${env:TEST_VAR2}    ← FAIL: second placeholder not resolved
  [2] --profiles-dir
  [3] /tmp/$1/path/$USER
  [4] --target
  [5] x/x
```

Same inputs with the NEW implementation from this PR:

```
=== NEW implementation output ===
  [0] --vars
  [1] a=alpha,b=beta                 ← PASS
  [2] --profiles-dir
  [3] /tmp/$1/path/$USER             ← PASS
  [4] --target
  [5] x/x                            ← PASS
```

All 6 params pass assertion against expected output.

### Bug 2 — `$&` backreference corruption

`$1` on its own doesn't visibly manifest Bug 2 because the inner regex has zero capture groups, so `$1` is treated as literal per ECMA-262. `$&`, however, always references the whole match. Tested separately with `WEIRD=/tmp/$&/backup`:

```
Input:     path=${env:WEIRD}
Env WEIRD: /tmp/$&/backup
Want:      path=/tmp/$&/backup

OLD code:  path=/tmp/${env:WEIRD}/backup    FAIL   ← $& expanded to whole match
NEW code:  path=/tmp/$&/backup              PASS
```

### Unit tests

`src/test/suite/resolveSettingsVariables.test.ts` — 9/9 passing locally. Both bug scenarios are regression-guarded with explicit assertions.

```
PASS src/test/suite/resolveSettingsVariables.test.ts
  resolveSettingsVariables
    ✓ resolves a single ${env:VAR} placeholder
    ✓ resolves MULTIPLE ${env:VAR} placeholders in one string
    ✓ resolves repeated occurrences of the same ${env:VAR}
    ✓ leaves unresolved ${env:VAR} placeholders untouched
    ✓ treats env values containing $ characters as literal
    ✓ resolves ${workspaceFolder} using the provided Uri
    ✓ treats workspace folder paths containing $ characters as literal
    ✓ handles mixed ${env:VAR} and ${workspaceFolder} substitutions
    ✓ returns empty string unchanged
Tests:       9 passed, 9 total
```

### Confirmed in bundle

`grep "envValue !== undefined ? envValue : match" dist/extension.js` → match found — the compiled extension bundle served by code-server contains the fix.

mdesmet (approved):

PR #1879, opened 2026-04-10
fix: [#1720] recognize singular data tests in Run Test action

### LINKED ISSUE #1720: "Run Test" action does not recognize singular data test
### Expected behavior

If I'm editing a singular test under `tests`, running the "Run Test" command should launch `dbt test` with the appropriate arguments.

### Actual behavior

It actually spawns `dbt run`.

### Steps To Reproduce

1. Create a singular test under `tests/`.
2. In the command palette, select `Run Test`.
3. Gaze at the glorious log trace graciously stating how it will run no model, while there is nothing about running any test.

### Log output/Screenshots

_No response_

### Operating System

Ubuntu 25.04

### dbt version

1.8.7

### dbt Adapter

Databricks (1.8.7)

### dbt Power User version

0.58.7

ralphstodomingo:
## Summary

Closes #1720.

Invoking `Run Test` or `Test dbt Model` on a singular data test file under `tests/` spawned `dbt run` instead of `dbt test`. This PR adds a dedicated `RunTest` class that classifies the active file via the parsed manifest and dispatches singular tests through `dbt test --select <test_name>`.

## Root cause

`dbtPowerUser.runTest` is registered as a tree-item command. When invoked from the command palette without a tree item, it fell through to `runModelOnActiveWindow(RunModelType.TEST)`, which unconditionally called `runDBTModel(fullPath, type)`. The `type === TEST` branch was never handled, so the active file was executed as a model via `dbt run`. The same bug also affected `dbtPowerUser.testCurrentModel` on a singular test file (routed through `runModelTest`, which resolves tests attached to a model — not the right path for a file that *is* a test) and `dbtPowerUser.runCurrentModel` (spawns `dbt run` on a test file).

## Fix

A new `RunTest` class (`src/commands/runTest.ts`) owns singular-test classification and execution. The command registrations in `src/commands/index.ts` consult `RunTest` first and fall back to `RunModel` only when the active file is not a singular test.

**Classification** (`RunTest.getSingularTestName`):
- Looks up the active file in the manifest's `testMetaMap`
- Matches on `path === uri.fsPath` **and** `test_metadata === undefined`
- Generic tests (`not_null`, `unique`, etc.) live in `schema.yml` and carry `test_metadata`, so they're filtered out
- Uses the parsed manifest, not a hardcoded `tests/` prefix, so projects with custom `test-paths` in `dbt_project.yml` work

**Dispatch** (`RunTest.runSingularTestOnActiveWindowIfApplicable`): runs the active file as a singular test and returns `true` when it did, `false` otherwise — so command registrations can cleanly fall through to the model path.

**Command wiring** in `src/commands/index.ts`:
- `dbtPowerUser.runCurrentModel` — route through `RunTest` first (`dbt run` on a test file is never meaningful).
- `dbtPowerUser.testCurrentModel` — singular tests must be selected by their own name, not by a surrounding model.
- `dbtPowerUser.runTest` — tree-item invocation is unchanged (tree items are always models); command-palette invocation (no tree item) goes through `RunTest` first.

## Files

- `src/commands/runTest.ts` — new `RunTest` class
- `src/commands/index.ts` — inject `RunTest`, dispatch at the three command entry points
- `src/inversify.config.ts` — bind `RunTest`, wire into `VSCodeCommands`
- `src/test/suite/runTest.test.ts` — 8 Jest cases covering the classifier and the dispatch boolean contract

### Unit tests

```
PASS src/test/suite/runTest.test.ts
  RunTest — singular test classification and dispatch
    getSingularTestName
      ✓ returns the test name for a singular test file
      ✓ returns undefined for a regular model file
      ✓ returns undefined for a generic test (has test_metadata)
      ✓ returns undefined when the manifest event is missing
    runSingularTestOnActiveWindowIfApplicable
      ✓ dispatches runTest and returns true for a singular test file
      ✓ returns false for a regular model file (caller falls back to RunModel)
      ✓ returns false when there is no active editor
      ✓ returns false when the manifest has no matching test
```

Full Jest suite: **21 suites, 265 passed, 1 skipped, 0 failing**. `tsc --noEmit` clean.

### Docker runtime verification

Paired before/after runs against `origin/master` HEAD (`9a0ba3ad`) and this branch, same jaffle-shop-duckdb project, same Playwright sequence (`F1 → Run Test → Enter`), grepping `dbt.log` for `command_name` after each invocation. Summary:

| State | Active file | Command invoked | `command_name` |
|---|---|---|---|
| Master HEAD | `tests/singular_nonnull_orders.sql` | `dbt Power User: Run Test` | `run` ← bug reproduces |
| This branch | `tests/singular_nonnull_orders.sql` | `dbt Power User: Run Test` | `test` ← fixed |
| This branch | `models/orders.sql` | `dbt Power User: Run dbt Model` | `run` ← regression check, model path unchanged |

Full log snippets and the side-by-side writeup are in the follow-up comment on this PR.

## Test plan

- [x] `npx jest src/test/suite/runTest.test.ts` — 8/8 pass
- [x] Full Jest suite green (265 passing, 1 skipped)
- [x] `tsc --noEmit` clean
- [x] `npx webpack --mode development` — builds clean
- [x] Docker before/after: `Run Test` on a singular test file returns `command_name: "run"` on master HEAD and `command_name: "test"` on this branch
- [x] Regression check: `Run dbt Model` on a regular model file still returns `command_name: "run"` on this branch (model path unaffected by the `RunTest` dispatcher)
- [x] `dist/extension.js` contains the dispatch hook at the expected call sites (0 on master, 4 on this branch)
- [ ] Manual verification on a real dbt-core + snowflake/databricks project with seeded data
- [ ] Confirm `Test dbt Model` on a regular model still runs the tests attached to that model (generic tests path unchanged)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

Generated with [Claude Code](https://claude.com/claude-code)

## New Features
- Added automatic detection and intelligent routing for singular dbt data tests
- Enhanced model and test run commands to automatically detect singular test files and route them to the appropriate executor
- Improved developer workflow with seamless test file identification and execution, eliminating confusion between model runs and test executions


src/commands/runModel.ts
```diff
@@ -3,17 +3,57 @@ import { RunModelType } from "@altimateai/dbt-integration";
...
+
+  /**
+   * Returns the singular test name for the given file URI, or undefined if
+   * the file is not a singular data test in the current project's manifest.
+   *
+   * Singular tests are SQL files under the project's configured `test-paths`
+   * that appear in the manifest as test resources without `test_metadata`
+   * (generic tests have `test_metadata`, singular tests do not).
+   */
+  private getSingularTestName(uri: Uri): string | undefined {
```
mdesmet: This is not a model, can we not create a new class instead and dispatch based on the location (models folder vs tests foolder)?

ralphstodomingo:
## Before/after verification against `origin/master` HEAD

Two containers, same worktree state-swapped via `git reset --hard`, same jaffle-shop-duckdb project with a singular test file at `tests/singular_nonnull_orders.sql`, same sequence in both (navigate via Playwright → F1 → type "Run Test" → Enter → grep `dbt.log` for `command_name`).

Master HEAD at time of run: `9a0ba3ad chore: bump @altimateai/ui-components to 0.0.82-beta.4 (#1876)`.

### BEFORE — `dbtpu-1720m` (built from `origin/master` HEAD)

Bundle integrity: `grep -c runSingularTestOnActiveWindowIfApplicable dist/extension.js` → **0** (dispatch hook not present, as expected).

After triggering `dbt Power User: Run Test` on `tests/singular_nonnull_orders.sql`:

```
Running with dbt=1.11.8
...
Resource report: {"command_name": "run", "command_success": true, ...}
```

**`command_name: "run"` — bug reproduces on master HEAD.**

### AFTER — `dbtpu-1720f` (built from `fix/1720-run-test-singular` HEAD)

Bundle integrity: `grep -c runSingularTestOnActiveWindowIfApplicable dist/extension.js` → **4** (dispatch hook + 3 command dispatchers).

After triggering `dbt Power User: Run Test` on the same `tests/singular_nonnull_orders.sql`:

```
  select * from "jaffle_shop"."main"."orders" where customer_id is null
...
Resource report: {"command_name": "test", "command_success": false, ...}
```

**`command_name: "test"` — fix works.**

The compiled SQL in the dbt log is the singular test's actual query (`select * from {{ ref("orders") }} where customer_id is null`), proving dbt test was invoked with the right test selection, not `dbt run` on a model. `command_success: false` is expected — the test is designed to detect null `customer_id` rows, and the seeded data has some, so the test fires as it should.

### Secondary code-path / regression check

The PR wires the `RunTest` dispatcher into three command entry points: `runCurrentModel`, `testCurrentModel`, and `runTest`. For non-test files the classifier returns `undefined` and the dispatcher falls through to `RunModel` unchanged. Verified on the fix container by opening `models/orders.sql` and invoking `dbt Power User: Run dbt Model`:

```
Resource report: {"command_name": "run", "command_success": false, ...}
```

`command_name: "run"` on a regular model file — the model path is unaffected by the dispatcher. Exactly the same behavior as master for model files.

### Summary

| State | Active file | Command invoked | `command_name` |
|---|---|---|---|
| Master HEAD | `tests/singular_nonnull_orders.sql` | `dbt Power User: Run Test` | `run` ← bug |
| Fix branch  | `tests/singular_nonnull_orders.sql` | `dbt Power User: Run Test` | `test` ← fixed |
| Fix branch  | `models/orders.sql` | `dbt Power User: Run dbt Model` | `run` ← regression clean |

The primary bug reproduces on master HEAD, the fix resolves it, and the model path is untouched. Both containers torn down after the run.

src/commands/runModel.ts
ralphstodomingo: Addressed in the follow-up commit `2c879791` — the singular-test handling was extracted into a dedicated `RunTest` class (`src/commands/runTest.ts`) that gets bound in the DI container and dispatched from `VSCodeCommands` at three entry points (`runCurrentModel`, `testCurrentModel`, `runTest`). `RunTest.getSingularTestName` uses the manifest's `resource_type === 'test'` plus `original_file_path` as the classifier, and falls through to `RunModel` unchanged for non-test files. `runModel.ts` has been reverted to its pre-fix state in this PR. Let me know if the placement/dispatch shape isn't what you had in mind.

mdesmet (approved):

PR #1880, opened 2026-04-10
fix: [#1706] disambiguate model and snapshot sharing the same name in docs editor

### LINKED ISSUE #1706: Model documentation won't work if a model and snapshot share the same name
### Expected behavior

The documentation editor should retrieve the metadata of the model, not the snapshot.

### Actual behavior

The documentation editor currently doesn't allow to retrieve or edit the documentation.

### Steps To Reproduce

Create a model and snapshot with the same name (for example `orders.sql`)

### Log output/Screenshots

Image

### Operating System

macos

### dbt version

1.9

### dbt Adapter

snowflake

### dbt Power User version

0.58.2

ralphstodomingo:
## Summary

Fixes #1706. When a dbt project has a model and a snapshot that share the same filename (e.g. `models/marts/orders.sql` and `snapshots/orders.sql`), the Documentation editor either refused to load with *"This file appears to be a snapshot, macro, test, or other dbt resource type which is not supported"* or silently showed the snapshot's metadata when opening the model file.

## Root cause

`DocGenService.getCurrentNode` resolved the manifest node via `nodeMetaMap.lookupByBaseName(modelName)`, which is keyed by bare filename. The underlying `NodeMetaMapImpl` (in `@altimateai/dbt-integration`) builds `modelNameLookupMap` from `path.parse(fullPath).name` for **every** `isResourceNode` entry — and `isResourceNode` returns `true` for **both** models and snapshots. As a result, a snapshot with the same filename as a model last-write-wins the name map, shadowing the model. The extension then either:

1. Returned the snapshot, tripped the `resource_type !== "model"` guard in `getDocumentation`, and surfaced the snapshot error message — even though the user opened the model file, or
2. Returned whichever was iterated last non-deterministically across reloads.

## Fix

Resolve the current node by iterating `nodeMetaMap.nodes()` and matching on the full file `path` **AND** `resource_type === "model"`. A snapshot can never win the tie because its resource_type is `"snapshot"`. Falls back to `lookupByBaseName` only when no node has a matching `path` (e.g. dbt Cloud manifests, where packaged nodes may have `path === undefined`), preserving existing behaviour for cloud setups.

## Files changed

- `src/services/docGenService.ts` — rewrite `getCurrentNode` to accept `filePath` and resolve via `path` + `resource_type` with a dbt-Cloud fallback. Update the single caller in `getDocumentation`.
- `src/test/suite/docGenService.test.ts` — new Jest regression suite (5 cases) covering both manifest iteration orders, snapshot-path resolution, cloud fallback, and missing-event guard.

### Unit tests (passing)

```
PASS src/test/suite/docGenService.test.ts
  DocGenService - model/snapshot name collision (#1706)
    ✓ returns the model when the snapshot is iterated last (would previously shadow it)
    ✓ returns the model when the model is iterated last
    ✓ returns the snapshot when the active file is actually the snapshot path
    ✓ falls back to lookupByBaseName when no node has a matching path (dbt Cloud)
    ✓ returns undefined when there is no manifest event
```

Full `npx jest` run: **21 suites, 262 passed, 1 skipped**. ESLint clean. `tsc --noEmit` clean.

### Docker runtime verification ✅

Reproduced the collision end-to-end against the real `@altimateai/dbt-integration@0.2.12` package and a real `target/manifest.json` inside `dbtpu-1706` (jaffle-shop-duckdb + a `snapshots/orders.sql` alongside the existing `models/orders.sql`):

- `NodeParser.createNodeMetaMap` returns a 9-node map containing both `model.jaffle_shop.orders` and `snapshot.jaffle_shop.orders` with their correct file paths
- This PR's `getCurrentNode` logic returns the **model** (`unique_id: model.jaffle_shop.orders`) when queried with `/home/coder/jaffle-shop-duckdb/models/orders.sql`
- The **old** `lookupByBaseName("orders")` path still returns the **snapshot** — reproducing the reported bug against the same package version

Full evidence (script, output, manifest snippet) in a follow-up comment on this PR.

## Test plan

- [x] `npx jest src/test/suite/docGenService.test.ts` — all 5 cases pass
- [x] Full `npx jest` green (262 passing, 1 skipped)
- [x] ESLint clean on `src/services/docGenService.ts` and the new test file
- [x] `tsc --noEmit` clean
- [x] Docker repro on jaffle-shop-duckdb with a conflicting `snapshots/orders.sql` — verified against real `NodeParser` + real manifest (see comment below)
- [ ] Regression check: opening a standalone model (no conflicting snapshot) still loads its doc
- [ ] Regression check: opening a standalone snapshot still produces the "snapshot is not supported" warning (not a silent pass)
- [ ] dbt Cloud smoke test (if available): doc editor still works when `path === undefined`

### DIFF
--- a/src/services/docGenService.ts
+++ b/src/services/docGenService.ts
@@ -121,13 +121,55 @@ export class DocGenService {
     };
   }
 
-  private getCurrentNode(modelName: string): NodeMetaData | undefined {
+  /**
+   * Resolve the manifest node backing a .sql file for the documentation editor.
+   *
+   * The underlying `nodeMetaMap` stores both models and snapshots in a single
+   * name-keyed map, so `lookupByBaseName` cannot disambiguate when a model and
+   * snapshot share a filename (e.g. `orders.sql` exists as both). To guarantee
+   * the model is returned (never the snapshot), we iterate the node list and
+   * match on the file path AND `resource_type === "model"`.
+   *
+   * We fall back to `lookupByBaseName` only when no node has a matching `path`
+   * (e.g. dbt Cloud, where packaged nodes may have `path === undefined`), so
+   * existing behaviour is preserved for setups without a local path.
+   */
+  private getCurrentNode(
+    modelName: string,
+    filePath: string,
+  ): NodeMetaData | undefined {
     const eventResult = this.queryManifestService.getEventByCurrentProject();
     if (!eventResult || !eventResult.event) {
       return undefined;
     }
 
     const { event } = eventResult;
+
+    // Primary resolution: match by full file path AND model resource type so
+    // a snapshot sharing the same filename never shadows the model.
+    let pathMatchFound = false;
+    for (const node of event.nodeMetaMap.nodes()) {
+      if (node.path !== filePath) {
+        continue;
+      }
+      pathMatchFound = true;
+      if (node.resource_type === RESOURCE_TYPE_MODEL) {
+        return node;
+      }
+    }
+
+    // If any node matched the path but none was a model (e.g. only a snapshot
+    // lives at this path), return the first path-matched node so the caller's
+    // downstream `resource_type` guard can produce an accurate error message.
+    if (pathMatchFound) {
+      for (const node of event.nodeMetaMap.nodes()) {
+        if (node.path === filePath) {
+          return node;
+        }
+      }
+    }
+
+    // Fallback: dbt Cloud / remote manifests may not populate `path`.
     return event.nodeMetaMap.lookupByBaseName(modelName);
   }
 
@@ -354,8 +396,9 @@ export class DocGenService {
       };
     }
 
-    // Node validation
-    const currentNode = this.getCurrentNode(modelName);
+    // Node validation - resolve by file path to disambiguate models from
+    // snapshots that share the same filename.
+    const currentNode = this.getCurrentNode(modelName, filePath);
     if (!currentNode) {
       return {
         documentation: undefined,
--- a/src/test/suite/docGenService.test.ts
+++ b/src/test/suite/docGenService.test.ts
@@ -1,0 +1,176 @@
+import {
+  NodeMetaData,
+  NodeMetaMap,
+  RESOURCE_TYPE_MODEL,
+} from "@altimateai/dbt-integration";
+import { describe, expect, it, jest } from "@jest/globals";
+import { DocGenService } from "../../services/docGenService";
+
+/**
+ * Regression tests for https://github.com/AltimateAI/vscode-dbt-power-user/issues/1706
+ *
+ * The documentation editor resolved manifest nodes by bare filename via
+ * `nodeMetaMap.lookupByBaseName(modelName)`. Because both models and snapshots
+ * populate the same name-keyed lookup map, a snapshot sharing a filename with
+ * a model (e.g. `orders.sql` existing as both) would shadow the model and the
+ * editor either showed snapshot metadata or refused to load with
+ * "This file appears to be a snapshot". The fix routes resolution through the
+ * file path AND resource_type so the collision can never occur.
+ */
+describe("DocGenService - model/snapshot name collision (#1706)", () => {
+  const MODEL_PATH = "/workspace/project/models/marts/orders.sql";
+  const SNAPSHOT_PATH = "/workspace/project/snapshots/orders.sql";
+
+  const modelNode: NodeMetaData = {
+    unique_id: "model.project.orders",
+    path: MODEL_PATH,
+    database: "db",
+    schema: "schema",
+    alias: "orders",
+    name: "orders",
+    package_name: "project",
+    description: "The canonical orders model",
+    patch_path: "",
+    columns: {
+      order_id: {
+        name: "order_id",
+        description: "PK",
+        data_type: "integer",
+        meta: {},
+      },
+    },
+    config: { materialized: "table" },
+    resource_type: RESOURCE_TYPE_MODEL,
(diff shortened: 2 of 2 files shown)


mdesmet:
isnt this an issue for altimate-dbt-integration?

ralphstodomingo:
## Docker runtime verification ✅

Reproduced the collision scenario end-to-end against the real `@altimateai/dbt-integration@0.2.12` package and a real `target/manifest.json`:

**Setup (inside `dbtpu-1706` code-server container):**

```bash
# 1. Add a colliding snapshot alongside the existing model
mkdir -p /home/coder/jaffle-shop-duckdb/snapshots
cat > /home/coder/jaffle-shop-duckdb/snapshots/orders.sql << 'EOF'
{% snapshot orders %}
  {{ config(target_schema="snapshots", unique_key="order_id", strategy="check", check_cols=["status"]) }}
  select * from {{ ref("stg_orders") }}
{% endsnapshot %}
EOF

# 2. Populate the manifest
cd /home/coder/jaffle-shop-duckdb && dbt seed --quiet && dbt run --quiet && dbt snapshot --quiet
```

Manifest now contains both nodes:

```json
{
  "model.jaffle_shop.orders": "model",
  "snapshot.jaffle_shop.orders": "snapshot"
}
```

**Verification script** (runs the real `NodeParser` against the real manifest, then replicates the fix's exact logic and compares against the old `lookupByBaseName` path):

```js
const { NodeParser, RESOURCE_TYPE_MODEL } = require("@altimateai/dbt-integration");
const manifest = require("/home/coder/jaffle-shop-duckdb/target/manifest.json");

const parser = new NodeParser({ debug: () => {}, error: console.error });
const nodeMetaMap = await parser.createNodeMetaMap(manifest.nodes, {
  getProjectRoot: () => "/home/coder/jaffle-shop-duckdb",
  getProjectName: () => "jaffle_shop",
  getPackageInstallPath: () => "/home/coder/jaffle-shop-duckdb/dbt_packages",
  getTargetPath: () => "/home/coder/jaffle-shop-duckdb/target",
});
// ... enumerate nodes, run fix logic, compare with lookupByBaseName
```

**Output:**

```
total nodes in map: 9
nodes by resource_type: { "model": 5, "snapshot": 1, "seed": 3 }
orders-named nodes: [
  {
    "unique_id": "model.jaffle_shop.orders",
    "resource_type": "model",
    "path": "/home/coder/jaffle-shop-duckdb/models/orders.sql"
  },
  {
    "unique_id": "snapshot.jaffle_shop.orders",
    "resource_type": "snapshot",
    "path": "/home/coder/jaffle-shop-duckdb/snapshots/orders.sql"
  }
]

=== FIX VERIFICATION ===
query filePath: /home/coder/jaffle-shop-duckdb/models/orders.sql
pathMatchFound: true
model match: { unique_id: 'model.jaffle_shop.orders', resource_type: 'model' }   ✅
fallback: null

=== OLD (buggy) lookupByBaseName("orders") ===
{
  unique_id: 'snapshot.jaffle_shop.orders',
  resource_type: 'snapshot',
  path: '/home/coder/jaffle-shop-duckdb/snapshots/orders.sql'
}   ← exactly the reported bug
```

**Conclusions:**

- The real runtime `NodeMetaMap` contains both the model and the snapshot with their correct full paths, so the path-based disambiguation in this PR's `getCurrentNode` fires correctly.
- The old `lookupByBaseName("orders")` path still returns the snapshot — reproducing the bug from the issue against the same package version the extension ships with today.
- The new code path returns `model.jaffle_shop.orders` as expected.

The full-UI flow (opening the documentation editor from the command palette in code-server) was blocked by code-server's default `mssql` language association on `.sql` files — the `goToDocumentationEditor` command's `when: "resourceLangId =~ /^sql$|^jinja-sql$/"` hides the palette entry unless the file is re-associated. The verification above exercises the exact code path (`getCurrentNode` running against a real `NodeMetaMap` built by `@altimateai/dbt-integration`) without being gated on the language-association dance, so it's both stronger and more deterministic.

Container name: `dbtpu-1706`, host port `3021`, torn down after verification.

ralphstodomingo:
Closing in favor of an upstream fix in `AltimateAI/altimate-dbt-integration`.

You're right that this is a root-cause issue in `@altimateai/dbt-integration`. `NodeParser` (in `src/parsers/nodeParser.ts`) builds `modelNameLookupMap` by iterating every `isResourceNode` entry — which includes **models, seeds, analyses, AND snapshots** — and keys them by bare filename in a single map. Any name collision across resource types silently wins last-write. The method is called `lookupByBaseName`, so callers can't even tell they might be getting a snapshot back when they expect a model.

Survey of callers across `vscode-dbt-power-user` (grepped the current `master`):

| Caller | Expects |
|---|---|
| `docGenService` (this issue) | model only |
| `dbtTestService`, `sqlLineagePanel`, `newLineagePanel` | model only |
| `modelHoverProvider`, `modelDefinitionProvider`, `modelTreeviewProvider` | model only |
| `validateSql`, `commands/index` | model only |
| `conversationProvider` | any resource |
| `graphParser` (inside altimate-dbt-integration) | any resource |

Vast majority want models. The cleanest fix is a **models-first** two-pass build in `NodeParser`:

1. First pass: populate `modelNameLookupMap` from nodes where `resource_type === "model"` only
2. Second pass: populate entries for snapshots/seeds/analyses only if the name isn't already claimed by a model

This is zero breakage for existing callers — callers that don't care about type still get *something* back, and the many callers that want models always get the model in collision cases. No API change, no version-bumping caller concerns.

New PR incoming in `AltimateAI/altimate-dbt-integration`. I'll link it here once it's up.

PR #1894, opened 2026-04-17
feat: add "Execute dbt Model" command (#1875)

### LINKED ISSUE #1875: Extension is incompatible with upstream-prod package
### Expected behavior

I'm the maintainer of [upstream-prod](https://github.com/LewisDavies/upstream-prod) - I'm having an nightmare trying to make the package work with dbt Power User and could use some help. 

When I execute an open dbt model, I expect the extension to use the package logic to determine whether refs should be directed to production or development relations. My package relies on [selected_resources](https://docs.getdbt.com/reference/dbt-jinja-functions/selected_resources) being populated to do this. Since Power User doesn't run commands directly, it skips my package logic and refs are always directed to prod relations.

When I use `dbt run/build/compile` myself, this [macro](https://github.com/LewisDavies/upstream-prod/blob/main/macros/populate_cache.sql) is triggered at the start of the run to get the last_altered time of all upstream models. I'd like the same to happen when when running a model with Power User.

### Actual behavior

After some recent changes to my package I realised that Power User has been bypassing upstream-prod when running a dbt model with cmd+Enter. This seems to have been happening for long time. What makes it especially confusing is that Power User seems to maintain a local cache that changes query behaviour, so I'm getting very different error messages across runs while trying to develop a workaround.

Note I have only observed this when previewing a dbt model with cmd+Enter; it may be happening when other Power User commands are triggered but I prefer using the dbt in the terminal when running complete models.

### Steps To Reproduce

1. Add the package to your project and set the [required variables](https://github.com/LewisDavies/upstream-prod?tab=readme-ov-file#setup)
``` 
# packages.yml
- package: LewisDavies/upstream_prod
   version: 0.10.1 

# dbt_project.yml
vars:
  # Add the other package variables outlined in the README
  upstream_prod_prefer_recent: true
```
2. Open a model that has already been materialised in production. Run it with cmd+Enter and you should see the 500-row preview.
3. Open the terminal and run a model upstream of the one you just previewed.
4. Run the model from point 2 again. Instead of using the more decent dev relation, it will still be using the prod version as it was before.

### Log output/Screenshots

```
Compilation Error in sql_operation t_dd3ce838cfff432f87a4365daf8adeb3 (from remote system.sql)
No first item, sequence was empty. > in macro ref (macros/ref.sql) > called by macro ref (macros/ref.sql) > called by sql_operation t_dd3ce838cfff432f87a4365daf8adeb3 (from remote system.sql)
```

### Operating System

macOS 15.6 (24G84)

### dbt version

1.9.10

### dbt Adapter

Snowflake

### dbt Power User version

0.60.4

ralphstodomingo:
Closes #1875.

## Problem

`Execute SQL` previews the editor contents with `dbt show --inline`, which compiles the SQL as an **anonymous node** (`inline_<hash>`). Any Jinja that reads the current node's identity — `model.name`, `this`, `model.config`, or a macro derived from them — therefore resolves against a placeholder rather than the model in front of you.

Two symptoms, same cause:

- #1875 (`upstream-prod`): `--inline` leaves `selected_resources` empty, so packages reading it misbehave.
- #2028: a model whose macro asserts on `model.name` fails to compile, and the user sees an unreadable error.

`--inline` is the right thing for a highlighted fragment or an unsaved edit — there is no real node to point dbt at. It is the wrong thing for a saved model file, and dbt offers no way to have both: there is no `--as-node <name>` for inline SQL.

## Change

**1. A second, opt-in command.** `Execute dbt Model` (`Ctrl/Cmd+Shift+Enter`) previews the file as its real node via `dbt show --select <model>`. `Execute SQL` (`Ctrl/Cmd+Enter`) is unchanged and still uses `--inline`.

Deliberately opt-in rather than auto-routing: pressing a different key *is* the user declaring intent, so nothing has to be inferred. Auto-detection was considered and rejected — `getQuery()` returns the selection when one exists, so highlighting a CTE inside `stg_orders.sql` yields `modelName="stg_orders"` with fragment SQL. Auto-routing there would silently run the whole model and return rows the user never asked for, **with no error at all** — strictly worse than today's loud failure. Two of the four `executeSQL` call sites also pass synthetic names (`"cte_query"`, `cte_<name>_<hash>`).

**2. A hint, so the command is discoverable at the moment it's needed.** The opt-in design's one weakness is that nobody finds it. When an inline preview fails to compile, the error now carries a line naming the command. Detection gates *a sentence, never behaviour*, so a false positive costs a line of text rather than wrong results.

Shown only when all of these hold:

- the whole saved file was previewed — not a selection, not unsaved edits
- the file resolves to a real node in the manifest
- the integration implements the command (Python-bridge mode throws `NotImplementedError`, so no hint there)

The hint names the command rather than its keybinding: the chord differs per platform and users can rebind it.

## Verified end to end

Built and driven in code-server against a dbt project whose macro asserts on `model.name`, on the `corecommand` path (which shares `--inline` and was the worse of the two — it previously died with a bare `TypeError`):

- **before:** `TypeError: Cannot read properties of undefined (reading 'data')`
- **after:** `Model 'inline_query' must start with stg_ … Hint: … use the "Execute dbt Model" command.`
- **`Execute dbt Model`:** returns `node_name = stg_identity_probe` — identity preserved, macro passes, real rows
- **with a selection:** same underlying error, **no hint** — gating confirmed live

End-to-end verification also caught a defect that no unit test would have: setting `Error.name` made the query panel render `InlinePreviewCompilationError:` as the user-facing heading. Fixed library-side and re-verified.

## Merge order — this PR is blocked

The extension side compiles only against an unreleased library. CI fails with:

```
error TS2305: Module '"@altimateai/dbt-integration"' has no exported member 'NotImplementedError'.
error TS2305: Module '"@altimateai/dbt-integration"' has no exported member 'isInlinePreviewCompilationError'.
error TS2551: Property 'executeModelWithLimit' does not exist on type 'DBTProjectIntegrationAdapter'.
```

Expected, not a regression. It needs, in order:

1. AltimateAI/altimate-dbt-integration#52 — `executeModel` API **and** the inline compile-error surfacing (both halves, consolidated into that one PR)
2. a `@altimateai/dbt-integration` release, and the version bump here

Until then this stays draft and red. Verified locally against a build of that branch: typecheck clean, 638 tests passing.

## Tests

7 new unit tests covering the gate and the wrapper: hint offered for a saved unedited model; suppressed for a selection, for unsaved edits, for Python-bridge mode, and for a file that is not a manifest node; hint appended to an inline compile failure; unrelated errors passed through untouched; successful results unaffected.


ralphstodomingo:
## Docker E2E — end-to-end verification through the extension

Verified the full chain: keybinding → command handler → integration adapter → CLI invocation → dbt runtime.

### Setup

- `codercom/code-server` with the extension bundled via `webpack --mode development`
- `@altimateai/dbt-integration` npm-linked to the branch in AltimateAI/altimate-dbt-integration#52
- `jaffle-shop-duckdb` fixture at `/home/coder/project` with `dbt.dbtIntegration: "corecommand"` and a `log_selected_resources()` macro invoked from `customers.sql`

### Evidence

**Command registered with correct keybinding:**

Command palette shows `dbt Power User: Execute dbt Model` with `Control+Shift+Enter` — confirming `package.json` wiring is correct.

**`Cmd+Shift+Enter` on `customers.sql` → `dbt show --select customers`:**

The dbt log captured the exact CLI command the extension produced:

```
dbt show --log-level debug --select customers --limit 500 --output json --log-format json
    --project-dir /home/coder/project --profiles-dir /home/coder/project --target dev
```

This is exactly what `DBTCoreCommandProjectIntegration.executeModel` (in dbt-integration#52) builds. The keybinding → `runModel.executeModelOnActiveWindow` → `dbtProjectContainer.executeModel` → `dbtProject.executeModelOnQueryPanel` → `adapter.executeModelWithLimit` → integration chain resolves correctly.

**`selected_resources` populated (the #1875 fix):**

```
SELECTED_RESOURCES=[
  "model.jaffle_shop.customers",
  "test.jaffle_shop.relationships_orders_customer_id__customer_id__ref_customers_.c6ec7f58f2",
  "test.jaffle_shop.not_null_customers_customer_id.5c9bf9911d",
  "test.jaffle_shop.unique_customers_customer_id.c5af1ff4b1"
]
```

The model node is present with its full `unique_id` — packages like `upstream-prod` can now resolve `ref()` calls correctly. Contrast with the `--inline` path verified in AltimateAI/altimate-dbt-integration#52, where `selected_resources` was `["sql_operation.jaffle_shop.inline_query"]` (the synthetic inline-query node) with no model identity.

**Query Results panel opens and renders output:**

The extension's webview received the `executeQuery` event with the new `QueryExecution` from `executeModel` and opened the query-results panel — confirming the integration's output shape matches what the webview consumes (no webview changes needed).

### Regression guard

- `Cmd+Enter` (existing `executeSQL`) continues to produce `--inline` invocations — unchanged on this branch.
- The existing query panel's "re-run current window" button still uses `executeQueryFromActiveWindow` (reads editor + `--inline`), intentionally unchanged.
- Jest tests in dbt-integration#52 include an `executeSQL still uses --inline (unchanged)` regression guard.

ralphstodomingo:
## Docker E2E — full flow through the extension

Re-run end to end on the current branch (the earlier E2E note predated the hint and the rebase onto master, so it has been replaced). Verified the whole chain: keybinding → command handler → integration adapter → CLI invocation → dbt runtime → query panel.

### Setup

- `codercom/code-server` with the extension built by `rsbuild` from this branch
- `@altimateai/dbt-integration` linked to AltimateAI/altimate-dbt-integration#52 (which now carries both the `executeModel` API and the compile-error surfacing)
- `jaffle-shop-duckdb` fixture with `dbt.dbtIntegration: "corecommand"`, a `check_name()` macro that asserts on `model.name`, and a model echoing its own identity

`corecommand` is the harshest case: it shares the `--inline` path with Fusion and previously had no empty-preview guard at all.

### 1. Before — the failure users actually hit

`Cmd+Enter` on the saved model, on clean `master`:

before: bare TypeError

`TypeError: Cannot read properties of undefined (reading 'data')` — nothing about the real problem, which is that the macro rejected the anonymous node name.

### 2. After — the real error, plus the way out

Same keystroke, same file, this branch:

after: real compile error with hint

The actual dbt error surfaces (`Model 'inline_query' must start with stg_`), followed by the hint naming `Execute dbt Model`. Against real Fusion output the message shrinks from **3,913 characters of raw JSON to 256**.

Note the heading reads `Error:` and not the exception class name — surfacing `InlinePreviewCompilationError` to users was a defect this E2E caught and that no unit test would have; fixed library-side.

### 3. The command the hint points at works

`Cmd+Shift+Enter` on the same file:

after: Execute dbt Model returns rows

`node_name = stg_identity_probe`, `this_ident = stg_identity_probe` — identity preserved, the macro passes, real rows come back. This is the payload of the fix in one screenshot.

### 4. The hint stays quiet when it would be wrong

Selecting only line 1 and pressing `Cmd+Enter`:

after: selection suppresses the hint

Status bar shows `18 selected`. Same underlying compile error, **no hint** — because with a selection active, "run it as the real model" is the wrong advice. Gating confirmed in the running product, not just in unit tests.

### 5. `selected_resources` — the #1875 half

Re-verified in the same container rather than carried over from the earlier run:

```
--select  →  SELECTED_RESOURCES=['model.jaffle_shop.stg_identity_probe']
--inline  →  SELECTED_RESOURCES=['sql_operation.jaffle_shop.inline_query']
```

The `--select` path yields the real model node with its full `unique_id`, so packages like `upstream-prod` can resolve `ref()` correctly. The `--inline` path yields the synthetic sql_operation node with no model identity — the original #1875 symptom.

### Regression guard

- `Cmd+Enter` still produces `--inline` invocations; `executeSQL` is untouched on this branch.
- The hint is suppressed for a selection (screenshot 4), for unsaved edits, for untitled buffers, for files that are not manifest nodes, and in Python-bridge mode where the command throws `NotImplementedError`.
- 638 extension tests and 292 library tests pass; typecheck clean against the linked library.

Reproduction fixture is scripted and reusable — see `test-fixtures/inline-preview-loses-node-identity/`.

ralphstodomingo:
Superseded by #2055.

This PR added an opt-in `Execute dbt Model` command plus a hint pointing at it. Review preferred hooking the existing Execute SQL path instead, so there is no new command or keybinding to discover and `Cmd+Enter` just works for the reported case.

#2055 is built fresh on `master` and is a smaller diff than this one — it drops the command, keybinding, menu entry and hint in favour of routing plus a capability check. Closing in favour of #2055.

PR #1926, opened 2026-04-29
fix: guard MCP extension exports undefined when MCP server disabled

ralphstodomingo:
## Why

Surfaced by `/telemetry-report` on dbt-power-user **0.60.7** (2026-04-29):

- **27 events / 17 distinct machines / 7 days, 0.60.7-only** for `DbtPowerUserMcpServer:updateMcpExtensionApiError`
- Single error message across all stacks: `Cannot read properties of undefined (reading 'ready')`

Clean regression from 0.60.5 → 0.60.7. No machines on 0.60.5 hit it.

While digging into the repro, also noticed a separate cosmetic bug in the same handler that mangles every emitted error log into `[object Object]`. Both are folded into this PR.

## Root cause — the `.ready` crash

`vscode-altimate-mcp-server/src/extension.ts:42-45` returns bare `undefined` when the user has `altimate.disableMcpServer = true`:

```ts
if (disableMcpServer) {
  logger.log("MCP Server is disabled via configuration");
  return;
}
```

In dbt-power-user, `src/mcp/index.ts:55` then accesses `extension.exports.ready` without a guard:

```ts
if (!extension.isActive) { await extension.activate(); }
await extension.exports.ready;            // crashes when exports is undefined
```

The throw is caught by the surrounding `try/catch`, so activation continues — but every disabled-MCP machine emits an `[error]` log + a telemetry error event on every startup.

`vscode-altimate-mcp-server/src/pythonEnvironment.ts:64` already uses the right pattern reading dbt-power-user's exports — this PR mirrors it.

## Root cause — the `[object Object]` log corruption

The three `dbtTerminal.error` calls in `mcp/index.ts` wrap the error in `{ message: ... }` before passing it as the third arg:

```ts
this.dbtTerminal.error(
  "DbtPowerUserMcpServer:updateMcpExtensionApiError",
  "Error updating MCP extension API",
  { message: (error as Error).message },  // ← wrapper
);
```

`vscodeTerminal.error` falls into the unknown branch and `${e}`-stringifies the wrapper, producing `[object Object]` in both the dbt output channel and `console.error`. `sendTelemetryError(name, e, { message })` also receives the wrapper, so App Insights stack traces are impoverished too.

## Fix

Three small changes in `src/mcp/index.ts`:

1. Early-return with `info` log when `extension.exports` is undefined (the `.ready` crash).
2. Make `await extension.exports.ready` conditional (current MCP server doesn't expose a `ready` field anyway, so the original await was a no-op).
3. Pass the real `Error` to `dbtTerminal.error` in both catch blocks; for the synthetic "extension not installed" branch, collapse the duplicate text into the second arg and pass `undefined` for the third.

Plus:

- New `src/test/suite/mcpServer.test.ts` — 5 tests covering: missing extension (with proper message + undefined error), disabled MCP (exports undefined), exports-with-ready, exports-without-ready, regression guard for the `[object Object]` corruption.
- `package-lock.json` version bump 0.60.6 → 0.60.7 (missed by the release commit).

## Pre / post-fix evidence (Docker code-server, `altimate.disableMcpServer = true`)

**Pre-fix** — `dbt` output channel (`screenshots/mcp-disabled-undefined-exports/before-dbt-output.log`):

```
[error] DbtPowerUserMcpServer:updateMcpExtensionApiError:Error updating MCP extension API:[object Object] []
```

Browser dev console: 18 errors total, including the line above. Note the `[object Object]` masking the real `Cannot read properties of undefined (reading 'ready')` message — that's the second bug in action.

**Post-fix** — `dbt` output channel (`screenshots/mcp-disabled-undefined-exports/after-cleanup-dbt-output.log`):

```
[info]  DbtPowerUserMcpServer: updateMcpExtensionApi:MCP extension is installed but not exporting an API (likely disabled via altimate.disableMcpServer); skipping integration []
[info]  DbtPowerUserMcpServer: registerToolsInMcpExtension:MCP extension API not found []
```

Browser dev console: 13 errors total — `updateMcpExtensionApiError` line gone. Remaining errors are unrelated (CORS to open-vsx, missing `macros/` in the test project).

**Happy path (MCP enabled, default settings)** — silent, no error, no info, no regression. Verified in `screenshots/mcp-disabled-undefined-exports/after-mcp-enabled-dbt-output.log`.

## Tests

```
$ npx jest
Test Suites: 28 passed, 28 total
Tests:       1 skipped, 461 passed, 462 total
```

Including 5 new tests for the MCP guard and the `[object Object]` regression.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MCP extension integration: now logs informative messages when the extension or its API is unavailable, preserves original errors when reporting failures, and conditionally waits for the extension to signal readiness before registering integrations.

* **Tests**
  * Added tests covering missing extension, disabled/absent API, readiness vs. immediate registration, and error propagation behavior.

### DIFF
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -43,8 +43,8 @@ export class DbtPowerUserMcpServer implements Disposable {
       if (!extension) {
         this.dbtTerminal.error(
           "DbtPowerUserMcpServer: enableMcpExtensionIntegration",
-          "Failed to install MCP extension",
-          { message: "Failed to install Altimate MCP Server extension" },
+          "Failed to install Altimate MCP Server extension",
+          undefined,
         );
         return;
       }
@@ -89,7 +89,7 @@ export class DbtPowerUserMcpServer implements Disposable {
       this.dbtTerminal.error(
         "DbtPowerUserMcpServer:updateMcpExtensionApiError",
         "Error updating MCP extension API",
-        { message: (error as Error).message },
+        error,
       );
     }
   }
@@ -132,7 +132,7 @@ export class DbtPowerUserMcpServer implements Disposable {
       this.dbtTerminal.error(
         "DbtPowerUserMcpServer:registerToolsInMcpExtensionError",
         "Error registering tools in MCP extension",
-        { message: (error as Error).message },
+        error,
       );
     }
   }
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -52,7 +52,18 @@ export class DbtPowerUserMcpServer implements Disposable {
       if (!extension.isActive) {
         await extension.activate();
       }
-      await extension.exports.ready;
+      // The MCP server's activate() returns undefined when
+      // `altimate.disableMcpServer = true`, leaving exports unset.
+      if (!extension.exports) {
+        this.dbtTerminal.info(
+          "DbtPowerUserMcpServer: updateMcpExtensionApi",
+          "MCP extension is installed but not exporting an API (likely disabled via altimate.disableMcpServer); skipping integration",
+        );
+        return;
+      }
+      if (extension.exports.ready) {
+        await extension.exports.ready;
+      }
       this.mcpExtensionApi = extension.exports as ToolRegistry;
 
       await this.mcpExtensionApi.addMcpIntegrationConfig([
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -42,7 +42,7 @@ export class DbtPowerUserMcpServer implements Disposable {
 
       if (!extension) {
         this.dbtTerminal.info(
-          "DbtPowerUserMcpServer: enableMcpExtensionIntegration",
+          "DbtPowerUserMcpServer:enableMcpExtensionIntegration",
           "Altimate MCP Server extension is not installed",
         );
         return;
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -41,10 +41,9 @@ export class DbtPowerUserMcpServer implements Disposable {
       );
 
       if (!extension) {
-        this.dbtTerminal.error(
+        this.dbtTerminal.info(
           "DbtPowerUserMcpServer: enableMcpExtensionIntegration",
-          "Failed to install Altimate MCP Server extension",
-          undefined,
+          "Altimate MCP Server extension is not installed",
         );
         return;
       }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "vscode-dbt-power-user",
-  "version": "0.60.6",
+  "version": "0.60.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "vscode-dbt-power-user",
-      "version": "0.60.6",
+      "version": "0.60.7",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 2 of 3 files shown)


src/mcp/index.ts
```diff
@@ -43,16 +43,27 @@ export class DbtPowerUserMcpServer implements Disposable {
...
+      if (!extension.exports) {
+        this.dbtTerminal.info(
+          "DbtPowerUserMcpServer: updateMcpExtensionApi",
+          "MCP extension is installed but not exporting an API (likely disabled via altimate.disableMcpServer); skipping integration",
+        );
+        return;
+      }
+      if (extension.exports.ready) {
+        await extension.exports.ready;
+      }
```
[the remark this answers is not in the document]
ralphstodomingo: @coderabbitai thanks for the catch — want to walk through my analysis and see if it changes the recommendation:

The premise of the suggestion is *"if this method is called again after a prior successful initialization, `registerToolsInMcpExtension()` will still use the stale object."*

Looking at the call graph, `updateMcpExtensionApi()` has exactly one caller — `dbtPowerUserExtension.ts:96`, inside `activate()`:

```bash
$ rg -nP '\.updateMcpExtensionApi\s*\(' src
src/dbtPowerUserExtension.ts:96:      await this.mcpServer.updateMcpExtensionApi();
```

Single invocation, once per extension lifetime. The other listeners in this class (`onDidChangeConfiguration("altimate.onboardedMcpServer")` and the `dbtProjectsInitialized` event) call `registerToolsInMcpExtension()`, not `updateMcpExtensionApi()`. And `altimate.disableMcpServer` itself isn't a hot-toggle setting — flipping it requires an extension reload (the MCP server's own `activate()` only checks it once at line 38-45 of `vscode-altimate-mcp-server/src/extension.ts`).

So on the disabled-MCP early return, `this.mcpExtensionApi` is still in its initial declared state of `undefined` (line 12) — there's nothing to clear. The stale-reference scenario you describe would require a call path that doesn't currently exist.

Two questions:

1. Did the analysis miss a re-invocation path? If you can point at one, happy to add the clear.
2. If not, would you still recommend setting `this.mcpExtensionApi = undefined` defensively, in case a future change wires up re-invocation? I'm leaning against it (we'd revisit the function anyway when adding such a path) but open to hearing the case for belt-and-suspenders.

For the `dispose()` part — the instance is GC'd right after `dispose()` runs, so explicitly nulling the reference there is cosmetic. Unless the concern is something else (e.g. the listener registered in the constructor capturing `this`)?

src/mcp/index.ts
```diff
@@ -43,16 +43,27 @@ export class DbtPowerUserMcpServer implements Disposable {
       if (!extension) {
         this.dbtTerminal.error(
           "DbtPowerUserMcpServer: enableMcpExtensionIntegration",
```
mdesmet: I think this should have Error Suffix as this logs a telemetry error. why we need to remove the message?

mdesmet: DbtPowerUserMcpServer:enableMcpExtensionIntegrationError

mdesmet: convert into info

ralphstodomingo: Done in 79b9ec54 — converted to `dbtTerminal.info` with message "Altimate MCP Server extension is not installed" and updated the corresponding test. Confirmed the convention holds for the two remaining `.error` calls in this file (`updateMcpExtensionApiError`, `registerToolsInMcpExtensionError`).

dev-punia-altimate:
## ✅ Tests — All Passed



cc @ralphstodomingo

mdesmet (approved):

PR #1939, opened 2026-05-07
fix: don't fail extension activation when a workspace folder has no dbt project

ralphstodomingo:
## Summary

Production telemetry surfaced cluster `extensionActivationError` (**83 unique machines / 104 events / 24h**, both 0.60.7 and 0.61.0) with the stack pinned at:

```
Error: no projects found. retrying
    at DBTWorkspaceFolder.retryWithBackoff
    at async DBTWorkspaceFolder.discoverProjects
    at async DBTProjectContainer.registerWorkspaceFolder
    at async Promise.all (index 0)
    at async DBTProjectContainer.initializeDBTProjects
    at async DBTPowerUserExtension.activate
```

Distribution: ~95% darwin + linux, ~95% `core` integration mode, every top-version contributing — consistent with the multi-root workspace pattern common on dev machines, and inconsistent with a regression.

## Root cause

VS Code's `workspaceContains:**/dbt_project.yml` activation matches when **any** workspace folder contains a `dbt_project.yml`. The extension then runs `discoverProjects` on **every** workspace folder via `Promise.all(folders.map(registerWorkspaceFolder))`. For a folder that genuinely has no dbt project — typical for multi-root workspaces where a sibling folder triggered activation — `findFiles` returns `[]` on every attempt, `retryWithBackoff` exhausts after ~10s (5 attempts with 1+2+3+4 backoff), and the thrown `"no projects found. retrying"` propagates up to the activate-level catch.

The visible damage isn't just the telemetry noise — the throw also short-circuits the rest of `activate()`, so `statusBars.initialize()` and the workspace-folder change listener never run. Users with a mixed-folder workspace get a working dbt project for the folder that matched, plus silently-degraded status-bar UX for the rest of the session.

## Fix

In `retryWithBackoff`, treat a consistently-empty result after the retry budget as legitimately empty and return `[]` instead of throwing. The retry schedule and the empty/non-empty differentiation are unchanged — only the terminal disposition. Real errors from `fn()` (permission failures, malformed pattern, indexer crashes) still throw after retries, preserving the existing escalation for non-empty failure modes.

The constructor's `createConfigWatcher` (line 60) keeps watching for `dbt_project.yml` creation in this folder, so a project added later is still picked up automatically.

This also restructures the loop so empty results are handled by normal control flow rather than via a thrown-and-caught sentinel error, which is cleaner and removes the previously-unreachable `"no projects found after maximum retries"` throw at the bottom of the function.

## Pre-fix → post-fix evidence

A new test file (`src/test/suite/discoverProjectsRepro.test.ts`) instantiates the **real `DBTWorkspaceFolder`** and stubs only `workspace.findFiles`. Same harness shape both pre- and post-fix.

**Pre-fix** (against this branch's HEAD~1, i.e. origin/master) — the harness asserted on the *bug* shape and passed, confirming the production stack reproduces deterministically:

```
✓ discoverProjects throws 'no projects found. retrying' after 5 retries (10007 ms)
✓ retry exhaustion takes ~10s (matches production telemetry cluster pattern) (10005 ms)
✓ multi-root: empty folder rejects in parallel with happy folder (Promise.all index 0 pattern) (10013 ms)
```

**Post-fix** (this branch) — the assertions flip to verify the new behaviour, plus a new fourth case guards against accidentally swallowing genuine errors:

```
✓ returns no projects (does not throw) when findFiles is consistently empty (10024 ms)
✓ retry budget is still ~10s — confirms backoff schedule unchanged (10006 ms)
✓ multi-root: empty folder no longer fails Promise.all alongside happy folder (10007 ms)
✓ real findFiles errors still propagate after retries (preserves existing escalation) (10015 ms)
```

Saved at `screenshots/no-projects-found-investigation/repro-output.txt` (pre-fix) and `screenshots/no-projects-found-investigation/after-fix-repro.txt` (post-fix).

## Test plan

- [x] New test file passes (4 cases, ~40s total — backoff drives the duration)
- [x] Full Jest suite — 32 of 33 suites pass; the only failure (`jupyterlabServicesCompat.test.ts`) is pre-existing on origin/master (missing build artifact in local env), confirmed unrelated by stashing this PR's changes and re-running it
- [x] Retry schedule unchanged — backoff window asserted at `[9.5s, 12s)` in test 2
- [x] Real-error escalation preserved — test 4 confirms `findFiles` rejecting with EACCES still throws after 5 retries

## Summary by CodeRabbit

* **Bug Fixes**
  * Folders with no project files no longer cause activation failures: repeated empty search results are treated as legitimately empty after retry attempts and resolve successfully.
  * Multi-folder workspaces activate even when one or more folders lack project files; genuine errors still trigger retries and are reported via telemetry.

* **Tests**
  * Added regression tests covering multi-root behavior, retry/backoff timing, suppression of telemetry for empty results, and telemetry for real errors.

### DIFF
--- a/src/dbt_client/dbtWorkspaceFolder.ts
+++ b/src/dbt_client/dbtWorkspaceFolder.ts
@@ -30,6 +30,11 @@ import {
   readAndParseProjectConfig,
 } from "@altimateai/dbt-integration";
 
+// Sentinel for the empty-after-retries case in `retryWithBackoff`. Lets the
+// caller distinguish "this folder has no dbt project" (silent, expected for
+// multi-root workspaces) from a real `findFiles` failure (telemetry-worthy).
+class NoProjectsFound extends Error {}
+
 export class DBTWorkspaceFolder implements Disposable {
   private watcher: FileSystemWatcher;
   readonly projectDiscoveryDiagnostics =
@@ -94,26 +99,16 @@ export class DBTWorkspaceFolder implements Disposable {
     while (attempt < retries) {
       try {
         const result = await fn();
-        const isEmptyArray = Array.isArray(result) && result.length === 0;
-        if (!isEmptyArray) {
-          return result;
-        }
-        // Empty array — try again, in case `findFiles` returned early
-        // during VS Code's initial workspace indexing.
-        attempt++;
-        if (attempt < retries) {
+        if (Array.isArray(result) && result.length === 0) {
           this.dbtTerminal.debug(
             "discoverProjects",
             "no projects found. retrying...",
             false,
           );
-          await new Promise((resolve) =>
-            setTimeout(resolve, backoff * attempt),
-          );
+          throw new NoProjectsFound("no projects found. retrying");
         }
+        return result;
       } catch (error) {
-        // Real error from fn() (e.g. permission failure). Preserve the
-        // existing retry-then-throw behaviour for these.
         attempt++;
         if (attempt >= retries) {
           throw error;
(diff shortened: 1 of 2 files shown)


ralphstodomingo:
## Reply to @coderabbitai review

Triaged the three nitpicks:

### `src/dbt_client/dbtWorkspaceFolder.ts:88-144` — *defensive `lastEmpty ?? []`* — **applied** in latest force-push

Pushed as `(lastEmpty ?? []) as T;`. Tiny defensive change, removes the theoretical `retries <= 0` footgun where the loop body never executes. The private callers all pass `retries = 5` today so it's not a real path, but the cast was technically unsound and this costs nothing.

### `src/test/suite/discoverProjectsRepro.test.ts:79-155` — *fake timers to avoid 40s overhead* — **skipping**

Three of the four cases would run instantly with fake timers, but the second test (`retry budget is still ~10s`) genuinely needs real-clock timing — that's its whole assertion. Mixing fake/real timers across cases in the same file is messier than uniform real timers, and 40s for an integration-shape regression test that exercises a real backoff schedule end-to-end against the actual `DBTWorkspaceFolder` source is a deliberate trade. The point of these tests is to prove the timing contract (10s window unchanged) and the post-loop disposition (`return [] vs throw`); shrinking three of them to mock-time tests would weaken that.

### `src/test/suite/discoverProjectsRepro.test.ts:108-109` — *upper bound 12000ms tight on throttled CI* — **deferring**

Reasonable concern, but the `[9500, 12000)` window is fitting cleanly in CI today across all three platform runners (mac / linux / win). Loosening the upper bound to 15000+ reduces the assertion's value (it stops tightly verifying "this is actually a 10-second schedule"). I'd rather wait for an actual flaky run before relaxing — happy to bump it then.

src/dbt_client/dbtWorkspaceFolder.ts
```diff
@@ -91,32 +91,60 @@ export class DBTWorkspaceFolder implements Disposable {
...
-      "no projects found after maximum retries",
+      "no projects found after maximum retries — treating folder as empty",
       false,
     );
-    throw new Error("no projects found after maximum retries");
+    // Defensive `?? []` covers the theoretical `retries <= 0` case where the
+    // loop never runs and `lastEmpty` stays undefined. The private callers
+    // only pass `retries = 5` today, but keeping the cast sound costs
+    // nothing.
+    return (lastEmpty ?? []) as T;
```
mdesmet:
```suggestion
    return [];
```

ralphstodomingo: Applied — dropped the `lastEmpty` tracking entirely. The post-loop now just `return [] as T;`. The cast is needed because TypeScript can't statically prove `T extends Array` from the runtime `Array.isArray` branch, but the function is private with one caller (`workspace.findFiles` returning `Uri[]`) so it's sound. Pushed in the latest force-push.

src/dbt_client/dbtWorkspaceFolder.ts
mdesmet: Can we not just add try catch here?

mdesmet: And revert above change?

dev-punia-altimate:
## ✅ Tests — All Passed



cc @ralphstodomingo

src/dbt_client/dbtWorkspaceFolder.ts
```diff
@@ -116,22 +121,35 @@ export class DBTWorkspaceFolder implements Disposable {
...
+          workspace.findFiles(
+            new RelativePattern(this.workspaceFolder, `**/${DBT_PROJECT_FILE}`),
+            new RelativePattern(this.workspaceFolder, excludePattern),
+          ),
+        5,
+        1000,
+      );
+    } catch (error) {
+      if (!(error instanceof NoProjectsFound)) {
+        this.telemetry.sendTelemetryError("discoverProjectsError", error);
```
mdesmet, line 149: This is not really an error that we need to track though. this just means that there is no projects found?

ralphstodomingo: @mdesmet not the case.

`!(error instanceof NoProjectsFound)` means it's deliberately ignoring no projects found here

there was an activation-level catch that was surfacing all of these, both the 'no projects found' and everything else findFiles could reject for (EMFILE, EACCES, etc)

this PR makes it so it no longer throws an error that's caught there when it's just no dbt projects found, but without this, errors that the old structure would have caught at activation level would just be silently ignored

mdesmet (approved):

Issue #98, opened 2022-01-26
"Missing profile" error when opening a workspace


ljhopkins2:
I've just started having an issue with the extension (v0.5.9). I recently upgraded to dbt v1.0.1 from v0.21.
```
dbt version: 1.0.1
python version: 3.7.7
python path: /Users/<me>/Documents/GitHub/data-warehouse/.venv/bin/python3
os info: Darwin-21.1.0-x86_64-i386-64bit
Using profiles.yml file at /Users/<me>/.dbt/profiles.yml
Using dbt_project.yml file at /Users/<me>/Documents/GitHub/data-warehouse/dbt_project.yml

Configuration:
  profiles.yml file [OK found and valid]
  dbt_project.yml file [OK found and valid]

Required dependencies:
 - git [OK found]
```
I've confirmed the python path above is the one being used by the VS Code workspace.

When first opening a workspace with my dbt project in it, the `Tasks - dbt` terminal displays the following:
```
> Executing task:  dbt list

> Executing task:  dbt list

> Executing task:  dbt list

07:53:31  Encountered an error while reading the project:
07:53:31    ERROR: Runtime Error
  dbt cannot run because no profile was specified for this dbt project.
  To specify a profile for this project, add a line like the this to
  your dbt_project.yml file:
  
  profile: [profile name]
  
  Here, [profile name] should be replaced with a profile name
  defined in your profiles.yml file. You can find profiles.yml here:
  
  /Users/<me>/.dbt/profiles.yml
  
07:53:31  Encountered an error:
Runtime Error
  Could not run dbt
07:53:31  Encountered an error while reading the project:
07:53:31    ERROR: Runtime Error
  Could not find profile named 'integration_tests'
07:53:31  Encountered an error:
Runtime Error
  Could not run dbt

[dbt listing of models begins here...]
```

From that moment on, automatic listing of models doesn't occur when I save a `.sql` file.

bdtoole:
I'm going through the upgrade right now, and I've noticed the same issue. Do you happen to have the `calogica/dbt_date` package in your environment? I noticed the dbt_project.yml file within that package references an `integration_tests` profile, and I'm wondering if there's some compatibility issue between this extension, the package, and the latest version (perhaps due to the default `dbt_modules` path being changed to `dbt_packages`?

```
name: 'dbt_date'
version: '0.5.0'

config-version: 2

target-path: "target"
clean-targets: ["target", "dbt_packages"]
[...]

Issue #64, opened 2021-04-15
Cannot use extension


dreso:
After updating to dbt 0.19 I cannot check Parent and Children models using the extension :/

mdesmet:
Could you provide more details? Which platform? Any errors in log (You can check the logs through https://stackoverflow.com/questions/30765782/what-is-the-use-of-the-developer-tools-in-vs-code#:~:text=2%20Answers&text=When%20you%20select%20the%20option,built%20with%20HTML%20and%20Javascript.)

dreso:
You mean the following? 
<img width="1239" alt="Screenshot 2021-04-16 at 10 53 52" src="https://user-images.githubusercontent.com/52713281/114999695-61584f00-9ea2-11eb-88fb-66658152dd8a.png">

shr1k:
I'm facing the same issue -- it looks like the extension fails to load, as I can't see it in my status bar at all. Here's a screenshot of the error in the dev console:

![image](https://user-images.githubusercontent.com/927606/115411690-98549a80-a1eb-11eb-9d83-f93cdab3149f.png)

mdesmet:
It's definitely a different error. In your case he extension is not able to be activated. If you can provide me with a minimal project, I can try to reproduce the issue.

mdesmet:
@shr1k : Your issue should be resolved with 0.5.3.

@dreso: Do you still have this issue? Does it happen in every project?

shr1k:
@mdesmet that was quick! :smile: it works now, thanks

dreso:
Yes still have it. It is the same for all projects unfortunately

haf:
Same here, nothing works. It starts and then gives *many* errors in the dev console but nothing actionable.

mdesmet:
Can you post the errors here?

haf:
It's a lot; like file not found errors, but it just started working in part.

<img width="934" alt="image" src="https://user-images.githubusercontent.com/193115/131970025-ffbae427-b353-4a99-99b9-b7403eabc345.png">

mdesmet:
Can you guyst retest? Some of the issues look a lot like #57. Should be fixed with latest version.

PR #1940, opened 2026-05-07
feat: route unhandled rejections through sendTelemetryError for richer error data

ralphstodomingo:
## Summary

Production cluster `unhandlederror` with `Channel closed` (**~45 unique machines on 0.61.x**, mostly Win32) is opaque in App Insights even after the `error.name` / `error.message` / `error.code` visibility fix in #1928 — that fix only applies to events emitted via our `sendTelemetryError` helper. Uncaught promise rejections fire VS Code's built-in `unhandlederror` event from `@vscode/extension-telemetry`'s `baseTelemetrySender.sendErrorData`, which only forwards `name` / `message` / `stack` and **drops `error.code`**.

`error.code` is the field that distinguishes IPC failure modes (`ERR_IPC_CHANNEL_CLOSED`, `EPIPE`, `EBADF`, `ECONNRESET`, ...). Without it, every `Channel closed` event looks identical in App Insights regardless of root cause — making it impossible to triage which subsystem the IPC channel was attached to.

## What this PR does

Adds a `process.on('unhandledRejection')` listener in `DBTPowerUserExtension.activate(context)` that routes the rejection reason through `this.telemetry.sendTelemetryError("unhandledRejection", reason)`. The helper already handles `error_name` / `error_message` / `error_code` extraction (`telemetry/index.ts:128-133`) plus the standard `dbtIntegrationMode` / `instanceName` / `localMode` envelope.

```ts
const onUnhandledRejection = (reason: unknown) => {
  try {
    this.telemetry.sendTelemetryError("unhandledRejection", reason);
  } catch {
    // Telemetry failures must never re-enter the rejection path.
  }
};
process.on("unhandledRejection", onUnhandledRejection);
context.subscriptions.push({
  dispose: () => process.off("unhandledRejection", onUnhandledRejection),
});
```

Three deliberate properties:

1. **Distinct event name.** `unhandledRejection` (camelCase) vs upstream `unhandlederror` (lowercase) — both events stream to App Insights separately, so the new one is queryable independently and the upstream stream stays untouched. Lets us deprecate using `unhandlederror` later without breaking existing dashboards.
2. **Disposable cleanup.** Registered against `context.subscriptions` so the listener unhooks on extension deactivation. Without this, handlers accumulate one-per-reactivation across the lifetime of a VS Code window.
3. **Try/catch around the telemetry call.** A throwing telemetry pipeline must not re-enter the rejection path — that would surface as another unhandled rejection, caught by our handler, throwing again, infinite loop.

## What this enables

The next telemetry rotation will give us `error_code` on cluster `unhandledRejection`, which means we can split "Channel closed" into:
- `ERR_IPC_CHANNEL_CLOSED` — node child_process IPC torn down (this is the one tied to PR #98's atomic-swap race in `altimate-dbt-integration`)
- `EPIPE` — write to a closed pipe
- `EBADF` — bad file descriptor (file-handle leak, fixed in earlier PR #64 but worth verifying)
- ... whatever else surfaces

Today these are all anonymous `Error: Channel closed` rows. Post-fix they become individually triageable.

## Test plan

Adds `src/test/suite/unhandledRejectionRouting.test.ts` with four cases:

```
PASS src/test/suite/unhandledRejectionRouting.test.ts
  unhandledRejection routing
    ✓ forwards an uncaught Promise rejection through sendTelemetryError (3 ms)
    ✓ captures error.code via sendTelemetryError's existing extractErrorFields
    ✓ dispose unhooks the listener so handlers don't accumulate across reactivations
    ✓ a throwing telemetry call does not re-enter the rejection path
```

- [x] `npx jest` — 32 of 33 suites pass; only failure is the pre-existing `jupyterlabServicesCompat` environmental issue (missing build artifact) confirmed unrelated by stashing this PR's changes
- [x] No public API surface change
- [x] Listener registered as `context.subscriptions` disposable — verified by test 3 (no calls after dispose)
- [x] Tests drive `process.emit` directly because Jest's own `unhandledRejection` listener for failure reporting interferes with letting `Promise.reject` bubble naturally — emitting the event is identical from the listener's perspective

## Summary by CodeRabbit

* **Improvements**
  * Capture unhandled promise rejections and route them to telemetry for improved diagnostics.
  * Ensure telemetry failures are safely contained so they don’t trigger additional unhandled-rejection issues.

* **Tests**
  * Added tests validating routing of unhandled rejections to telemetry, preservation of error payload details, and proper listener cleanup and disposal.

### DIFF
--- a/src/dbtPowerUserExtension.ts
+++ b/src/dbtPowerUserExtension.ts
@@ -108,7 +108,7 @@ export class DBTPowerUserExtension implements Disposable {
       // parallel — both events stream to App Insights, queryable separately.
       const onUnhandledRejection = (reason: unknown) => {
         try {
-          this.telemetry.sendTelemetryError("unhandledRejection", reason);
+          this.telemetry.sendTelemetryError("catchAllError", reason);
         } catch {
           // Telemetry failures must never re-enter the rejection path.
         }
--- a/src/dbtPowerUserExtension.ts
+++ b/src/dbtPowerUserExtension.ts
@@ -93,6 +93,31 @@ export class DBTPowerUserExtension implements Disposable {
 
   async activate(context: ExtensionContext): Promise<void> {
     try {
+      // VS Code's `@vscode/extension-telemetry` library auto-emits an
+      // `unhandlederror` event for uncaught promise rejections, but it only
+      // captures `name`/`message`/`stack` (baseTelemetrySender.sendErrorData)
+      // and skips `error.code`. That makes IPC failures like `Channel
+      // closed` (errno `ERR_IPC_CHANNEL_CLOSED`), `EPIPE`, `EBADF`, etc.
+      // indistinguishable from each other in App Insights — all just show
+      // up with `name="Error"` and `message="Channel closed"`.
+      //
+      // Route uncaught rejections through `sendTelemetryError` in addition,
+      // so they pick up our consistent `error_name` / `error_message` /
+      // `error_code` fields plus `dbtIntegrationMode` / `instanceName` /
+      // `localMode`. The upstream `unhandlederror` event keeps firing in
+      // parallel — both events stream to App Insights, queryable separately.
+      const onUnhandledRejection = (reason: unknown) => {
+        try {
+          this.telemetry.sendTelemetryError("unhandledRejection", reason);
+        } catch {
+          // Telemetry failures must never re-enter the rejection path.
+        }
+      };
+      process.on("unhandledRejection", onUnhandledRejection);
+      context.subscriptions.push({
+        dispose: () => process.off("unhandledRejection", onUnhandledRejection),
+      });
+
       await this.mcpServer.updateMcpExtensionApi();
       this.dbtProjectContainer.setContext(context);
       this.dbtProjectContainer.initializeWalkthrough();
--- a/src/test/suite/unhandledRejectionRouting.test.ts
+++ b/src/test/suite/unhandledRejectionRouting.test.ts
@@ -30,7 +30,7 @@ function attachUnhandledRejectionRouting(
 ): (reason: unknown) => void {
   const onUnhandledRejection = (reason: unknown) => {
     try {
-      telemetry.sendTelemetryError("unhandledRejection", reason);
+      telemetry.sendTelemetryError("catchAllError", reason);
     } catch {
       // Telemetry failures must never re-enter the rejection path.
     }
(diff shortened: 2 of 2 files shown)


src/test/suite/unhandledRejectionRouting.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        "unhandledRejection",
+        new Error("triggering"),
+        Promise.resolve(),
+      ),
+    ).not.toThrow();
+
+    expect(telemetry.sendTelemetryError).toHaveBeenCalledTimes(1);
+
+    context.subscriptions.forEach(d => d.dispose());
+  });
```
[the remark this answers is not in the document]
ralphstodomingo: Applied — moved cleanup into a shared `afterEach` that disposes whatever `currentContext` was last set to, so a failing assertion in any test still unhooks the listener.

Each test now does `currentContext = context;` before `attachUnhandledRejectionRouting(...)`, and the `afterEach` does `currentContext?.subscriptions.forEach((d) => d.dispose())`. Verified all four tests still pass post-refactor (4.8s suite time, no `MaxListenersExceededWarning`).

Pushed as part of the latest force-push.

src/dbtPowerUserExtension.ts
```diff
@@ -93,6 +93,31 @@ export class DBTPowerUserExtension implements Disposable {
...
+      // up with `name="Error"` and `message="Channel closed"`.
+      //
+      // Route uncaught rejections through `sendTelemetryError` in addition,
+      // so they pick up our consistent `error_name` / `error_message` /
+      // `error_code` fields plus `dbtIntegrationMode` / `instanceName` /
+      // `localMode`. The upstream `unhandlederror` event keeps firing in
+      // parallel — both events stream to App Insights, queryable separately.
+      const onUnhandledRejection = (reason: unknown) => {
+        try {
+          this.telemetry.sendTelemetryError("unhandledRejection", reason);
```
mdesmet:
```suggestion
          this.telemetry.sendTelemetryError("catchAllError", reason);
```

ralphstodomingo: Applied in 0c4e084f — renamed to `catchAllError` in `dbtPowerUserExtension.ts:111` and updated the matching assertions in `unhandledRejectionRouting.test.ts`. Picked the convention up as a follow-up — opening a separate PR with a custom ESLint rule that enforces the `-Error` suffix on `sendTelemetryError`'s first arg, so the next non-conforming call gets caught at lint time rather than review.

dev-punia-altimate:
## ✅ Tests — All Passed



cc @ralphstodomingo

mdesmet (approved):

PR #1953, opened 2026-05-18
fix: drop unused nonExistingAllowListFolders telemetry, fix warn condition

ralphstodomingo:
## Summary

`DBTWorkspaceFolder.getAllowListFolders` warns to the dev-tools console when it filters out `dbt.allowListFolders` entries that don't exist on disk. The condition on that warn was inverted since commit `e7220cb6` (Dec 2023):

```ts
if (nonFilteredAllowListFolders.length === allowListFolders.length) {
  console.warn("filtered out non-existing allowListFolders", ...);
  this.telemetry.sendTelemetryEvent("nonExistingAllowListFolders");
}
```

That fires when **nothing** has been filtered out — opposite of what the warning message says. The same condition also gated a `sendTelemetryEvent` call.

### 1. Flip the warn condition `===` → `!==`

The console.warn message says "filtered out non-existing allowListFolders" but only fires when the lengths are equal — i.e., when filtering didn't happen. The intended diagnostic signal is "some configured paths are missing on disk," which is the `!==` case.

Behavior change matrix:

| Configured | After existence filter | Old warn | New warn |
|---|---|---|---|
| `[]` | `[]` | fires (wrong) | doesn't fire |
| `[a, b, c]` (all exist) | `[a, b, c]` | fires (wrong) | doesn't fire |
| `[a, b, missing]` | `[a, b]` | doesn't fire (wrong) | **fires** |
| `[missing1, missing2]` | `[]` | doesn't fire (wrong) | **fires** |

### 2. Remove the `nonExistingAllowListFolders` telemetry emission

This is a bare-name event (`sendTelemetryEvent`, not `sendTelemetryError`) — it doesn't appear in any `*Error` dashboard or alert. It has no current consumer: the inverted-condition bug went unnoticed for 18 months, which is the strongest possible evidence that nobody mines this signal.

Production data on 0.61.3 showed it firing for **756 machines / 24h** — 40% of all users on the latest version — almost entirely the empty-config case (both arrays length 0). Even after the condition fix, the corrected event would carry no payload (no path list, no count, no project context), so any future triage would still need to follow up with the user to learn anything actionable.

The right surface for this diagnostic is **the user's own UI** (a one-time notification pointing at the misconfigured paths in their setting), not our telemetry. That UX can be designed when there's actual demand. Until then, removing the dead emission is simpler than keeping a corrected-but-unused event.

The `console.warn` survives for dev-mode self-diagnosis (no behavior change there).

### Naming convention context

Confirmed against the codebase: `sendTelemetryEvent` events are bare verbs / namespaced (`runModel`, `installDeps`, `altimateScan:Done`), while `sendTelemetryError` events always end in `Error` (`catchAllError`, `extensionActivationError`, etc.). One-to-one mapping. `nonExistingAllowListFolders` was correctly bare (not in error dashboards) — just unused.

Also fixes the long-standing variable-name typo (`nonFilteredAlolowListFolders` → `nonFilteredAllowListFolders`) on the lines being touched.

## Test plan

No unit test added — the predicate is one operator flip gating a `console.warn`, not load-bearing logic.

- [x] Full jest suite green (488 passed / 1 skipped, no regressions)
- [x] ESLint clean on modified file
- [x] Bundle builds via `rsbuild build --mode development` and confirmed `nonExistingAllowListFolders` symbol removed (grep count = 0)

**Post-merge expectation**: the ~9,151 events / 756 machines / 24h of `nonExistingAllowListFolders` simply disappear from App Insights. Users with genuinely-misconfigured paths still see the console.warn in their dev tools if they look.

## Release Notes

* **Bug Fixes**
  * Fixed dbt allowlist folder validation logic to properly check folder existence on the filesystem and display clearer warnings when configured folders are not found, helping users identify configuration issues.

[Review Change Stack](https://app.coderabbit.ai/change-stack/AltimateAI/vscode-dbt-power-user/pull/1953?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


dev-punia-altimate:
## ✅ Tests — All Passed



cc @ralphstodomingo

saravmajestic (approved):

PR #1968, opened 2026-05-21
fix: CTE profiler returns 0 rows on Snowflake-like adapters

ralphstodomingo:
## Summary

- "Profile CTEs" reports 0 rows for every CTE on Snowflake (and Oracle) even when the same CTE clearly returns data via the "Execute query" codelens. Root cause: the profiler emits `SELECT COUNT(*) AS _profile_count FROM <cte>` with an **unquoted** alias, which Snowflake folds to UPPERCASE. The adapter zips `column_names` verbatim into the row dict, so the row arrives as `{ _PROFILE_COUNT: n }` and the previous `data[0]["_profile_count"]` lookup misses → falls through to 0.
- Fix is one line: read the COUNT result by position via `Object.values(data[0])[0]`. The profiler builds the query itself, knows it's a single-column COUNT, and the positional read sidesteps every dialect's identifier folding rule. Empty-result and non-numeric safety preserved.
- Postgres, Redshift, DuckDB, and BigQuery fold to lowercase / preserve case, so the lookup happened to work there — which is why this slipped through dev and CI.

## Test plan

- [x] New unit test `src/test/suite/cteProfilerService.test.ts` exercises:
  - uppercase Snowflake/Oracle row shape (`{ _PROFILE_COUNT: 99 }`) — the case that was returning 0 before
  - lowercase Postgres/DuckDB row shape (regression guard)
  - BigInteger-as-string coercion (some adapters stringify long counts)
  - empty rows (returns 0)
  - non-numeric values (returns 0)
- [x] `npx tsc --noEmit` clean
- [x] `npx eslint` clean on the changed files
- [ ] Manual: load a Snowflake-backed dbt project (jaffle_shop_core or equivalent), open a model with multiple CTEs, run "Profile CTEs" codelens, confirm the decorations now show real row counts instead of 0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where CTE row counts were incorrectly displayed as zero, which could impact query profiling accuracy.

* **Tests**
  * Added test coverage for row count extraction to improve reliability across different database systems.

[Review Change Stack](https://app.coderabbit.ai/change-stack/AltimateAI/vscode-dbt-power-user/pull/1968?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/cte_profiler/cteProfilerService.ts
+++ b/src/cte_profiler/cteProfilerService.ts
@@ -271,7 +271,12 @@ export class CteProfilerService implements Disposable {
     if (data.length === 0) {
       return 0;
     }
-    const count = data[0]["_profile_count"];
+    // Read the single COUNT(*) value by position rather than by name. The
+    // profiler builds the query itself and aliases the result, but adapters
+    // fold the alias according to their dialect's identifier rules — Snowflake
+    // and Oracle uppercase unquoted aliases, so a lowercase key lookup
+    // misses the row and the profiler reports 0 for every CTE.
+    const count = Object.values(data[0])[0];
     return typeof count === "number" ? count : Number(count) || 0;
   }
 
--- a/src/test/suite/cteProfilerService.test.ts
+++ b/src/test/suite/cteProfilerService.test.ts
@@ -1,0 +1,54 @@
+import { DBTTerminal } from "@altimateai/dbt-integration";
+import { beforeEach, describe, expect, it } from "@jest/globals";
+import { CteProfilerService } from "../../cte_profiler/cteProfilerService";
+import { DBTProjectContainer } from "../../dbt_client/dbtProjectContainer";
+
+describe("CteProfilerService.extractRowCount", () => {
+  let svc: CteProfilerService;
+
+  beforeEach(() => {
+    const dbtTerminal = {
+      debug: jest.fn(),
+      warn: jest.fn(),
+      error: jest.fn(),
+      info: jest.fn(),
+      log: jest.fn(),
+      trace: jest.fn(),
+    } as unknown as DBTTerminal;
+    const dbtProjectContainer = {} as DBTProjectContainer;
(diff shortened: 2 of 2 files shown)


ralphstodomingo:
## E2E verification on Snowflake (master → fix branch)

Reproduced both pre- and post-fix in code-server docker against a real Snowflake-backed dbt project — `jaffle_shop_core` with private-key auth, populated via `dbt seed && dbt run` (99 orders, 113 payments materialized in `js_dbt`).

### Before (master, `5e2d1c73`)

Click the **⏱ Profile CTEs** codelens on `orders.sql`:

before

Decorations:
- `with orders as (    ⏱ 4.5s · 0 rows`
- `payments as (    ⏱ 0ms · 0 rows`
- `order_payments as (    ⏱ 846ms · 0 rows`

All zero — despite the staging views having real data. Confirmed independently by clicking **Execute CTE: orders** on the same CTE; that path renders the actual rows in the query panel.

### After (this PR, built together with the dbt-integration sibling PR for the unrelated Validate SQL bug)

Same click on the same file, same Snowflake tables:

after

Decorations:
- `with orders as (    ⏱ 4.5s · 99 rows`
- `payments as (    ⏱ 0ms · 113 rows`
- `order_payments as (    ⏱ 30ms · 99 rows`

Real row counts from Snowflake, matching what `dbt run` materialized.

### Test plan
- [x] New unit test `src/test/suite/cteProfilerService.test.ts` — 5/5 pass (uppercase Snowflake row, lowercase Postgres row, BigInteger-as-string string-coercion, empty rows, non-numeric)
- [x] `npx tsc --noEmit` clean on the changed files
- [x] `npx eslint` clean on the changed files
- [x] Manual verification in the extension against a real Snowflake project via code-server docker

dev-punia-altimate:
## ✅ Tests — All Passed



cc @ralphstodomingo

mdesmet (approved):

Issue #111, opened 2022-05-05
dbt shows connected but extension has zero functionality


elctrc-rhino:
After installing dbt-power-user and confirming dbt is connected, none of the functionality of the extension is working. Nothing populates in the parent/child models area and none of my models link out to definitions, nor does autocomplete work. 

I believe this is most closely related to the following issues: [Editor functionality not working](https://github.com/innoverio/vscode-dbt-power-user/issues/57) and [Cannot read manifest](https://github.com/innoverio/vscode-dbt-power-user/issues/81). I have updated to the latest version of both VS Code and dbt-power-user. I have tried removing other plugins. I have tried opening the base dbt jaffleshop code but have the same issue. The console seems to be showing errors in the dbt code but I don't know that it's related (based on the other tickets). An example of one of the console errors:

```File not found at '/dbt/macros/generate_schema_name.sql', probably compiled is outdated! Error: ENOENT: no such file or directory, open '/dbt/macros/generate_schema_name.sql'```

In the above case, the file is definitely there, and works as expected and does not give any errors in dbt. But that is just one of what look like dozens of similar errors. It seems like possibly a red herring. When I just open the base jaffleshop code, for example, it just gives different errors, like missing the "manifest.json" file - again, perhaps just a red herring.

```
System info:
Mac OS Monterey 12.3.1
VS Code 1.66.2
dbt Power User v0.5.16
```

mdesmet:
Are you using a docker container setup where the path in your container doesn't match with the filesystem's path? 

In that case you should setup your vscode to [open the workspace in the container](https://code.visualstudio.com/docs/remote/containers#_open-an-existing-workspace-in-a-container).

krvrd:
I don't use Docker container for a project but am stuck with the same issue. I can't surf through models ref and I don't see any information in a dbt-power-user tab. 

MacOS 12
dbt v.1.1.1 installed by homebrew
vscod shows that dbt connected

korompaiistvan:
did y'all associate sql files with jinja-sql? I had this issue when I forgot that step

elctrc-rhino:
Yup. Not using a Docker container and I have associated the files with jinja-sql.

mdesmet:
Do you still have the same issue with latest version? If so please reopen the issue.

PR #1905, opened 2026-04-21
feat: ERD overlay — getRelationships extension command + source-pointing

ralphstodomingo:
## Summary

Power-user side of the lineage ERD overlay feature. Wires the new webview canvas (AltimateAI/altimate-frontend#2555) to dbt-integration's `RelationshipParser` (AltimateAI/altimate-dbt-integration#62) via the `getRelationships` command, threading all four ref sources through and gating the whole feature behind a **validated** Altimate API key (`ValidationProvider.isAuthenticated()`).

While here, also tightens the lineage panel's existing `aiEnabled` flag: `getStartingNode` now reports `aiEnabled = validationProvider.isAuthenticated()` instead of `altimate.enabled()` (which only checks that the `altimateAiKey` + `altimateInstanceName` settings are non-empty strings). Other lineage features that gate on `aiEnabled` (e.g. AI Settings sections) inherit the stricter semantics.

## What's in

- **`src/webview_provider/newLineagePanel.ts`** — `getRelationships` handler chains all four ref sources:
  - `parser.fromTests(testMetaMap)` — Phase 1, dbt `relationships` data tests
  - `parser.fromContracts(nodeMetaMap, sourceMetaMap)` — Phase 2, model-contract `foreign_key` constraints
  - `parser.fromInference(nodeMetaMap, sourceMetaMap, options)` — Phase 3, naming-convention inference (sources / self-references opt-in via params; parser confidence floor 0.6)
  - `parser.fromSemanticEntities(semanticModelMetaMap)` — Phase 4, semantic-layer foreign↔primary entity pairings
  Early-returns `{ refs: [] }` when `validationProvider.isAuthenticated()` is false. Defense-in-depth alongside the frontend's `aiEnabled` check.
- **`getStartingNode`** switched from `altimate.enabled()` to `validationProvider.isAuthenticated()` for the `aiEnabled` field in the render payload.
- **`ValidationProvider`** added to the `NewLineagePanel` constructor; DI binding in `inversify.config.ts` updated.
- **`webview_panels/src/modules/lineage/LineageView.tsx`** — registers `getRelationships` in the `ApiHelper.get` switch.
- **`webview_panels/src/modules/lineage/ActionWidget.tsx`** — `<div id="refs-container" />` between `#export-container` and `#settings-container`. The frontend's `RefsToggleButton` portals into this when `aiEnabled`.

### Manifest cache event extension
[...]

Issue #62, opened 2021-04-09
Extension does not work with 2 dbt projects in the same workspace


Elektry-On:
Hi All!

My workspace structure is as follows:
![image](https://user-images.githubusercontent.com/62945720/114211208-a2130e00-9960-11eb-8153-276c6d77881d.png)

While the extension works properly in the first project ("DBTTransform"):
![image](https://user-images.githubusercontent.com/62945720/114211650-2ebdcc00-9961-11eb-88e6-418ce031f489.png)

It seems to not work in the second project ("DBTTransform_Synapse"):
![image](https://user-images.githubusercontent.com/62945720/114211606-22397380-9961-11eb-8c92-43206adc5e06.png)

Extension installed version: v0.5.1
DBT installed version: 0.19.0

Otherwise it's a great extension, thanks a lot for the hard work!
Sven

Elektry-On:
Looking at the Tasks - dbt output console it looks like it does both 2 dbt list commands for both projects.

Elektry-On:
I tested different orderings of the folder paths in the code-workspace settings with an interesting outcome. If I move the "DBTTransform_Synapse" folder at the first position and the "DBTTransform" at second position like this:
![image](https://user-images.githubusercontent.com/62945720/114213118-e9020300-9962-11eb-83b2-f312df8fd08b.png)
It does work.

With this ordering:
![image](https://user-images.githubusercontent.com/62945720/114213312-1b136500-9963-11eb-9eaf-8279b6f84ca6.png)
It does NOT work

mdesmet:
That's interesting. Could you setup a project that i can reproduce the issue?

Elektry-On:
This should do. You should see the extension working in the first project but not in the second.
[PowerDBT.zip](https://github.com/innoverio/vscode-dbt-power-user/files/6289635/PowerDBT.zip)

mdesmet:
Thanks, I've found the error and fixed it in v0.5.2. Let me know if it works now!

If you get value from this extension please rate the extension on https://marketplace.visualstudio.com/items?itemName=innoverio.vscode-dbt-power-user

Elektry-On:
I tested it with version 0.5.2 and it worked. Thanks a lot for the quick support!

PR #1981, opened 2026-05-27
feat: hide foreign-key-only edges in model lineage

ralphstodomingo:
> [!IMPORTANT]
> **This bug has two alternative fixes. Ship one, close the other.** Both produce the same lineage-panel result; they differ only in blast radius.
>
> - **Approach A (global filter):** `AltimateAI/altimate-dbt-integration#107` drops the edges inside the shared dbt-integration parser. Simplest, but changes the dependency graph for *every* consumer (Validate SQL, the AI impact-analysis tool, depth), not just the panel.
> - **Approach B (annotate + scope, recommended):** `AltimateAI/altimate-dbt-integration#111` (library) tags the edges and this PR hides them only in the lineage panel. Other consumers keep the full graph.
>
> **This PR is the extension half of Approach B.**

## Problem

The model lineage panel draws an upstream edge for any table referenced by a declared foreign-key constraint (for example `to: ref('dim_contact')`), even when the model's SQL never selects from that table. dbt records FK targets in `depends_on.nodes` for build ordering, so they are indistinguishable from real data-flow dependencies in the graph. Result: dimension tables appear as upstream sources of a fact even though no data flows from them.

## Change

`@altimateai/dbt-integration` now tags each lineage edge with `edgeType: "data" | "constraint"` (`AltimateAI/altimate-dbt-integration#111`). This PR makes the model-lineage view honor that tag:

- `DbtLineageService.getConnectedTables` drops edges with `edgeType === "constraint"`.
- `getConnectedNodeCount` excludes them so the upstream / downstream counts match what is drawn.

## Scope of effect

Scoped deliberately to the lineage service. The shared `graphMetaMap` is untouched, so:

- Validate SQL (`getNonEphemeralParents`) still fetches the complete upstream set
- the MCP `getChildrenModels` impact-analysis tool still sees FK relationships
- model depth is unchanged
- column lineage is unaffected (it derives edges from compiled SQL via sqlglot, not from `depends_on`)

Only the model lineage panel changes. This is the difference from Approach A (`AltimateAI/altimate-dbt-integration#107`), which removes the edges from the shared graph for all consumers.

## Verification

- `tsc -p ./ --noEmit` clean, `eslint` clean, **4 new jest unit tests** (upstream / downstream hiding, untagged edge treated as data-flow, count exclusion).
- E2E in code-server (docker), real dbt-duckdb parse. Fixture `fct_fk_demo` reads only `customers` and declares an FK to `orders`:

| Before (baseline extension) | After (this PR) |
|---|---|
| before | after |
| `customers` and `orders` upstream of `fct_fk_demo` | only `customers`, the FK-only `orders` edge is hidden |

## Sequencing

This PR consumes the `edgeType` field added in `AltimateAI/altimate-dbt-integration#111`, so it depends on that change being merged and published. Until then CI typecheck here fails against the released `@altimateai/dbt-integration` (`Property 'edgeType' does not exist on type 'NodeData'`). Order: merge and publish #111, bump the dependency here, then this goes green. Verified locally against a build of #111.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined lineage visualization to exclude foreign-key constraint edges, displaying only direct data-flow relationships in upstream and downstream table views. Lineage counts now accurately reflect filtered connections.

* **Tests**
  * Added comprehensive test suite validating constraint edge filtering in lineage traversal.

* **Chores**
  * Updated dependency version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,16 @@
 {
   "name": "vscode-dbt-power-user",
-  "version": "0.61.4",
+  "version": "0.61.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "vscode-dbt-power-user",
-      "version": "0.61.4",
+      "version": "0.61.6",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
-        "@altimateai/dbt-integration": "^0.3.2",
+        "@altimateai/dbt-integration": "^0.3.3",
         "@jupyterlab/coreutils": "^6.2.4",
         "@jupyterlab/nbformat": "^4.2.4",
         "@jupyterlab/services": "^7.0.0",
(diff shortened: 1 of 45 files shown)


ralphstodomingo:
## E2E verification — published dbt-integration 0.3.3

Verified in Docker code-server against the **published** `@altimateai/dbt-integration@0.3.3` (which carries the `edgeType` annotation this PR reads). Fixture: a `fct_fk_demo` model in `jaffle-shop-duckdb` that **reads only `customers` in SQL** but declares a **foreign-key constraint to `orders`** (which it never selects from). Manifest confirms the shape — `depends_on.nodes = [customers, orders]`, with `orders` present purely via the FK constraint.

### Panel before/after
| Before (FK-only edge shown — the bug) | After (FK-only edge hidden) |
|---|---|
| before | after |

- ✅ **Before** (filter disabled): `fct_fk_demo` shows **both** `customers` and `orders` upstream.
- ✅ **After** (this PR): `fct_fk_demo` shows **only `customers`** — the FK-only `orders` edge is hidden, even though it remains in `depends_on`.

### Scope preserved (the point of annotate-and-scope)
This PR's only source change is `src/services/dbtLineageService.ts` (+ unit tests). `validateSql.ts`, `dbtProject.getNonEphemeralParents` (Validate SQL upstream), and the MCP impact-analysis paths are **untouched** — they read the unmodified graph, so the FK edge is still available to non-panel consumers. The hiding is local to the lineage panel.

### ⚠️ Build/merge-order dependency
Published `0.3.3` bundles **all three** dbt-integration PRs — including #62's `RelationshipParser`, which adds a required `ParsedManifest.semanticModelMetaMap` field and a new constructor arg. The consumer-side adaptations for those live in **#1905**, not here. So **this branch does not compile against published 0.3.3 on its own** (TS2741 `semanticModelMetaMap` missing / TS2554 arg count). The above E2E was run on the **post-merge state** (this branch + #1905 merged locally), which builds clean and is what `master` will look like once both land.

**Recommended merge order: #1905 first, then rebase this PR on `master`** (its own `0.3.3` bump then becomes a no-op and it carries only the lineage-service filter).

dev-punia-altimate:
## ✅ Tests — All Passed



cc @ralphstodomingo

PR #1982, opened 2026-05-27
fix: detect-from-terminal no longer writes probe fragment to dbtPythonPathOverride

ralphstodomingo:
## Problem

The **Detect Python from terminal** command could write a fragment of its own probe command into `dbt.dbtPythonPathOverride`, after which every dbt/pip invocation failed with `Command not found: "); print(sys.executable); print("`.

`detectPythonFromShell` runs a Python one-liner in the integrated terminal via the shell-integration API and recovers `sys.executable` by searching the captured output for unique markers:

```
python3 -c 'import sys; __import__("dbt"); print("__DBT_DETECT_START__"); print(sys.executable); print("__DBT_DETECT_END__")'
```

VS Code shell integration echoes the executed command line into the **same stream** that `executeInTerminal` reads back. Because the markers were embedded as literal strings, `indexOf("__DBT_DETECT_START__")` / `indexOf("__DBT_DETECT_END__")` matched the **echoed command** instead of stdout, and the extracted "path" was the code between them:

```
"); print(sys.executable); print("
```

That value was then offered as the interpreter ("Found Python with dbt at: …") and persisted to `dbt.dbtPythonPathOverride`. This surfaced most visibly with a Homebrew Python (which also can't `pip install` globally under PEP 668), but the parsing defect is **not platform-specific** — it triggers whenever the shell echoes the command into the integration stream, and it mis-fires **even when dbt is importable** (the echo match precedes the real stdout markers).

Worse, once a machine has persisted that garbage value, it stays broken on **every** subsequent dbt/pip call until the user clears the setting by hand — the write-side fix below prevents new poisoning but cannot recover an already-affected install.

### 1. Stop producing the bad value (`detectPythonFromShell` / `executeInTerminal`)

1. **Break the self-referential marker.** Emit each marker as two string literals joined at runtime in Python (`"__DBT_DETE" + "CT_START__"`), so the *contiguous* token only ever appears in stdout — never in the echoed command. The marker search can no longer match the echo.
2. **Validate before persisting.** Only accept an absolute path that exists on disk; otherwise return `undefined`. Shell noise or a leaked echo can never again be written as the interpreter path.
3. **Clear the 10s race timeout** once the read resolves, so it doesn't keep the event loop alive after each probe.

### 2. Stop consuming a bad value, and self-heal already-poisoned configs (`get pythonPath()`)

The `pythonPath` getter previously trusted `dbt.dbtPythonPathOverride` unconditionally. It now validates the override via `isUsableInterpreterOverride`:

- accept a **bare command** resolved via `PATH` (`python`, `python3`, `python3.11`), or a **filesystem path** (absolute or containing a separator) that exists on disk;
- reject anything else (shell noise, a leaked probe fragment), `warn(..., /* sendTelemetry */ true)` so recovery is observable in telemetry, and fall back to normal interpreter detection.

This recovers the machines already carrying a poisoned setting **on next launch**, with no manual intervention, while leaving every legitimate override (bare command or real path) untouched.

### Parser layer — real shell, not platform-specific

The only VS-Code-specific link in the chain is "shell integration's `read()` includes the command echo" — which is ordinary PTY echo. Driving a **real interactive bash over a PTY**, typing the exact probe commands, and running the real marker-parser over the captured stream:

```
BEFORE FIX (literal markers, no validation)
  [OLD] python has dbt: false  -> command echoed into stream: yes  -> result: "); print(sys.executable); print("
  [OLD] python has dbt: true   -> command echoed into stream: yes  -> result: "); print(sys.executable); print("

AFTER FIX (split markers + absolute-path validation)
  [NEW] python has dbt: false  -> echo only in split form          -> result: undefined
  [NEW] python has dbt: true   -> result: /path/to/.venv/bin/python3
```

The `"); print(sys.executable); print("` value reproduced above is the exact string reported from the field.

### End-to-end — docker code-server, poisoned setting

A code-server container (jaffle-shop-duckdb, dbt 1.11.11) with `dbt.dbtPythonPathOverride` set to the exact leaked fragment `"); print(sys.executable); print("`:

**Before** — the poisoned value is spawned verbatim, dbt detection fails:

```
DBTCoreDetectionError:Detection failed: Command not found: ""); print(sys.executable); print(""...
Starting python bridge [{"pythonPath":"\"); print(sys.executable); print(\"", ...}]
Error: spawn "); print(sys.executable); print(" ENOENT
/bin/sh: 1: ); print(sys.executable); print(: not found
```

before – dbt core is not installed

**After** — the same poisoned value is ignored and logged, detection falls back to the real interpreter, dbt initializes (`target/` compiled, status bar healthy), and `0` spawn/detection errors remain:

```
[warning] pythonEnvironment:pythonPath:Ignoring invalid dbtPythonPathOverride (not a usable interpreter): "); print(sys.executable); print("
[debug]   Starting python bridge [{"pythonPath":"/bin/python3","cwd":"/home/coder/jaffle-shop-duckdb", ...}]
```

after – dbt core detected via fallback interpreter

## Tests

`src/test/suite/pythonEnvironmentDetect.test.ts` (10 cases, all passing):

*Detection probe (`detectPythonFromShell`):*
- returns the interpreter path when dbt is importable
- returns `undefined` (never a probe fragment) when only the command echo is captured
- rejects the exact probe-fragment value even if it lands between markers
- rejects a non-absolute value
- rejects an absolute path that doesn't exist on disk

*Read-side override guard (`get pythonPath()`):*
- keeps a valid absolute interpreter path that exists on disk
- keeps a bare command that resolves via `PATH` (e.g. `python3`)
- ignores the leaked probe-fragment value and falls back, with telemetry
- ignores an absolute override path that does not exist on disk
- falls back without warning when no override is set

`tsc --noEmit`, ESLint, and Prettier all clean.

## Note

A separate follow-up tracks proper support for installing dbt into a PEP 668 externally-managed (Homebrew) Python — that install path is out of scope here.


dev-punia-altimate:
## ✅ Tests — All Passed



cc @ralphstodomingo

mdesmet (approved):

PR #1984, opened 2026-05-29
fix: skip Lineage:getStartingNode telemetry for non-dbt files

ralphstodomingo:
## Problem
With the model lineage panel open, switching the active editor to a **non-dbt file** (`.sh`, `.md`, `.txt`, …) emits an info-level `Lineage:getStartingNode` telemetry event (`No node found for <file>`) on **every editor change**. A non-dbt file can never be a dbt node, so this is expected behavior being over-reported — and in the field it's the single highest-volume named telemetry event (~1,900 events across ~108 machines in 24h on the current release, all integration modes, all platforms).

## Root cause
`getStartingNode` (`src/webview_provider/newLineagePanel.ts`) looks the active file up in the node/function maps; for a non-dbt file the lookup never matches and control reaches `dbtTerminal.info("Lineage:getStartingNode", …)`, which defaults `sendTelemetry=true`.

This is a **distinct branch** from the one fixed by #1931 (the `No event found` manifest-load race — same event name, different message). That fix didn't touch this `No node found for <file>` path.

## Fix
Gate the telemetry on the file extension: keep the output-channel log for debugging, but pass `sendTelemetry=false` when the active file isn't a dbt file (`.sql`/`.py`/`.csv`). A dbt file with no matching node still reports telemetry — that remains a legitimate diagnostic.

## Pre/post-fix repro (unit test, mocked telemetry)
Telemetry can't be reliably asserted through Playwright, so coverage is a mocked-`DBTTerminal` test driving `getStartingNode`:

**Before** (against unpatched code):
```
✕ does NOT send telemetry when the active file is a non-dbt file
    Expected: "Lineage:getStartingNode", StringContaining "No node found", false
    Received: "Lineage:getStartingNode", "No node found for etc-backfill_6.sh"   ← no sendTelemetry arg → defaults true
```

**After**:
```
✓ does NOT send telemetry when the active file is a non-dbt file
✓ still sends telemetry for a dbt file with no matching node
Tests: 7 passed, 7 total
```

## Risk
Minimal, localized — one call-site argument + a regression test; output-channel logging is unchanged. Heads-up: open PR #1905 (ERD overlay) also edits `getStartingNode` (a different line), so a trivial rebase may be needed depending on merge order.

## Summary by CodeRabbit

* **Bug Fixes**
  * Telemetry logging now only occurs for recognized dbt files, preventing extraneous log entries for non-dbt scripts.

* **Tests**
  * Added test coverage for getStartingNode telemetry behavior validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/AltimateAI/vscode-dbt-power-user/pull/1984?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/test/suite/newLineagePanel.test.ts
+++ b/src/test/suite/newLineagePanel.test.ts
@@ -1,5 +1,5 @@
 import { beforeEach, describe, expect, it, jest } from "@jest/globals";
-import { workspace } from "vscode";
+import { window, workspace } from "vscode";
 import { NewLineagePanel } from "../../webview_provider/newLineagePanel";
 
 describe("NewLineagePanel", () => {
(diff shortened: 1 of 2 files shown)


dev-punia-altimate:
## ✅ Tests — All Passed



cc @ralphstodomingo

saravmajestic (approved):

PR #1991, opened 2026-06-01
ci: include release notes in release Slack notification

ralphstodomingo:
## Goal
The release Slack notifications post only a per-marketplace status line (`Tag: … release to … status: …`). This adds the **release notes body** to the notification, so the message includes what actually shipped.

## Change
- Consolidates the two notify jobs (`release-vsstudio-notify`, `release-openvsx-notify`) into a single `release-notify` job.
- Fetches release notes via `gh release view` (with a graceful fallback to tag-only text if no notes/Release).
- Posts one notification with the notes body + a **combined** VS Marketplace + OpenVSX status (reports failure if either publish failed).

## Notes
- The release jobs run on the `release: created` event (where `github.ref` is the tag), so a published Release exists and the notes resolve.
- Kept **inline** rather than referencing a shared reusable workflow: that reusable lives in a private repo, and this repo is public, so a cross-repo private dependency isn't appropriate here.
- Net −0 behavioral change to publishing; only the notification step changes. `actionlint` clean.

## Summary by CodeRabbit

* **Chores**
  * Improved release notification workflow to send consolidated notifications across all marketplace publishing activities with integrated release notes details.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -181,20 +181,6 @@ jobs:
           VSCE_PAT: ${{ secrets.VSCE_PAT }}
           VSCE_TARGET: ${{ matrix.target }}
 
-    # Send notification after all platform builds (only on the last matrix entry)
-  release-vsstudio-notify:
-    runs-on: ubuntu-latest
-    needs: release-vsstudio-marketplace
-    if: always() && startsWith( github.ref, 'refs/tags/')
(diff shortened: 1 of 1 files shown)


saravmajestic (approved):

dev-punia-altimate:
## ✅ Tests — All Passed



cc @ralphstodomingo

PR #1789, opened 2026-01-07
fix: Improve SQL formatter handling of long lines (#1717)

shreyastelkar:
## Summary
Fixes #1717 - SQL formatter corrupting files with long lines

## Problem
When sqlfmt splits a long line into multiple lines, the formatter inserted them incorrectly, corrupting the SQL.

## Screenshots:
**Before**:

**After**:

## Solution
- Track position in old document with `lastOldLineProcessed`
- Track consecutive ADD operations with `consecutiveAddCount` to insert in correct order
- Add early return for empty documents
- Simplify diff processing by removing complex inter-chunk logic

## Tests
Added 14 unit tests for `processDiffOutput` covering long lines, multi-chunk diffs, and edge cases.

## Checklist
- [x] All tests pass (149 existing + 14 new)
- [x] Build succeeds

PR #1992, opened 2026-06-02
chore: remove dead isDeleteChange method in SQL formatter

ralphstodomingo:
## Summary

Removes the dead `isDeleteChange` method in the SQL formatter — it has no callers.

## Background

`isDeleteChange` was wired into the original line-by-line diff algorithm to skip sqlfmt's `\ No newline at end of file` marker among delete changes. The chunk-range rewrite in #1789 reconstructs new content from add/normal changes and replaces whole old-ranges, so delete changes are no longer iterated — leaving `isDeleteChange` orphaned.

Investigation (App Insights + driving real sqlfmt 0.18 & 0.30) confirmed there is **no behavioral bug to fix here**: stock sqlfmt always emits the `\ No newline at end of file` marker on the **source** side, which `parse-diff` classifies as a `del` and the existing `isAddChange || isNormalChange` filter already excludes. The marker only leaks if classified as a synthetic `add` (formatted side missing a trailing newline), which stock sqlfmt never produces — and that path emits no telemetry anyway. So this is a pure dead-code cleanup, no behavior change.

## Changes

- Remove the unused `isDeleteChange` method.

## Testing

- `tsc --noEmit` and `prettier --check` clean. No behavior change (method had no callers).

## Summary by CodeRabbit

* **Refactor**
  * Removed unused internal helper method from document formatting provider to improve code maintainability.

### DIFF
--- a/src/document_formatting_edit_provider/dbtDocumentFormattingEditProvider.ts
+++ b/src/document_formatting_edit_provider/dbtDocumentFormattingEditProvider.ts
@@ -290,16 +290,4 @@ export class DbtDocumentFormattingEditProvider implements DocumentFormattingEdit
   ): change is parseDiff.NormalChange {
     return change.type === "normal";
   }
-
-  private isDeleteChange(
-    change: parseDiff.Change,
-  ): change is parseDiff.DeleteChange {
-    return (
-      /*
-          parseDiff reads sqlfmt's "\ No newline at end of file" diff output as a delete change.
-          This deceptive delete change should be skipped. So, adding an edge case to the expression.
-      */
-      change.type === "del" && change.content !== "\\ No newline at end of file"
-    );
-  }
 }


ralphstodomingo:
## Triage update — scope correction

Investigated whether this PR addresses the field `formatDbtModelApplyDiffError` signal. It does not — and that's fine; this is a defensive correctness fix.

**App Insights (staging, 30d):** `formatDbtModelApplyDiffError` runs ~175–188 events/30d, and **~100% carry `error_message: "sqlfmt not found"`** (the outer catch — sqlfmt not installed). **Zero** events come from `processDiffOutput` throwing or from this marker leak.

**The marker leak emits no telemetry** — it's a silent wrong-output, and only triggers when the *formatted* output lacks a trailing newline (parse-diff classifies the `\ No newline at end of file` marker as a synthetic `add`). I confirmed via this repo's `parse-diff` that stock sqlfmt (0.18 & 0.30) always attaches the marker to the source (`-`) side → classified `del` → already excluded by the existing `isAddChange || isNormalChange` filter. So a live CLI repro isn't achievable; the failing case is exercised by the new unit test `src/test/suite/dbtDocumentFormatting.test.ts`.

**Net:** worth landing as hardening — it makes marker exclusion classification-independent and removes confirmed dead code (`isDeleteChange` was never called) — but it should be read as defensive, not a fix for that telemetry signal. The real driver of that event ("sqlfmt not found" surfaced as an error + popup) is a separate UX concern.

saravmajestic (approved):

PR #2015, opened 2026-06-26
feat: support hover and autocomplete for dbt Mesh cross-project refs

wornjs:
### Problem

dbt Mesh allows referencing models across projects using the two-argument `ref('project_name', 'model_name')`. Previously, hover information, go-to-definition, and autocomplete did not resolve these cross-project references — they only searched within the current project's manifest, so external nodes were silently ignored.

### Solution

- **Autocomplete** (`modelAutocompletionProvider.ts`): Instead of returning only the current project's model list, aggregate models from all loaded projects (deduped by `projectName|packageName|modelName`) so cross-project suggestions always appear.
- **Hover** (`modelHoverProvider.ts`): Pass `packageName` (the first argument of a two-arg `ref`) into `getHoverMarkdownFor`. When present, search `nodeMetaMap.nodes()` by `name + package_name` before falling back to `lookupByBaseName`, which only covers local files.
- **Go-to-definition** (`modelDefinitionProvider.ts`): Same lookup strategy — search by `name + package_name` for cross-project nodes first, then fall back to the existing local path lookup.

### Screenshot/Demo

N/A — behavior change in language server features (autocomplete, hover, go-to-definition). No UI changes.

### How to test

1. Set up a dbt Mesh workspace with at least two projects where one project depends on the other via cross-project `ref`.
2. Open a model file that uses `ref('other_project', 'model_name')`.
3. **Autocomplete**: Inside a `ref()`, verify that models from the dependent project appear as suggestions.
4. **Hover**: Hover over a cross-project `ref(...)` and confirm the hover card shows the model's description.
5. **Go-to-definition**: Use go-to-definition (`F12`) on a cross-project ref and confirm it navigates to the correct file in the dependent project.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Autocomplete now aggregates model suggestions across all loaded projects, improving discovery of cross-project references.
  * Hover and go-to-definition now resolve cross-project `ref` expressions for referenced models in other projects.
* **Bug Fixes**
  * Improved deduplication of autocomplete suggestions and excluded analysis-only models from autocomplete results.
  * Cross-project model lookups now resolve more reliably to the correct target location for hover and navigation.
* **Chores**
  * Minor TypeScript/formatting and logging adjustments to test generation behavior; non-functional YAML import repositioning.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "vscode-dbt-power-user",
-  "version": "0.61.6",
+  "version": "0.61.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "vscode-dbt-power-user",
-      "version": "0.61.6",
+      "version": "0.61.7",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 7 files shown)


dev-punia-altimate:

No supported files changed.

ralphstodomingo:
Thanks for this — cross-project ref support is a nice addition. A couple of things before merge:

**🔴 Lockfile changes — please revert**

This PR doesn't touch `package.json`, so the lockfile diff is unintended (looks like an `npm install` behind a corporate proxy). It shouldn't land because it:
- rewrites 8 `@vscode/vsce-sign-*` `resolved` URLs to an internal `http://nexus.toss.bz/...` mirror (plain HTTP) instead of the public registry,
- bumps the package `version` `0.61.6 → 0.61.7`, and
- adds unrelated optional-platform entries and removes a `rollup` entry in `webview_panels/package-lock.json`.

Since `package.json` is unchanged, both lockfiles should be byte-identical to `master`:
```bash
git checkout origin/master -- package-lock.json webview_panels/package-lock.json
git commit -m "chore: revert unintended lockfile changes"
```

**🟡 Autocomplete now aggregates every open project (`modelAutocompletionProvider.ts`)**

`getAutoCompleteItems` switched from the current project's models to iterating *all* loaded projects, deduped only on `packageName|modelName`, with no check that the other project is an actual dependency. In a multi-root workspace where the projects aren't declared mesh dependencies, this suggests models that aren't referenceable and inserts a two-arg `ref("other_project", "model")` that won't compile. And since hover / go-to-definition only search the current project's `nodeMetaMap`, accepting such a suggestion yields a ref the other two providers can't resolve.

For a genuine dbt-loom/cloud mesh, the producing project's nodes are already injected into the consumer's manifest (flagged `is_external_project`), so they're already present in the current project's autocomplete map — meaning the aggregation's only net-new effect is surfacing models from projects that *aren't* dependencies. Could you scope aggregation to declared dependencies (or to external nodes already present in the current manifest)?

**🟡 Definition provider runs the new lookup on single-arg refs too (`modelDefinitionProvider.ts`)**

Unlike the hover provider — which gates its `.find()` behind `if (packageName)` — the definition provider's `nodes().find(n => n.name === modelName && n.package_name === projectName)` is unconditional, so it now runs on the single-arg `ref('model')` path as well, ahead of `lookupByBaseName`. Since `NodeMetaData` doesn't distinguish model versions, a versioned model has multiple nodes sharing `name` + `package_name`, and `.find()` would return whichever iterates first — which may not be the latest version. Could you gate this behind the cross-project arg the way hover does, so single-arg in-project lookups stay on `lookupByBaseName`?

dev-punia-altimate:

No supported files changed.

PR #2021, opened 2026-07-01
fix: lazy-init git watching in SQL codelens to avoid "Git model not found" crash

ralphstodomingo:
## Problem

`SqlActionsCodeLensProvider` reads the built-in Git extension API — `extensions.getExtension("vscode.git")?.exports?.getAPI(1)` — to know which files have uncommitted changes. `getAPI(1)` **throws** `Git model not found` when the Git extension's model isn't initialized yet (lazy-activation race) or git is disabled. That call ran from `initGitWatcher()` in the **constructor**, which executes during extension activation — exactly when the Git extension may not have finished initializing. The throw escaped an un-`.catch()`'d `.then()` and surfaced as a `catchAllError` unhandled-rejection telemetry event.

## Why does a SQL codelens provider need git?

Fair question — and the answer scopes the fix. Git is needed for exactly **one** codelens: `$(sparkle) Review` → "Review code changes with Altimate" (`dbtPowerUser.reviewCodeWithAltimate`), which is shown **only** on files that have uncommitted working-tree/index changes (so "review the changes you made" makes sense). Every other lens — Execute Query, Explain, Optimize, Document, and the YAML Run/Test/Document lenses — is git-independent.

So git is legitimately needed, but **only when the user is actually looking at a SQL/YAML file** — never at activation time. That's the key: the constructor had no reason to touch git.

## Change

Stop touching git during construction. Initialize the git watcher **lazily on the first `provideCodeLenses` call** (`ensureGitWatcher`, guarded to run once). By the time any codelens is requested, a SQL/YAML file is open and the Git extension has activated, so `getAPI(1)` no longer races its model init — the **root cause is removed, not masked**.

A `try/catch` around `getAPI(1)` (plus the two-arg `.then(onFulfilled, onRejected)` on `activate()`, since `activate()` returns a `Thenable` with no `.catch`) is retained as a belt, so a genuinely-unavailable git model (e.g. `git.enabled: false`) degrades gracefully instead of crashing.

### Why not just `try/catch` the throw and stop there?

A plain guard prevents the crash but has a hidden cost: `watchGitState()` only ran **once**, at construction. On exactly the machines that hit the race, the catch would fire and the **Review lens would be silently dead for the whole session** — trading a crash for a quietly missing feature. Lazy-init fixes the crash *and* keeps the feature working.

## Verification (code-server harness, current `master` @ 0.61.7)

**1. Feature preserved** — in a real git repo (`git init` the jaffle project + an uncommitted edit to `models/customers.sql`), the `$(sparkle) Review` lens appears on the modified file. This is the lazy git watcher attaching on first codelens request and populating `changedFiles`:

Review codelens present on a modified model in a git repo

**2. Crash gone** — with `"git.enabled": false`, opening a SQL file (which triggers the lazy init → `getAPI(1)` throw → caught) produces no `Git model not found` / `catchAllError` in the exthost log; extension activates cleanly:

```
=== search for Git model not found / watchGitState / catchAllError (should be ABSENT) ===
NONE — no crash
=== extension activated ===
1   # "Initialized dbt project"
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SQL code lens “Review” actions by deferring Git setup until code lenses are requested, avoiding race conditions during extension activation.
  * Improved robustness when Git activation fails or when the Git model isn’t available (no unhandled errors).
  * Prevented repeated Git watcher initialization across multiple code lens updates.
* **Tests**
  * Added Jest coverage for lazy Git initialization, safe handling of “Git model not found,” and correct gating of “Review” lenses based on working-tree changes.

### DIFF
--- a/src/test/mock/vscode.ts
+++ b/src/test/mock/vscode.ts
@@ -65,6 +65,18 @@ export class Location {
   }
 }
 
+export class CodeLens {
+  constructor(
+    public range: Range,
+    public command?: {
+      title: string;
+      command: string;
+      tooltip?: string;
+      arguments?: any[];
+    },
+  ) {}
+}
+
 export const DiagnosticSeverity = {
   Error: 0,
   Warning: 1,
@@ -130,6 +142,10 @@ export const window = {
   showInformationMessage: jest.fn().mockReturnValue(Promise.resolve()),
   showWarningMessage: jest.fn().mockReturnValue(Promise.resolve()),
   showErrorMessage: jest.fn().mockReturnValue(Promise.resolve()),
+  onDidChangeActiveTextEditor: jest
+    .fn()
+    .mockReturnValue({ dispose: jest.fn() }),
+  activeTextEditor: undefined as any,
   createOutputChannel: jest.fn().mockReturnValue({
     append: jest.fn(),
     appendLine: jest.fn(),
--- a/src/code_lens_provider/sqlActionsCodeLensProvider.ts
+++ b/src/code_lens_provider/sqlActionsCodeLensProvider.ts
@@ -44,9 +44,17 @@ export class SqlActionsCodeLensProvider
     this._onDidChangeCodeLenses.event;
   private disposables: Disposable[] = [];
   private changedFiles = new Set<string>();
+  private gitWatcherInitialized = false;
 
   constructor(private altimateCodeChatService: AltimateCodeChatService) {
-    this.initGitWatcher();
+    // Git watching is initialized lazily (see `ensureGitWatcher`, called from
+    // `provideCodeLenses`) rather than here. This provider is constructed during
+    // extension activation, which races the built-in Git extension's own model
+    // initialization — `getAPI(1)` throws "Git model not found" in that window.
+    // Git is only needed to gate a single codelens ("Review changes with
+    // Altimate") on files with uncommitted changes, so we don't need it until a
+    // SQL/YAML file is actually opened — by which point git has activated and
+    // the race is gone.
     this.disposables.push(
       window.onDidChangeActiveTextEditor((editor) => {
         if (editor) {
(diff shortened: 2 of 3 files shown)


ralphstodomingo:
### E2E evidence (code-server harness, current `master` @ 0.61.7)

This is a log-only unhandled rejection (no UI), so the evidence is the extension-host log. Forced the throw deterministically with `"git.enabled": false` (makes the built-in Git extension's `getAPI(1)` throw `Git model not found`).

**Before this change** — unhandled rejection → `catchAllError`:

```
[error] Error: Git model not found
    at Rs.getAPI (…/extensions/git/dist/main.js:115:32850)
    at SqlActionsCodeLensProvider.watchGitState (…/dist/extension.js:130:858515)
    at SqlActionsCodeLensProvider.initGitWatcher (…/dist/extension.js:130:858417)
    at new SqlActionsCodeLensProvider (…/dist/extension.js:130:857508)
```

**After this change** — same setup, grepping the exthost log for `Git model not found` / `watchGitState` / `catchAllError`:

```
=== search for Git model not found / watchGitState / catchAllError (should be ABSENT) ===
NONE — fix holds
=== sanity: extension activated? (git.enabled false) ===
1   # "Initialized dbt project" — extension came up cleanly
```

The `getAPI(1)` throw is now swallowed by the `try/catch`, so no unhandled rejection and the extension activates normally.

ralphstodomingo:
**Approach revised** (see updated description): moved from a plain `try/catch` guard to **lazy-init** — the git watcher now starts on the first `provideCodeLenses` call instead of in the constructor, which removes the activation-time race rather than masking it, and (unlike the guard) keeps the "Review" codelens working on the machines that hit the race. The `try/catch` is retained as a belt. Re-verified in the harness: Review lens present in a real git repo, and no `Git model not found` with `git.enabled:false`. My earlier before/after log comment above still holds for the crash-gone check.

ralphstodomingo:
Added unit tests (`src/test/suite/sqlActionsCodeLensProvider.test.ts`, jest + mocked vscode): git not touched at construction (lazy init), throwing `getAPI(1)` doesn't crash `provideCodeLenses`, Review lens gated on git-change state (present when changed / absent otherwise), watcher initializes only once. Extends the shared vscode mock with `CodeLens` + `window.onDidChangeActiveTextEditor`/`activeTextEditor` (additive). Full suite green (600 passed / 1 skipped).

src/test/mock/vscode.ts
```diff
@@ -130,6 +142,10 @@ export const window = {
   showInformationMessage: jest.fn().mockReturnValue(Promise.resolve()),
   showWarningMessage: jest.fn().mockReturnValue(Promise.resolve()),
   showErrorMessage: jest.fn().mockReturnValue(Promise.resolve()),
+  onDidChangeActiveTextEditor: jest
+    .fn()
+    .mockReturnValue({ dispose: jest.fn() }),
+  activeTextEditor: undefined as any,
```
dev-punia-altimate, line 148: **[🔵 LOW]** According to the review checklist, using the `any` type should be avoided. If using `any` is necessary here for mocking purposes, please add a comment explaining the reason, or consider using a more specific mock type (e.g., `TextEditor | undefined`) if possible.

**Suggested change:**
```suggestion
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  activeTextEditor: undefined as any,
```

dev-punia-altimate (commented):
## 🤖 Code Review — OpenCodeReview (Gemini) — 1 finding(s)

- 1 anchored to a line (posted inline when the comment stream is on)
- 0 without a line anchor

#### 1. `src/test/mock/vscode.ts` (L148)

**[🔵 LOW]** According to the review checklist, using the `any` type should be avoided. If using `any` is necessary here for mocking purposes, please add a comment explaining the reason, or consider using a more specific mock type (e.g., `TextEditor | undefined`) if possible.

**Suggested change:**
```suggestion
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  activeTextEditor: undefined as any,
```

</details>

dev-punia-altimate:

No supported files changed.

dev-punia-altimate:

No supported files changed.

saravmajestic (approved):

dev-punia-altimate:

No supported files changed.

PR #2031, opened 2026-07-08
fix: prevent phantom dbt project registration under dbt_packages

ralphstodomingo:
## What

The config file watcher registered any newly-created `dbt_project.yml` as a standalone dbt project. Its only guard, `notInDBtPackages`, compares the new file against each already-registered project's package install path — which is `undefined` while that project is still initializing (running `dbt deps`).

During initialization, `dbt deps` writes each installed package's own `dbt_project.yml` (`dbt_utils`, `dbt_expectations`, …). The watcher fires before the parent's install path is known, so those packages get registered as **phantom projects**. Each phantom then runs its own initialization + `dbt deps`, which cascades through nested packages (`dbt_utils`' integration_tests declares `- local: ../`).

The initial discovery pass already excludes these directories statically (`**/{dbt_packages,site-packages,dbt_internal_packages}`); the watcher did not. This applies the same static path-segment exclude in the watcher, so the guard no longer depends on initialization timing.

## Changes

- `dbtWorkspaceFolder.ts`: static `dbt_packages` / `site-packages` / `dbt_internal_packages` segment exclude at the top of the watcher's `onDidCreate`.
- **Unit test** (`dbtWorkspaceFolder.watcherRace.test.ts`): drives the real `onDidCreate` handler for a `dbt_packages` path while the parent's install path is `undefined`, asserting no registration; a genuinely separate nested project still registers.

## Impact

Phantom registration is incorrect on every platform. It registers bogus `unknown_<guid>` projects, runs redundant `dbt deps` invocations, and churns the filesystem during startup. The cost scales with filesystem latency: the race window is widest on network / VDI mounts, which is where the cascade is most expensive and where it was originally observed. The known user workaround was `dbt.installDepsOnProjectInitialization: false`; this removes the need for it.

**Scope note.** This work started from a report of many Python console windows on Windows. That visibility is a *separate* defect: dbt subprocesses were spawned without `windowsHide`, so every dbt command opened a console window — including in ordinary single-project workspaces, just far fewer of them. That is fixed independently in the dbt integration library and is **not** what this PR does. This PR removes the redundant spawns; it does not make any spawn windowless.

## Verification

**Unit test** — fails on `master` (the phantom project is registered) and passes with the fix:

```
PASS src/test/suite/dbtWorkspaceFolder.watcherRace.test.ts
  ✓ does not register a dbt_project.yml under dbt_packages/ when the parent project's install path is not yet resolved
  ✓ still registers a genuinely separate nested project
Tests: 2 passed, 2 total
```

**Real-extension end-to-end** — verified locally with an `@vscode/test-electron` harness driving the real VS Code `FileSystemWatcher` against a live `DBTWorkspaceFolder`: a `dbt_project.yml` created under `dbt_packages/` is not registered while a genuinely separate nested project is (fails without the exclude, passes with it). Not committed to CI because VS Code's recursive file watcher delivers create events non-deterministically across headless CI runners (ubuntu under xvfb drops deeper-subtree events; macOS delivers to independent watcher instances out of order), and an excluded path yields no positive signal to synchronize on — so it could not be made reliable cross-platform. The unit test covers the handler's exclude logic deterministically on every platform.

**Windows path handling** — the exclude uses `path.sep` (`\` on Windows); verified under Windows `node` that `dbt_packages` / nested cascade / `site-packages` and UNC paths (`\\host\share\...\dbt_packages\...`, matching the reporting customer's environment) are excluded while legit nested + root projects are not.

## Test

```
npx jest src/test/suite/dbtWorkspaceFolder.watcherRace.test.ts
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dbt project discovery to ignore `dbt_project.yml` files found inside installed package locations (including `dbt_packages/` and custom packages-install paths) and virtual-environment directories.
  * Prevented package projects from being falsely registered as standalone projects during workspace startup, while still registering genuinely separate nested projects exactly once.
* **Tests**
  * Added Jest coverage for watcher startup race conditions under `dbt_packages/` and custom packages-install paths to verify correct register vs. ignore behavior.
* **Chores**
  * Updated the `@altimateai/dbt-integration` dependency version.


mdesmet:
How is the issue about registering phantom projects related to visible console windows? Are these console windows also showing up for single dbt projects?

ralphstodomingo:
**They're two independent defects that compound.** They're only related because one made the other visible enough to report.

Every dbt command in core mode runs through `PythonDBTCommandExecutionStrategy.executeCommand`, which spawns `python.exe -c "<dbt code>"` via `CommandProcessExecution.spawn()`. That spawn had no `windowsHide`, so on Windows **each dbt command opened a console window**. (Worth noting it spawns `python.exe`, not `pythonw.exe` — the `pythonw` swap and `detached: true` only apply on the python-bridge path, which is why the bridge was never the culprit and the reported windows were all `python.exe`.)

So:

- **Missing `windowsHide`** → why *any* spawn is visible.
- **Phantom registration** (this PR) → why there are *dozens* of spawns: each phantom initializes and runs its own `dbt deps`, and `dbt_utils`' integration_tests declares `- local: ../`, so it cascades.

Neither alone produces the report. The flood needed both.

**> Are these console windows also showing up for single dbt projects?**

Yes. A plain single-project workspace on Windows also popped a `python.exe` window per dbt command — it's just one or two transient flickers instead of dozens of persistent windows, which is presumably why it was never reported. That's fixed by the `windowsHide` change in the dbt integration library (shipping in the next release), **not** by this PR — and it fixes it for every Windows user, single-project included.

**Which means this PR shouldn't be justified by console windows at all**, and you're right to push on that. Once the `windowsHide` fix lands, the windows disappear whether or not this merges. I've rewritten the description accordingly: the actual case for this change is platform-independent correctness — phantom `unknown_<guid>` projects get registered, run redundant `dbt deps` invocations, and churn the filesystem during startup, on every OS. The cost scales with filesystem latency, which is why it showed up on a VDI / network mount. The console-window flood was just the loudest symptom, and it's what led us here.

ralphstodomingo:
## Verification: real before/after against the built library

Reworked to call the library predicate (see the commit above). Adding the end-to-end evidence, since this now spans two repos.

**Fixture** — `test-fixtures/phantom-project-custom-packages-path/`, a project that moves its packages off the default location:

```yaml
packages-install-path: vendor
```

| Path | Expectation |
|---|---|
| `./dbt_project.yml` | the real project — must register |
| `./vendor/dbt_utils/` | installed package under the custom path — **must NOT register** |
| `./vendor_archive/` | legit project merely sharing the `vendor` prefix — must register |

**Driving the real built artifacts** against those real files (fusion detection, deterministic on the discovery path — no race needed):

Before, with the currently published `@altimateai/dbt-integration`:

```
REGISTERED   .                   (must register)     OK
REGISTERED   ./vendor/dbt_utils  (MUST NOT register) <-- WRONG
REGISTERED   ./vendor_archive    (must register)     OK
```

After, with the library change:

```
REGISTERED   .                   (must register)     OK
excluded     ./vendor/dbt_utils  (MUST NOT register) OK
REGISTERED   ./vendor_archive    (must register)     OK
```

`vendor_archive` staying registered in both confirms the segment-boundary fix doesn't over-exclude — the old `startsWith` test would have treated it as living inside `vendor`.

**Unit level** — the new watcher test fails on the previous static-list implementation and passes now, verified by reverting only the implementation:

```
✓ does not register a dbt_project.yml under dbt_packages/ when the parent project's install path is not yet resolved
✕ does not register a package under a custom packages-install-path while the parent is initializing   <-- old impl
✓ still registers a genuinely separate nested project
```

Full suite green: **606 passed, 45 suites**, typecheck and lint clean.

**Not covered:** the initialization race itself is still not reproducible on Linux — init wins on a fast SSD, which is why the deterministic unit test remains the primary evidence for that path. The fixture above deliberately exercises the discovery path instead, where the difference is deterministic.

**Sequencing:** CI here stays red at typecheck until the library release ships `isInsidePackagesPath` / `resolvePackagesInstallPath`; the version bump is the last commit on this branch.

dev-punia-altimate (commented):
## 🤖 Code Review — OpenCodeReview (Gemini) — 2 finding(s)

- 2 anchored to a line (posted inline when the comment stream is on)
- 0 without a line anchor

#### 1. `src/dbt_client/dbtWorkspaceFolder.ts` (L172)

**[🟠 MEDIUM]** **Potential Failure in Package Exclusion & Performance Trap**

1. **Logic Flaw (Initial Discovery)**: During the initial workspace load, `this.dbtProjects` is empty. Consequently, `this.isInstalledPackage` checks paths against an empty array of known package paths. If a user has configured a custom `packages-install-path` (e.g., `my_packages`), it will not be statically excluded by `workspace.findFiles` and will bypass this filter, erroneously initializing nested dependencies as standalone projects.
2. **Performance Bottleneck**: Calling `this.isInstalledPackage(uri.fsPath)` inside the `filter` loop recalculates `this.knownPackagesInstallPaths()` for every discovered file. If `discoverProjects` is ever called while `this.dbtProjects` is populated, this will repeatedly invoke `resolvePackagesInstallPath`, causing O(N*M) synchronous file reads.

**Suggestion**: 
- For performance, compute the known paths once outside the `filter` loop and pass them as an argument to `isInstalledPackage`. 
- To fix the initial discovery flaw, the custom package paths of the discovered directories must be resolved *before* or *during* the filtering process, rather than relying on `this.dbtProjects` which hasn't been populated yet.

#### 2. `src/dbt_client/dbtWorkspaceFolder.ts` (L385-L388)

**[🟠 MEDIUM]** **Redundant Synchronous File I/O in File Watcher**

When bulk file creation occurs (e.g., during `dbt deps`), the file watcher's `onDidCreate` event fires multiple times in quick succession. For each event, `isInstalledPackage` calls `knownPackagesInstallPaths()`, which evaluates this code block.

If a project is mid-initialization, `project.getPackageInstallPath()` will be `undefined`, causing `resolvePackagesInstallPath` to read the `dbt_project.yml` synchronously from disk. This results in the exact same file being read synchronously multiple times per watcher burst, which can bottleneck performance and block the main thread.

**Suggestion**: Cache or memoize the result of `resolvePackagesInstallPath` (e.g., using a short-lived cache or caching it within the project instance) so that multiple watcher events during initialization don't cause repetitive synchronous file reads.

</details>

src/dbt_client/dbtWorkspaceFolder.ts
```diff
@@ -167,7 +169,7 @@ export class DBTWorkspaceFolder implements Disposable {
 
     const projectDirectories = dbtProjectFiles
       .filter((uri) => existsSync(uri.fsPath) && statSync(uri.fsPath).isFile())
-      .filter((uri) => this.notInVenv(uri.fsPath))
+      .filter((uri) => !this.isInstalledPackage(uri.fsPath))
```
dev-punia-altimate: **[🟠 MEDIUM]** **Potential Failure in Package Exclusion & Performance Trap**

1. **Logic Flaw (Initial Discovery)**: During the initial workspace load, `this.dbtProjects` is empty. Consequently, `this.isInstalledPackage` checks paths against an empty array of known package paths. If a user has configured a custom `packages-install-path` (e.g., `my_packages`), it will not be statically excluded by `workspace.findFiles` and will bypass this filter, erroneously initializing nested dependencies as standalone projects.
2. **Performance Bottleneck**: Calling `this.isInstalledPackage(uri.fsPath)` inside the `filter` loop recalculates `this.knownPackagesInstallPaths()` for every discovered file. If `discoverProjects` is ever called while `this.dbtProjects` is populated, this will repeatedly invoke `resolvePackagesInstallPath`, causing O(N*M) synchronous file reads.

**Suggestion**: 
- For performance, compute the known paths once outside the `filter` loop and pass them as an argument to `isInstalledPackage`. 
- To fix the initial discovery flaw, the custom package paths of the discovered directories must be resolved *before* or *during* the filtering process, rather than relying on `this.dbtProjects` which hasn't been populated yet.

src/dbt_client/dbtWorkspaceFolder.ts
```diff
@@ -353,30 +351,41 @@ export class DBTWorkspaceFolder implements Disposable {
...
+    return this.dbtProjects.map((project) => {
+      // `getPackageInstallPath()` stays undefined until a project finishes
+      // initializing, and `dbt deps` writes each package's `dbt_project.yml`
+      // during that window — so a watcher event arriving mid-initialization
+      // must not depend on it. Reading the configured path back out of
+      // `dbt_project.yml` is synchronous and closes that gap.
+      return (
+        project.getPackageInstallPath() ??
+        resolvePackagesInstallPath(project.projectRoot.fsPath)
+      );
```
dev-punia-altimate, line 393: **[🟠 MEDIUM]** **Redundant Synchronous File I/O in File Watcher**

When bulk file creation occurs (e.g., during `dbt deps`), the file watcher's `onDidCreate` event fires multiple times in quick succession. For each event, `isInstalledPackage` calls `knownPackagesInstallPaths()`, which evaluates this code block.

If a project is mid-initialization, `project.getPackageInstallPath()` will be `undefined`, causing `resolvePackagesInstallPath` to read the `dbt_project.yml` synchronously from disk. This results in the exact same file being read synchronously multiple times per watcher burst, which can bottleneck performance and block the main thread.

**Suggestion**: Cache or memoize the result of `resolvePackagesInstallPath` (e.g., using a short-lived cache or caching it within the project instance) so that multiple watcher events during initialization don't cause repetitive synchronous file reads.

dev-punia-altimate:

No supported files changed.

dev-punia-altimate (commented):
## 🤖 Code Review — OpenCodeReview (Gemini) — 3 finding(s)

- 3 anchored to a line (posted inline when the comment stream is on)
- 0 without a line anchor

#### 1. `src/dbt_client/dbtWorkspaceFolder.ts` (L172)

**[🟠 MEDIUM]** This `.filter()` loop calls `this.isInstalledPackage(uri.fsPath)` for every single element in `dbtProjectFiles`. `isInstalledPackage` internally calls `knownPackagesInstallPaths()`, which performs synchronous file reads (via `resolvePackagesInstallPath`). This leads to a massive `O(N*M)` synchronous disk I/O performance bottleneck that can freeze the extension host when there are many files.

**Suggestion:**
Compute `knownPackagesInstallPaths` once outside the `.filter()` loop and pass it as an argument to `isInstalledPackage()`.

**Suggested change:**
```suggestion
      // Extract this calculation outside the loop:
      // const packagesInstallPaths = this.knownPackagesInstallPaths();
      // .filter((uri) => !this.isInstalledPackage(uri.fsPath, packagesInstallPaths))
      .filter((uri) => !this.isInstalledPackage(uri.fsPath))
```

#### 2. `src/dbt_client/dbtWorkspaceFolder.ts` (L385-L389)

**[🔴 HIGH]** 1. **Unhandled I/O Exceptions**: If `resolvePackagesInstallPath` accesses files synchronously, any file system errors (e.g., `ENOENT` or `EACCES` during `dbt deps`) could crash the extension host since this is not wrapped in a `try...catch`. 
2. **Type Mismatch/Undefined values**: If both `project.getPackageInstallPath()` and `resolvePackagesInstallPath` return `undefined`, this array will contain `undefined` values. Depending on how `isInsidePackagesPath` handles `undefined`, this might lead to runtime errors or unintended behavior. The return type of this function is `string[]`, so we should filter out `undefined` or null values to maintain type safety.

**Suggestion:**
Wrap the synchronous read in a `try...catch` and filter out any falsy values from the final array.

**Suggested change:**
```suggestion
      let resolvedPath;
      try {
        resolvedPath = resolvePackagesInstallPath(project.projectRoot.fsPath);
      } catch (e) {
        // Log error if necessary, or fail silently
        console.warn('Failed to resolve packages install path:', e);
      }
      return (
        project.getPackageInstallPath() ?? resolvedPath
      );
    }).filter((path): path is string => Boolean(path));
```

#### 3. `src/dbt_client/dbtWorkspaceFolder.ts` (L362-L366)

**[🟠 MEDIUM]** To support caching the paths in `findDBTProjects` and avoid repeated I/O in loop iterations, consider accepting `packagesInstallPaths` as an argument here. 

Additionally, the docstring mentions filtering out `virtualenv`, but the function is delegated to `isInsidePackagesPath`. Please verify that `isInsidePackagesPath` in `@altimateai/dbt-integration` actually handles the `site-packages` string check previously handled by `notInVenv`. If it doesn't, virtual environments might be incorrectly registered.

**Suggested change:**
```suggestion
  private isInstalledPackage(fsPath: string, packagesInstallPaths: string[] = this.knownPackagesInstallPaths()): boolean {
    const isPackage = isInsidePackagesPath(
      fsPath,
      packagesInstallPaths,
    );
```

</details>

src/dbt_client/dbtWorkspaceFolder.ts
```diff
@@ -353,30 +351,41 @@ export class DBTWorkspaceFolder implements Disposable {
...
+    });
```
dev-punia-altimate, line 394: **[🔴 HIGH]** 1. **Unhandled I/O Exceptions**: If `resolvePackagesInstallPath` accesses files synchronously, any file system errors (e.g., `ENOENT` or `EACCES` during `dbt deps`) could crash the extension host since this is not wrapped in a `try...catch`. 
2. **Type Mismatch/Undefined values**: If both `project.getPackageInstallPath()` and `resolvePackagesInstallPath` return `undefined`, this array will contain `undefined` values. Depending on how `isInsidePackagesPath` handles `undefined`, this might lead to runtime errors or unintended behavior. The return type of this function is `string[]`, so we should filter out `undefined` or null values to maintain type safety.

**Suggestion:**
Wrap the synchronous read in a `try...catch` and filter out any falsy values from the final array.

**Suggested change:**
```suggestion
      let resolvedPath;
      try {
        resolvedPath = resolvePackagesInstallPath(project.projectRoot.fsPath);
      } catch (e) {
        // Log error if necessary, or fail silently
        console.warn('Failed to resolve packages install path:', e);
      }
      return (
        project.getPackageInstallPath() ?? resolvedPath
      );
    }).filter((path): path is string => Boolean(path));
```

src/dbt_client/dbtWorkspaceFolder.ts
```diff
@@ -353,30 +351,41 @@ export class DBTWorkspaceFolder implements Disposable {
...
+   *
+   * The rule itself lives in `@altimateai/dbt-integration` so that every
+   * consumer applies it identically; this only supplies the packages paths of
+   * the projects registered in this folder.
+   */
+  private isInstalledPackage(fsPath: string): boolean {
+    const isPackage = isInsidePackagesPath(
+      fsPath,
+      this.knownPackagesInstallPaths(),
+    );
```
dev-punia-altimate: **[🟠 MEDIUM]** To support caching the paths in `findDBTProjects` and avoid repeated I/O in loop iterations, consider accepting `packagesInstallPaths` as an argument here. 

Additionally, the docstring mentions filtering out `virtualenv`, but the function is delegated to `isInsidePackagesPath`. Please verify that `isInsidePackagesPath` in `@altimateai/dbt-integration` actually handles the `site-packages` string check previously handled by `notInVenv`. If it doesn't, virtual environments might be incorrectly registered.

**Suggested change:**
```suggestion
  private isInstalledPackage(fsPath: string, packagesInstallPaths: string[] = this.knownPackagesInstallPaths()): boolean {
    const isPackage = isInsidePackagesPath(
      fsPath,
      packagesInstallPaths,
    );
```

src/dbt_client/dbtWorkspaceFolder.ts
ralphstodomingo: Checked both parts against the source and neither holds — this is safe as written.

**1. "Unhandled I/O exceptions could crash the host."** `resolvePackagesInstallPath` already wraps the read in try/catch (`altimate-dbt-integration/src/dbtIntegration.ts`):

```ts
if (!existsSync(dbtProjectFile)) {
  return defaultPath;
}
let parsed: unknown;
try {
  parsed = parse(readFileSync(dbtProjectFile, "utf8"));   // read is inside the try
} catch (error) {
  onParseError?.(error instanceof Error ? error.message : String(error));
  return defaultPath;
}
```

The `readFileSync` is *inside* the try, so an ENOENT (e.g. the file deleted between `existsSync` and the read during `dbt deps`) or EACCES is caught and falls back to the default path. It can't throw into the extension host. This is deliberate — the doc comment on the function says a malformed/unreadable `dbt_project.yml` must not abort discovery.

**2. "The array might contain `undefined`."** `resolvePackagesInstallPath` returns `string` in every branch (its declared type is `string`, not `string | undefined`) — it always resolves to either the configured path or `<dir>/dbt_packages`. So `getPackageInstallPath() ?? resolvePackagesInstallPath(...)` is always a `string`, and the array is genuinely `string[]` with no falsy members. The `.filter(Boolean)` would never remove anything.

So the suggested try/catch + filter guards a state that can't occur; I've left it out to avoid dead defensive code. Happy to reconsider if you can point at a path where `resolvePackagesInstallPath` returns undefined or throws.

src/dbt_client/dbtWorkspaceFolder.ts
ralphstodomingo: Two parts:

**Caching / repeated I/O** — good call, done in e0f177d8. `isInstalledPackage` now takes `packagesInstallPaths` as an optional arg (defaulting to `knownPackagesInstallPaths()`), and the discovery scan computes it once before the filter instead of per candidate. The watcher path keeps the default so it still reflects the current set of registered projects on each event. (In practice the scan runs with `dbtProjects` empty at activation, so the reads were near-zero, but computing once is cleaner regardless.)

**site-packages / virtualenv handling** — verified, it's covered. `isInsidePackagesPath` checks `EXCLUDED_PROJECT_DIRS`, which is `["dbt_packages", "site-packages", "dbt_internal_packages"]`, so `notInVenv`'s job carries over:

```
isInsidePackagesPath("/venv/lib/python3.11/site-packages/dbt_utils/dbt_project.yml") // => true
```

One deliberate difference from the old `notInVenv`: it did a substring `path.includes("site-packages")`, whereas the new check matches `site-packages` as a full path *segment*. That's stricter and more correct — a directory literally named `site-packages` is excluded, but an unrelated path like `.../my-site-packages-notes/...` no longer is. Covered by the `"excludes every directory name in EXCLUDED_PROJECT_DIRS"` unit test.

src/dbt_client/dbtWorkspaceFolder.ts
```diff
@@ -28,7 +28,9 @@ import {
   DBT_PROJECT_FILE,
   DBTProjectDetection,
   DBTTerminal,
+  isInsidePackagesPath,
   readAndParseProjectConfig,
+  resolvePackagesInstallPath,
```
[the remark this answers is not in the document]
ralphstodomingo: Resolved — this was against `@altimateai/dbt-integration@0.3.4`, which predated these exports. 0.3.7 (published, with `isInsidePackagesPath` / `resolvePackagesInstallPath`) is now the pinned dependency, so the typecheck/build failures this flagged are gone. CI is green on the current commit.

saravmajestic (approved):

dev-punia-altimate:

No supported files changed.

PR #2032, opened 2026-07-08
feat: What's New panel (editor-area changelog)

ralphstodomingo:
## What

A "What's New" panel that opens in the editor area after a minor/major version bump (patches stay silent) and on demand via **dbt Power User: Show What's New**. It renders this product's entries from the public changelog RSS feed.

## Data source

Reads the public changelog feed (`altimate.ai/changelog.rss.xml`) directly:

- **`<category>`** carries each entry's product, so the panel scopes itself to `dbt-power-user` rather than showing every product's updates.
- **`<altimate:version product="…">`** carries the per-product version, rendered as the entry's version chip. It's optional — most entries omit it, and the chip simply doesn't render.
- The tag arrives as a `[Label]` prefix on the title, so it's split back out and the title renders clean.

Parsing is deliberately targeted rather than a new XML dependency: the feed comes from a fixed template we control (flat items, no CDATA, a known escape set). If that stops holding, swap in a real parser.

## Panel

- **Month-grouped** changelog, newest first, with a **sticky timeline rail** + scrollspy.
- **Multi-select tag filters** — `new / improved / beta / fixed`, the changelog source's own vocabulary — with live counts, disabled-at-zero, and an empty state.
- Per entry: tag badge + icon, optional version chip, and date. **Entry titles are plain text** — the linked changelog entry only reproduces the tag/date/title/description already shown here.
- **"View full changelog" CTA** and an "Also from Altimate AI" aside — the one place a trip out of the editor earns itself.
- Fonts (**Geist Pixel** + **Manrope**, both OFL) **inlined as base64** so they load under the webview CSP with no CDN. Type scale anchored in `rem` so it tracks the editor font size. Dark palette with **derived WCAG-AA light counterparts**.
- Self-contained loading spinner — no external UI-library dependency.

## Extension

- `whatsNewPanel.ts`: editor-column panel, semver minor/major-only auto-open, always-persist version (opt-outs stick), fresh-install defers to the setup walkthrough. Marketplace icon on the tab; feed fetch logs the resolved URL on failure and falls back to the last good copy (globalState cache) when offline.
- `dbtPowerUser.showWhatsNew` command + `dbt.showChangelogOnUpdate` setting.
- Webview CSP: `font-src` now allows `data:` so panels can inline webfonts.

## Telemetry

Every clickable surface is tracked, each event carrying the running `version` so engagement reads per release:

- `full_changelog_clicked` with a `source` (`topbar` / `cta` / `footer` / `error_state`) — the four links lead to the same place, so `source` is what tells them apart.
- `whats_new_filter_toggled` (`tag`, `enabled`) — the per-item interest signal now that entries aren't links.
- `whats_new_timeline_navigated` (`month`), `whats_new_product_link_clicked` (`product`), `whats_new_site_link_clicked` (`source`).
- `whats_new_page_opened` (host-side, with `trigger` + `items_shown`).

The filter toggle decides from a ref, not from `activeTags`: reading the state directly went stale between two clicks in one render (reporting a toggle-off as `enabled: true`), while deciding inside the state updater would double-fire under `StrictMode`.

## No new UI-library dependency

The panel is hand-rolled (fonts/icons/spinner are local) — `@altimateai/lego` is not on npm yet, so nothing here depends on it. The lineage panel's imports stay on the published `@altimateai/ui-components/lineage`, pinned to the released **0.0.86**. No unpublished release is required to build or ship this.

## Verification

- Parser run against the **live feed**: correctly extracts this product's entry (title stripped of its `[New]` prefix, correct `tag` / `version` / `date`, anchor intact, entities decoded) and ignores the other products' entries.
- Rendered the **built webview bundle against live feed data** — entry, version chip, month grouping, filter counts (disabled at zero), timeline scrollspy, and both fonts loading. Verified in dark and light.
- Every telemetry event asserted from the rendered bundle by clicking each control — `source` values distinct, filter toggle correct on and off (including a same-tick double click), `version` present on all.
- Auto-open trigger unit-tested by replaying a release sequence (`0.62.1 → 0.62.2 → 0.63.0 → … → 1.0.0`): opens only on minor/major, silent on patch and downgrade, records the version even when suppressed, respects the opt-out. Hand-testing this is impractical — activation consumes the trigger by persisting the version whether or not the page opened.
- `tsc --noEmit` (extension + webview) — 0 errors. ESLint clean. Full Jest suite green.

## Summary by CodeRabbit

* **New Features**
  * Added a “What’s New” panel in the extension webview with month-based grouping, tag filters, timeline-style navigation, and links to the full changelog.
  * Added an auto-open experience for minor/major updates and a manual “Show What’s New” command.
  * Added a setting to control whether the “What’s New” prompt appears after updates.
* **Bug Fixes**
  * Improved webview font rendering by expanding Content Security Policy `font-src` to allow embedded fonts.
  * Added more resilient “What’s New” loading with caching fallback when the feed can’t be fetched or parsed.

## Version

Bumps to **0.63.0**. Minor rather than patch by necessity, not just convention: the page auto-opens only on a minor or major upgrade, so shipping this under a patch version would mean existing users never see it. This makes the PR self-contained for a release even if it is the only thing that merges.


package.json
```diff
@@ -129,6 +129,16 @@
             "description": "Enable dbt docs collaboration.",
             "default": true
           },
+          "dbt.showChangelogOnUpdate": {
+            "type": "boolean",
+            "description": "Show the What's New page automatically after the extension updates to a new minor or major version.",
+            "default": true
+          },
+          "dbt.whatsNewManifestUrl": {
```
ralphstodomingo: we no longer need this

ralphstodomingo: Removed in cb4660af — dropped the setting from `package.json` and collapsed the feed URL to the fixed constant in `whatsNewPanel.ts`. PR description updated too.

src/webview_provider/whatsNewPanel.ts
```diff
@@ -0,0 +1,372 @@
...
+
+    this.dbtTerminal.debug(
+      "whatsNew:fetchManifest",
+      `Fetching What's New feed from ${url}`,
+    );
+
+    try {
+      const response = await fetch(url, {
+        headers: { Accept: "application/rss+xml, application/xml, text/xml" },
+      });
```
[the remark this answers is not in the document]
ralphstodomingo: Good catch — fixed in 8b3a76d6. Added an `AbortController` with a 10s timeout, pass `signal` to `fetch`, and `clearTimeout` in a `finally`. On abort the existing `catch` falls back to the cached manifest, so a stalled endpoint degrades gracefully instead of hanging the panel.

dev-punia-altimate (commented):
## 🤖 Code Review — OpenCodeReview (Gemini) — 9 finding(s)

- 7 anchored to a line (posted inline when the comment stream is on)
- 2 without a line anchor

#### 1. `src/dbtPowerUserExtension.ts` (L98-L103)

**[🟠 MEDIUM]** The `whatsNewPanel` inherits from `AltimateWebviewProvider` (which implements `dispose()`). To prevent memory leaks when the extension is deactivated, `this.whatsNewPanel` should be added to the `this.disposables.push(...)` array.

**Suggested change:**
```suggestion
    private mcpServer: DbtPowerUserMcpServer,
    private altimateRequest: AltimateRequest,
    private altimateAuthService: AltimateAuthService,
    private whatsNewPanel: WhatsNewPanel,
  ) {
    this.disposables.push(
      this.whatsNewPanel,
```

#### 2. `webview_panels/src/modules/whatsNew/icons.tsx` (L73-L76)

**[🟠 MEDIUM]** The `AltimateWordmark` SVG uses highly generic IDs (`g0`, `g1`, `g2`) for its `linearGradient` definitions. Because SVG element IDs are globally scoped in the HTML document, generic IDs can lead to visual bugs (wrong gradients being applied) if another SVG on the same page uses the same IDs, or if this component is rendered multiple times. It is recommended to use more unique, namespaced IDs (e.g., `altimate-wordmark-g0`, `altimate-wordmark-g1`) and update the `fill="url(#g0)"` references accordingly.

**Suggested change:**
```suggestion
    <defs>
      <linearGradient
        id="altimate-wordmark-g0"
        x1="166.476"
```

#### 3. `webview_panels/src/modules/whatsNew/icons.tsx` (L113-L114)

**[🔵 LOW]** The `tag` prop is currently typed as a broad `string`. To improve type safety and leverage IDE autocomplete, it should be restricted to a union of the specific supported string literals that match your switch cases and the `Tag` type in `WhatsNew.tsx`.

**Suggested change:**
```suggestion
export const TagIcon = ({ tag }: { tag: "new" | "improved" | "beta" | "fixed" }): JSX.Element | null => {
  switch (tag) {
```

#### 4. `src/webview_provider/whatsNewPanel.ts`

**[🔴 HIGH]** This logic inadvertently suppresses the What's New panel for all **existing users** upon their first upgrade to this feature. Because `LAST_SEEN_VERSION_KEY` is introduced in this release, `lastSeen` will be `undefined` for both fresh installs and existing users upgrading, causing the extension to exit early here.

To differentiate existing users from true fresh installs, you should check a pre-existing state key, such as `showSetupWalkthrough`:

```typescript
    const showSetupWalkthrough = this.dbtProjectContainer.getFromGlobalState("showSetupWalkthrough");
    if (showSetupWalkthrough === undefined || showSetupWalkthrough === true) {
      // Truly a fresh install: defer to the setup walkthrough.
      return;
    }

    // For existing users on their first upgrade to this feature, lastSeen is undefined.
    if (!lastSeen || WhatsNewPanel.isMinorOrMajorUpgrade(lastSeen, current)) {
      this.show("auto");
    }
```

#### 5. `webview_panels/src/modules/whatsNew/WhatsNew.tsx` (L167-L176)

**[🟠 MEDIUM]** The `IntersectionObserver` callback's `entries` parameter only includes elements whose intersection status has *changed*, not all currently observed elements. If multiple sections change state (e.g., when scrolling quickly or when an element leaves the viewport), `.filter((entry) => entry.isIntersecting)` only looks at the elements that just entered the window, ignoring elements that are already in the window and haven't changed status. This can cause the active month indicator to fall out of sync or get stuck.

To properly implement scrollspy, consider maintaining a state or `ref` map of all currently intersecting elements and recalculating the topmost element from that complete collection whenever `entries` update.

#### 6. `webview_panels/src/modules/whatsNew/WhatsNew.tsx` (L179-L184)

**[🟠 MEDIUM]** Using `document.querySelector` inside a React component bypasses the Virtual DOM and is considered a React anti-pattern. If there are ever multiple instances of this component or other DOM elements with the same data attribute, it may select the wrong element or cause cross-talk. 

It is highly recommended to use React `refs` to track and interact with these DOM nodes (for instance, using a callback ref to populate a `Map<string, HTMLElement>`).

#### 7. `webview_panels/src/modules/whatsNew/WhatsNew.tsx` (L432-L435)

**[🟠 MEDIUM]** Similar to the observation logic above, avoid using `document.querySelector` to manipulate the DOM directly in React handlers. It's recommended to retrieve the corresponding DOM element from a React `ref` (like a `Map<string, HTMLElement>`) to call `scrollIntoView`.

#### 8. `webview_panels/src/modules/whatsNew/WhatsNew.tsx` (L126-L131)

**[🔴 HIGH]** If the `getWhatsNewManifest` response object is malformed or missing the `items` array, spreading `manifest.items` will throw a `TypeError: manifest.items is not iterable` and crash the component render tree. 

Please add a safety check, such as `manifest.items ? [...manifest.items] : []` or `Array.isArray(manifest.items)`, to avoid potential runtime crashes due to external API changes.

#### 9. `webview_panels/src/modules/whatsNew/WhatsNew.tsx`

**[🔵 LOW]** According to the code quality guidelines, business-related hardcoded strings—especially URL paths—should be avoided in component files. To improve maintainability and ensure consistency, consider moving these external URLs (`PRODUCT_LINKS`, `FULL_CHANGELOG_URL`, `ALTIMATE_SITE_URL`) to a centralized configuration or constants file (e.g., `AppConstants.tsx`).

</details>

src/dbtPowerUserExtension.ts
```diff
@@ -97,6 +98,7 @@ export class DBTPowerUserExtension implements Disposable {
     private mcpServer: DbtPowerUserMcpServer,
     private altimateRequest: AltimateRequest,
     private altimateAuthService: AltimateAuthService,
+    private whatsNewPanel: WhatsNewPanel,
   ) {
     this.disposables.push(
```
dev-punia-altimate, line 103: **[🟠 MEDIUM]** The `whatsNewPanel` inherits from `AltimateWebviewProvider` (which implements `dispose()`). To prevent memory leaks when the extension is deactivated, `this.whatsNewPanel` should be added to the `this.disposables.push(...)` array.

**Suggested change:**
```suggestion
    private mcpServer: DbtPowerUserMcpServer,
    private altimateRequest: AltimateRequest,
    private altimateAuthService: AltimateAuthService,
    private whatsNewPanel: WhatsNewPanel,
  ) {
    this.disposables.push(
      this.whatsNewPanel,
```

webview_panels/src/modules/whatsNew/icons.tsx
```diff
@@ -0,0 +1,227 @@
...
+    <path
+      fillRule="evenodd"
+      clipRule="evenodd"
+      d="M502.747 76.6319V136.295C502.747 141.143 503.785 143.836 505.148 145.215C506.511 146.465 509.345 147.477 514.558 147.477H530.608V168.377H511.638C501.363 168.377 493.1 166.029 487.411 160.793C481.592 155.449 479.039 147.025 479.039 136.295V76.6319H465.217V56.1199H479.039V29.3589H502.747V56.1199H530.587V76.6319H502.747Z"
+      fill="#FAFAFA"
+    />
+    <defs>
+      <linearGradient
+        id="g0"
+        x1="166.476"
```
dev-punia-altimate, line 76: **[🟠 MEDIUM]** The `AltimateWordmark` SVG uses highly generic IDs (`g0`, `g1`, `g2`) for its `linearGradient` definitions. Because SVG element IDs are globally scoped in the HTML document, generic IDs can lead to visual bugs (wrong gradients being applied) if another SVG on the same page uses the same IDs, or if this component is rendered multiple times. It is recommended to use more unique, namespaced IDs (e.g., `altimate-wordmark-g0`, `altimate-wordmark-g1`) and update the `fill="url(#g0)"` references accordingly.

**Suggested change:**
```suggestion
    <defs>
      <linearGradient
        id="altimate-wordmark-g0"
        x1="166.476"
```

webview_panels/src/modules/whatsNew/icons.tsx
```diff
@@ -0,0 +1,227 @@
...
+      >
+        <stop stopColor="#247EFE" />
+        <stop offset="1" stopColor="#30FAFE" />
+      </linearGradient>
+    </defs>
+  </svg>
+);
+
+export const TagIcon = ({ tag }: { tag: string }): JSX.Element | null => {
+  switch (tag) {
```
dev-punia-altimate: **[🔵 LOW]** The `tag` prop is currently typed as a broad `string`. To improve type safety and leverage IDE autocomplete, it should be restricted to a union of the specific supported string literals that match your switch cases and the `Tag` type in `WhatsNew.tsx`.

**Suggested change:**
```suggestion
export const TagIcon = ({ tag }: { tag: "new" | "improved" | "beta" | "fixed" }): JSX.Element | null => {
  switch (tag) {
```

webview_panels/src/modules/whatsNew/WhatsNew.tsx
```diff
@@ -0,0 +1,540 @@
...
+      (entries) => {
+        const topmost = entries
+          .filter((entry) => entry.isIntersecting)
+          .sort(
+            (a, b) => a.boundingClientRect.top - b.boundingClientRect.top,
+          )[0];
+        if (topmost) {
+          setActiveMonth(topmost.target.getAttribute("data-month-key"));
+        }
+      },
```
dev-punia-altimate, line 212: **[🟠 MEDIUM]** The `IntersectionObserver` callback's `entries` parameter only includes elements whose intersection status has *changed*, not all currently observed elements. If multiple sections change state (e.g., when scrolling quickly or when an element leaves the viewport), `.filter((entry) => entry.isIntersecting)` only looks at the elements that just entered the window, ignoring elements that are already in the window and haven't changed status. This can cause the active month indicator to fall out of sync or get stuck.

To properly implement scrollspy, consider maintaining a state or `ref` map of all currently intersecting elements and recalculating the topmost element from that complete collection whenever `entries` update.

webview_panels/src/modules/whatsNew/WhatsNew.tsx
```diff
@@ -0,0 +1,540 @@
...
+      { rootMargin: "-90px 0px -70% 0px" },
+    );
+    for (const group of groups) {
+      const el = document.querySelector(`[data-month-key="${group.key}"]`);
+      if (el) {
+        observer.observe(el);
+      }
+    }
```
dev-punia-altimate, line 220: **[🟠 MEDIUM]** Using `document.querySelector` inside a React component bypasses the Virtual DOM and is considered a React anti-pattern. If there are ever multiple instances of this component or other DOM elements with the same data attribute, it may select the wrong element or cause cross-talk. 

It is highly recommended to use React `refs` to track and interact with these DOM nodes (for instance, using a callback ref to populate a `Map<string, HTMLElement>`).

webview_panels/src/modules/whatsNew/WhatsNew.tsx
```diff
@@ -0,0 +1,540 @@
...
+                key={group.key}
+                className={classes.timelineLink}
+                type="button"
+                data-active={activeMonth === group.key}
+                onClick={() => {
+                  onTimelineNavigate(group.key);
+                  setActiveMonth(group.key);
+                  document
+                    .querySelector(`[data-month-key="${group.key}"]`)
+                    ?.scrollIntoView({ behavior: "smooth", block: "start" });
```
dev-punia-altimate, line 471: **[🟠 MEDIUM]** Similar to the observation logic above, avoid using `document.querySelector` to manipulate the DOM directly in React handlers. It's recommended to retrieve the corresponding DOM element from a React `ref` (like a `Map<string, HTMLElement>`) to call `scrollIntoView`.

webview_panels/src/modules/whatsNew/WhatsNew.tsx
```diff
@@ -0,0 +1,540 @@
...
+      });
+  }, []);
+
+  // Newest first — the month grouping below relies on this order.
+  const items = useMemo(() => {
+    if (!manifest) {
+      return [];
+    }
+    return [...manifest.items].sort((a, b) => b.date.localeCompare(a.date));
+  }, [manifest]);
```
dev-punia-altimate, line 136: **[🔴 HIGH]** If the `getWhatsNewManifest` response object is malformed or missing the `items` array, spreading `manifest.items` will throw a `TypeError: manifest.items is not iterable` and crash the component render tree. 

Please add a safety check, such as `manifest.items ? [...manifest.items] : []` or `Array.isArray(manifest.items)`, to avoid potential runtime crashes due to external API changes.

dev-punia-altimate:

No supported files changed.

webview_panels/src/modules/whatsNew/WhatsNew.tsx
ralphstodomingo: Fixed in 4a7c6732 — the timeline's `scrollIntoView` now resolves the section from the same ref map.

webview_panels/src/modules/whatsNew/WhatsNew.tsx
ralphstodomingo: Fixed in 4a7c6732 — `items` is now checked with `Array.isArray` before spreading. Verified by feeding the panel a manifest with `items` deleted: it mounts and degrades to "0 updates" with all pills disabled, instead of throwing.

src/dbtPowerUserExtension.ts
ralphstodomingo: Fixed in 4a7c6732 — `whatsNewPanel` is now registered in `this.disposables`, matching every other injected provider.

webview_panels/src/modules/whatsNew/icons.tsx
ralphstodomingo: Fixed in 4a7c6732 — ids namespaced to `altimate-wordmark-g0/1/2` with matching `url(#…)` refs. Verified in the rendered bundle: all three gradients resolve, no orphan refs, wordmark still paints.

webview_panels/src/modules/whatsNew/WhatsNew.tsx
ralphstodomingo: Fixed in 4a7c6732 — month sections are now tracked in a `Map` via callback refs; no `document.querySelector` remains in the panel.

webview_panels/src/modules/whatsNew/WhatsNew.tsx
ralphstodomingo: Fixed in 4a7c6732 — you were right, this was a real defect. The observer now accumulates a set of visible months across callbacks and recomputes the topmost from live `getBoundingClientRect()` reads (the rects on `entries` go stale during a fast scroll). Verified by scrolling through all 5 month sections down **and back up** — the highlight tracks correctly in both directions, where before it stuck on the way up.

webview_panels/src/modules/whatsNew/icons.tsx
ralphstodomingo: Fixed in 4a7c6732 — `tag` is now an exported `IconTag` union (`"new" | "improved" | "beta" | "fixed"`) rather than `string`.

saravmajestic (approved):

PR #2047, opened 2026-07-23
fix: stop What's New links opening two browser tabs

ralphstodomingo:
## What

External links in the What's New panel opened **two browser tabs** per click. Fix: add `stopPropagation()` to all seven link handlers.

## Root cause

Each link carries both an `href` (for real-link semantics — focusable, announced as a link, hover URL) and an `onClick` that calls `preventDefault()` and routes the open through the host's `openURL`. `preventDefault()` stops the iframe from navigating, but the click still **bubbled to VS Code's own webview anchor handler**, which opened the same URL a second time via the host's native external-link handling. One click → two opens.

Adding `stopPropagation()` keeps the click from reaching the native handler, so only our controlled `openURL` path fires. The `href` stays, so accessibility is unchanged.

Applied to all seven handlers: the four "Also from Altimate AI" product links, the topbar changelog + altimate.ai links, the hero CTA, and the footer.

## Verification (code-server E2E)

Driven in a real webview (code-server on this branch), one trusted mouse click per link:

| | before | after |
|---|---|---|
| product link (`onProductLink`) | tab opens immediately **+** a second on confirming the prompt = **2 tabs** | one prompt → **1 tab** |
| hero CTA (`onFullChangelog`) | — | one prompt → **1 tab** |

The premature immediate-open is gone; each click now yields exactly one tab.

## Scope note

Reproduced and fixed in **code-server** (browser-based VS Code), where "open external" is a browser tab and both the extension's `openExternal` and the webview's native anchor handler fire. Whether native VS Code (Electron) exhibits the same double-open was **not** verified here — Electron routes external opens differently and may not run the second handler. The fix is a no-op in the single-open case, so it is safe regardless.

Fixes the double-open finding from QA (tracked in the internal ticket).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed external links in the What’s New panel opening twice by preventing the webview host from also handling the click.
* **Improvements**
  * Updated link handling across the changelog entry, product links, header, footer, and main call-to-action to ensure consistent external navigation behavior.
* **Chores**
  * Bumped the extension version to 0.64.0.
  * Updated the `@altimateai/ui-components` dependency.

## Version

Bumps to **0.64.0**.

Minor rather than patch on purpose. 0.63.0 shipped the What's New panel but **never showed it to anyone**: the auto-open compares against a `whatsNew.lastSeenVersion` marker that the feature itself introduces, so on its debut release no user had one and everybody hit the fresh-install early return. Anyone who ran 0.63.0 now *has* that marker, so a minor bump is what actually surfaces the page for them — a patch release would leave it hidden for another cycle.

So this release does two things: ships the double-open fix, and makes the What's New page appear for the first time.

## Also carries: ui-components 0.0.87

Bumps `@altimateai/ui-components` 0.0.86 → 0.0.87 so #3010 (AI-7764 — lineage Depth-controls disabled state + CLL nudges) rides into the extension's lineage panel with this release.

Verified in code-server: `tsc` clean (no API drift), webview builds clean, and the lineage panel renders the `customers` model graph with the updated `TableDetails`/Depth controls and **0 console errors**.

### DIFF
--- a/webview_panels/src/modules/whatsNew/WhatsNew.tsx
+++ b/webview_panels/src/modules/whatsNew/WhatsNew.tsx
@@ -311,6 +311,10 @@ const WhatsNew = (): JSX.Element => {
             title={`Open ${FULL_CHANGELOG_URL} in your browser`}
             onClick={(e) => {
               e.preventDefault();
+              // Also stop the click reaching VS Code's own webview anchor handler,
+              // which would open the URL a second time (once via our openURL, once
+              // via the host's native external-link handling).
+              e.stopPropagation();
               onFullChangelog("error_state");
             }}
           >
@@ -343,6 +347,10 @@ const WhatsNew = (): JSX.Element => {
           title="Open altimate.ai in your browser"
           onClick={(e) => {
             e.preventDefault();
+            // Also stop the click reaching VS Code's own webview anchor handler,
+            // which would open the URL a second time (once via our openURL, once
+            // via the host's native external-link handling).
+            e.stopPropagation();
             onSiteLink("logo");
           }}
         >
@@ -355,6 +363,10 @@ const WhatsNew = (): JSX.Element => {
             title={`Open ${FULL_CHANGELOG_URL} in your browser`}
             onClick={(e) => {
               e.preventDefault();
+              // Also stop the click reaching VS Code's own webview anchor handler,
+              // which would open the URL a second time (once via our openURL, once
+              // via the host's native external-link handling).
+              e.stopPropagation();
               onFullChangelog("topbar");
             }}
           >
@@ -366,6 +378,10 @@ const WhatsNew = (): JSX.Element => {
             title="Open altimate.ai in your browser"
             onClick={(e) => {
               e.preventDefault();
+              // Also stop the click reaching VS Code's own webview anchor handler,
+              // which would open the URL a second time (once via our openURL, once
+              // via the host's native external-link handling).
+              e.stopPropagation();
               onSiteLink("topbar");
             }}
           >
@@ -392,6 +408,10 @@ const WhatsNew = (): JSX.Element => {
             title={`Open ${manifest.base_url} in your browser`}
             onClick={(e) => {
               e.preventDefault();
+              // Also stop the click reaching VS Code's own webview anchor handler,
+              // which would open the URL a second time (once via our openURL, once
+              // via the host's native external-link handling).
+              e.stopPropagation();
               onFullChangelog("cta");
             }}
           >
@@ -411,6 +431,10 @@ const WhatsNew = (): JSX.Element => {
                 title={`Open ${link.url} in your browser`}
                 onClick={(e) => {
                   e.preventDefault();
+                  // Also stop the click reaching VS Code's own webview anchor handler,
+                  // which would open the URL a second time (once via our openURL, once
+                  // via the host's native external-link handling).
+                  e.stopPropagation();
                   onProductLink(link.label, link.url);
                 }}
               >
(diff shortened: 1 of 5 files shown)


dev-punia-altimate (commented):
## 🤖 Code Review — OpenCodeReview (Gemini) — 1 finding(s)

- 1 anchored to a line (posted inline when the comment stream is on)
- 0 without a line anchor

#### 1. `webview_panels/src/modules/whatsNew/WhatsNew.tsx` (L314-L317)

**[🟠 MEDIUM]** The logic for preventing default behavior and stopping event propagation to VS Code's webview anchor handler is repeated multiple times across the file. 

To improve code maintainability and reduce boilerplate, consider extracting this common logic into a reusable higher-order function. For example:

```tsx
const handleLinkClick = (handler: () => void) => (e: React.MouseEvent) => {
  e.preventDefault();
  // Also stop the click reaching VS Code's own webview anchor handler,
  // which would open the URL a second time (once via our openURL, once
  // via the host's native external-link handling).
  e.stopPropagation();
  handler();
};
```

Then you can simplify your `onClick` handlers:
```tsx
onClick={handleLinkClick(() => onFullChangelog("error_state"))}
```

**Suggested change:**
```suggestion
              // Consider extracting this logic into a helper function as suggested in the review comment
              e.stopPropagation();
```

</details>

webview_panels/src/modules/whatsNew/WhatsNew.tsx
```diff
@@ -311,6 +311,10 @@ const WhatsNew = (): JSX.Element => {
             title={`Open ${FULL_CHANGELOG_URL} in your browser`}
             onClick={(e) => {
               e.preventDefault();
+              // Also stop the click reaching VS Code's own webview anchor handler,
+              // which would open the URL a second time (once via our openURL, once
+              // via the host's native external-link handling).
+              e.stopPropagation();
```
dev-punia-altimate: **[🟠 MEDIUM]** The logic for preventing default behavior and stopping event propagation to VS Code's webview anchor handler is repeated multiple times across the file. 

To improve code maintainability and reduce boilerplate, consider extracting this common logic into a reusable higher-order function. For example:

```tsx
const handleLinkClick = (handler: () => void) => (e: React.MouseEvent) => {
  e.preventDefault();
  // Also stop the click reaching VS Code's own webview anchor handler,
  // which would open the URL a second time (once via our openURL, once
  // via the host's native external-link handling).
  e.stopPropagation();
  handler();
};
```

Then you can simplify your `onClick` handlers:
```tsx
onClick={handleLinkClick(() => onFullChangelog("error_state"))}
```

**Suggested change:**
```suggestion
              // Consider extracting this logic into a helper function as suggested in the review comment
              e.stopPropagation();
```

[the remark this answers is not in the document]
ralphstodomingo: Applied in 18b72a7b — extracted an `openExternal(open)` helper so the `preventDefault()`/`stopPropagation()` pair and its rationale are documented once instead of seven times (net −25 lines). Behaviour is unchanged: same two calls in the same order, same callbacks. Re-verified in code-server after the refactor — the product link and the hero CTA each open exactly one tab with no premature open before the prompt.

ralphstodomingo: Applied in 18b72a7b — extracted an `openExternal(open)` helper so the `preventDefault()`/`stopPropagation()` pair and its rationale are documented once instead of seven times (net −25 lines). Behaviour is unchanged: same two calls in the same order, same callbacks. Re-verified in code-server after the refactor — the product link and the hero CTA each open exactly one tab with no premature open before the prompt.

saravmajestic (approved):

dev-punia-altimate:

No supported files changed.

